PR #95, opened 2017-03-09
add validation client id

Lagovas:


gene-eu:
@secumod, mind looking at input sanitization with extreme precision you're capable of? ;)

PR #103, opened 2017-11-01
unify makefile

vixentael:
unify makefile: add `install`, `install_all`, `test_all` commands

similar to
https://github.com/cossacklabs/themis/pull/248

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,9 +37,12 @@ temp_copy:
 	@GOPATH=$(ABS_TEMP_GOPATH) go get github.com/cossacklabs/acra/cmd/...
 
 
-build: temp_copy
+install: temp_copy
 	@mkdir -p $(BIN_PATH)
 	@cp $(TEMP_GOPATH)/bin/* $(BIN_PATH)
+
+# alias for unification with other products
+install_all: install
 
 clean:
 	@rm -rf $(BIN_PATH)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -37,10 +37,13 @@ temp_copy:
 	@GOPATH=$(ABS_TEMP_GOPATH) go get github.com/cossacklabs/acra/cmd/...
 
 
-build: temp_copy
+install: temp_copy
```
Lagovas: here `make install` do not the same as in other our products. here it will copy binaries to `REPO_DIR/build` (https://github.com/cossacklabs/acra/blob/fa860a2f0f815165f954617985ef8ce1ae9129dd/Makefile#L4) when other our products place binaries to /usr/lib/*. I think we should install acra binaries to /usr/bin as default
and after changing path is important to drop this command - https://github.com/cossacklabs/acra/blob/fa860a2f0f815165f954617985ef8ce1ae9129dd/Makefile#L48 and explicitly list all binaries that should be deleted (acraproxy, acraserver, acra_genkeys, acra_rollback, acra_genpoisonrecord, acra_addzone) or it will delete `/usr/bin` folder when user will run `sudo make clean` :)

vixentael:
as result we decided that it's okay to install acra to the local folder using `make install`, because "normal users" will use `go get github.com/cossacklabs/acra/cmd/...` that will install acra according to their environment

PR #106, opened 2018-02-23
refactor networking, postgresql tls support

Lagovas:
* added support of tls connections between app and database. now acraserver in such case works like mitm and work for clients like tls server and for postgresql like tls client
* move connection logic into network package that allow to wrap raw go lang net.Conn connections with other implementations like secure sesson/tls, added ability to connect via unix socket that remove problems with testing and TCP_WAIT between tests
* added tls implementation of connection wrapper
* add circleci v2.0 config and drop old
* run integrations tests in ci
* start to use `github.com/sirupsen/logrus` package for logging that provide good structured and controlled logs
* some fixes in tests
* use keystore in secure session callbacks instead hardcoded lookup of keys


cmd/acraproxy/acraproxy.go
```diff
@@ -198,25 +73,38 @@ func handleClientConnection(config *Config, connection net.Conn) {
...
 	if err != nil {
 		log.Printf("Error: %v\n", utils.ErrorMessage("can't connect to acra", err))
 		return
 	}
 	defer acraConn.Close()
-	ssession, err := initializeSecureSession(config, acraConn)
+
+	log.Printf("Info: send client id <%v>\n", string(config.ClientId))
+
+	acraConn.SetDeadline(time.Now().Add(time.Second * 2))
```
vixentael: how long is the connection live time? 2 seconds?

cmd/acraproxy/acraproxy.go
```diff
@@ -275,31 +194,56 @@ func main() {
...
 		os.Exit(1)
 	}
-	if *acraHost == "" {
-		fmt.Println("Error: you must specify host to acra")
-		flag.Usage()
-		os.Exit(1)
-	}
 
 	if *verbose {
+		log.SetLevel(log.InfoLevel)
```
vixentael: i believe that in next commits we should use own logger interface, that is linked to any logging system we want, instead of using several log providers inside code

cmd/acraproxy/acraproxy.go
```diff
@@ -198,25 +73,38 @@ func handleClientConnection(config *Config, connection net.Conn) {
...
-		log.Printf("Error: %v\n", utils.ErrorMessage("can't initialize secure session", err))
+		log.Printf("Error: %v\n", utils.ErrorMessage("can't wrap acra connection with secure session", err))
 		return
 	}
-	errCh := make(chan error)
-	log.Println("Debug: secure session initialized")
-	go proxyClientConnections(connection, acraConn, ssession, errCh)
-	go proxyAcraConnections(connection, acraConn, ssession, errCh)
-	err = <-errCh
+	acraConn.SetDeadline(time.Time{})
```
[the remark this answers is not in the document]
Lagovas: it reset deadline

cmd/acraproxy/acraproxy.go
Lagovas: it was one-time tests and I forgot to drop one log system and use the second
I will drop in this PR soon

vixentael:
LGTM

PR #108, opened 2018-02-25
Basic firewall interface

storojs72:


firewall/filesystem_storage_firewall.go
```diff
@@ -0,0 +1,41 @@
+package firewall
+
+import (
+	"errors"
+)
+
+type FilesystemAcraFirewall struct {
```
Lagovas: does this firewall use any filesystem specific logic? if not then will be good to remove `filesystem` prefix and name just `Firewall` (we haven't any rules to add prefixes like `acra` everywhere :)

firewall/filesystem_storage_firewall.go
```diff
@@ -0,0 +1,41 @@
...
+
+	handlers []QueryHandler
+
+	storageDirectory string
```
Lagovas: this field doesn't used anywhere in this module

firewall/filesystem_storage_firewall.go
```diff
@@ -0,0 +1,41 @@
...
+var ErrHandlersNotSpecified = errors.New("at least one query handler should be specified")
+
+
+func NewFilesystemAcraFirewall(storageDirectory string) (*FilesystemAcraFirewall, error) {
+
+	return &FilesystemAcraFirewall{storageDirectory:storageDirectory}, nil
+}
+
+
+func (firewall *FilesystemAcraFirewall) AddSpecificHandler(handler QueryHandler){
```
Lagovas: what do we mean by `specific`? if nothing then enough to name `AddHandler` I think or `AddQueryHandler`. It will be more informative

firewall/filesystem_storage_firewall.go
```diff
@@ -0,0 +1,41 @@
...
+
+	firewall.handlers = append(firewall.handlers, handler)
+}
+
+func (firewall *FilesystemAcraFirewall) HandleQuery(query string) error{
+
+	if len(firewall.handlers) < 1 {
```
Lagovas: Firewall may contain 0 handlers if user will turn off all rules. As I remember we allow to configure it and leave the ability to run without any checks. If so we don't need this check and error.

firewall/firewall_test.go
```diff
@@ -0,0 +1,163 @@
...
+		"INSERT INTO SalesStaff3 (StaffID, FullName)",
+		"INSERT INTO Customers (CustomerName, ContactName, Address, City, PostalCode, Country) VALUES ('Cardinal', 'Tom B. Erichsen', 'Skagen 21', 'Stavanger', '4006', 'Norway');",
+		"INSERT INTO Customers (CustomerName, City, Country) VALUES ('Cardinal', 'Stavanger', 'Norway');",
+		"INSERT INTO Production.UnitMeasure (Name, UnitMeasureCode,	ModifiedDate) VALUES (N'Square Yards', N'Y2', GETDATE());",
+		"INSERT INTO T1 DEFAULT VALUES;",
+		"INSERT INTO dbo.Points (PointValue) VALUES (CONVERT(Point, '1,5'));",
+		"INSERT INTO dbo.Points (PointValue) VALUES (CAST ('1,99' AS Point));",
+	}
+
+	home, err := homedir.Dir()
```
Lagovas: why home directory? How it will be used? is it cleaned after tests?
Better to use [tempdir](https://golang.org/pkg/io/ioutil/#TempDir) in tests to not leave trash in user's homedir.

firewall/firewall_test.go
```diff
@@ -0,0 +1,163 @@
...
+		"INSERT INTO dbo.Points (PointValue) VALUES (CONVERT(Point, '1,5'));",
+		"INSERT INTO dbo.Points (PointValue) VALUES (CAST ('1,99' AS Point));",
+	}
+
+	blackList := [] string {
+		"INSERT INTO SalesStaff1 VALUES (1, 'Stephen', 'Jiang');",
+		"SELECT AVG(Price)FROM Products;",
+	}
+
+	home, err := homedir.Dir()
```
Lagovas: same question about homedir

firewall/handlers/blacklist_handler.go
```diff
@@ -0,0 +1,58 @@
...
+	yes, _ := handler.contains(query)
+	if yes {
+		return ErrQueryInBlacklist
+	}
+	return nil
+}
+
+func(handler * BlacklistHandler) AddQueriesToBlacklist(queries []string) {
+
+	for index := 0; index < len(queries); index++ {
```
Lagovas:
```
for _, query := range queries{
  handler.blackQueries = append(handler.blackQueries, query)
}
```
looks shorter and more readable, did it?

firewall/handlers/blacklist_handler.go
```diff
@@ -0,0 +1,58 @@
...
+
+	for i := 0; i < len(queries); i++ {
+		yes, index := handler.contains(queries[i])
+		if yes {
+			handler.blackQueries = append(handler.blackQueries[:index], handler.blackQueries[index+1:]...)
+		}
+	}
+}
+
+func(handler * BlacklistHandler) contains(query string) (bool, int) {
```
Lagovas: this function used twice in two handlers
what about to make it as simple function like `contains(query string, queries[]string)(bool, int)` and use it in handlers? then we need test this function just once

firewall/handlers/blacklist_handler.go
```diff
@@ -0,0 +1,58 @@
...
+
+	var index int
+	for _, blackQuery := range handler.blackQueries {
```
Lagovas: first arg in forloop is current index. we can use it and make function shorter and simpler
```
for i, badQuery := range handler.blackQueries {
  if badQuery == query{
     return true, i
  }
}
return false, 0
```

firewall/handlers/handlers_util.go
```diff
@@ -0,0 +1,15 @@
+package handlers
+
+func removeDuplicates(xs *[]string) {
```
Lagovas: too hard to read this function. Here `xs`, `i`, `x`, `j` four short unreadable variables used in 4 rows. 
would be great to add tests for this function

vixentael: it is looking for exact string match, right? (for now)

firewall/handlers/whitelist_handler.go
```diff
@@ -0,0 +1,50 @@
...
+
+type WhitelistHandler struct {
+	whiteQueries[] string
+}
+
+var ErrQueryNotInWhitelist = errors.New("query not in whitelist")
+
+func NewWhitelistHandler(whiteQueries []string) (*WhitelistHandler, error) {
+
+	return &WhitelistHandler{whiteQueries:whiteQueries}, nil
```
vixentael: shouldn't we check for duplicates here?

firewall/handlers/whitelist_handler.go
```diff
@@ -0,0 +1,50 @@
...
+
+func(handler * WhitelistHandler) AddQueriesToWhitelist(queries []string) {
+
+	for index := 0; index < len(queries); index++ {
+		handler.whiteQueries = append(handler.whiteQueries, queries[index])
+	}
+	removeDuplicates(&handler.whiteQueries)
+}
+
+func (handler * WhitelistHandler) RemoveQueriesFromWhitelist(queries []string) {
```
vixentael: i think we should add test for this function

firewall/firewall_test.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	for i := 0; i < len(sqlInsertQueries); i++ {
+		err = firewall.HandleQuery(sqlInsertQueries[i])
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	blacklistHandler.AddQueriesToBlacklist([]string{"INSERT INTO dbo.Points (PointValue) VALUES (CONVERT(Point, '1,5'));"})
```
vixentael: can we move this string rule into separate variable? because we used it 4 times `"INSERT INTO dbo.Points (PointValue) VALUES (CONVERT(Point, '1,5'));"`

vixentael:
I really like how this functionality is appearing. It's gonna be a cool feature when implemented

firewall/filesystem_storage_firewall.go
storojs72: Agree. Done.

firewall/filesystem_storage_firewall.go
storojs72: Agree.

firewall/filesystem_storage_firewall.go
storojs72: Agree. Changed.

firewall/filesystem_storage_firewall.go
storojs72: Agree. Removed

firewall/firewall_test.go
storojs72: Removed as unnecessary

firewall/firewall_test.go
storojs72: Removed

firewall/handlers/blacklist_handler.go
storojs72: Agree. Changed

firewall/handlers/blacklist_handler.go
storojs72: Agree. Moved to 'handlers' subpackage

firewall/handlers/blacklist_handler.go
storojs72: Agree. Changed

firewall/handlers/handlers_util.go
storojs72: Agree. Test is added

storojs72: @vixentael yes

firewall/handlers/whitelist_handler.go
storojs72: Thanks. Added.

firewall/handlers/whitelist_handler.go
storojs72: It is implicitly tested within main firewall tests of black/white list

firewall/firewall_test.go
storojs72: Agree. Done.

firewall/handlers/handlers_util.go
```diff
@@ -1,6 +1,7 @@
 package handlers
 
 func removeDuplicates(xs *[]string) {
```
Lagovas: what about to return result as new slice with unique values? anyway here memory consumption not decreased and in both cases where we use `removeDuplicates` we copy all values and trash at end of slice `*xs = (*xs)[:j]` it doesn't create new slice, it just move `len` fieldn in slice structure, capacity of slice left unchanged

storojs72: It doesn't matter for me actually. I have found this code fragment on stackoverflow and it had >20 upvotes so I decided not to create a wheel. If you really think it needs to be refactored, I will do it

Lagovas: will be good and code will look simpler

PR #110, opened 2018-02-27
refactor logging

Lagovas:
use logrus instead default logger to use helpers and semantically correct method instead prefixes (`log.Errorln` instead `log.Println("Error:")`) 
plus it add more structured logs that looks like `time="2018-02-27T12:53:20+02:00" level=error msg="can't connect to acra" error="dial unix /tmp/unix_socket_10003: connect: no such file or directory"` which we easily can parse/process if need by level or extra fields that we can add in a future (for example to measure performance or count events)


cmd/acraserver/listener.go
```diff
@@ -84,15 +77,15 @@ func (server *SServer) handleConnection(connection net.Conn) {
...
 		if closeErr := connection.Close(); closeErr != nil {
-			log.WithError(closeErr).Println("Error: can't close connection")
+			log.WithError(closeErr).Println("can't close connection")
 		}
 		return
 	}
 	clientSession.connection = wrappedConnection
 
-	log.Println("Debug: secure session initialized")
+	log.Debugln("secure session initialized")
```
vixentael: maybe `log.Infof`?

cmd/acraserver/listener.go
```diff
@@ -129,38 +122,39 @@ to db and decrypting responses from db
...
 	}
 
 	wrappedConnection, _, err := server.config.ConnectionWrapper.WrapServer(connection)
 	if err != nil {
+		log.WithError(err).Errorln("can't wrap connection")
 		return
 	}
 	clientSession.connection = wrappedConnection
-	log.Println("Debug: http api secure session initialized")
+	log.Debugln("http api secure session initialized")
```
vixentael: same, maybe use info level?

cmd/acraserver/listener.go
Lagovas: on my opinion it's unnecessary info and useful only for debugging

PR #112, opened 2018-02-28
direct connection to acraserver with ssl

Lagovas:
* dropped hardcoded path to tls certificates
* added handling SIGINT to correct closing sockets (@mozhmike maybe will be useful in webui to handle SIGINT and call `server.Close()`)
* added test without acraproxy: app(tls) <-> acraserver(tls) <-> postgresql(tls)

because official postgresql image doesn't support ssl, I changed image to non-official with such support and added server.[key|crt] from image to `tests/` to use via acraserver.
and because postgresql support tls only over tcp (not over unix), and test with tls usage use TCP (with TCP TIME WAIT problem when port is binded even after killing process) I added changing ports between go versions


cmd/acraserver/acraserver.go
```diff
@@ -16,14 +16,15 @@ package main
 import (
 	"crypto/tls"
 	"flag"
+	"net/http"
+	_ "net/http/pprof"
```
vixentael: is it used?

cmd/utils.go
```diff
@@ -30,6 +33,39 @@ func init() {
...
+}
+
+func (handler *SignalHandler) AddCallback(callback SignalCallback) {
+	handler.callbacks = append(handler.callbacks, callback)
+}
+
+// Register should be called as goroutine
+func (handler *SignalHandler) Register() {
+	signal.Notify(handler.ch, handler.osSignal)
+	<-handler.ch
```
vixentael: is it channel? https://tour.golang.org/concurrency/2

where it sends value from `handler.ch`?

tests/test.py
```diff
@@ -166,28 +176,35 @@ def tearDownModule():
...
-
+    ACRA_PORT = int(os.environ.get('TEST_ACRA_PORT', 10003))
     ACRA_BYTEA = 'hex_bytea'
     DB_BYTEA = 'hex'
     WHOLECELL_MODE = False
     ZONE = False
     DEBUG_LOG = False
     TEST_DATA_LOG = False
     maxDiff = None
+    # hack to simplify handling errors on forks and don't check `if hasattr(self, 'proxy_1')`
```

cmd/utils.go
Lagovas: https://golang.org/pkg/os/signal/#Notify 
here we pass channel and block by reading until app receive signal

cmd/acraserver/acraserver.go
Lagovas: yes, here - https://github.com/cossacklabs/acra/pull/112/files/591904efc2140f6b44cbc2d3ab8441e4c72811eb#diff-963cfdc2dd5ec26b3315f2be5214f061R169
for debugging golang provide way to fetch profiling data over http

Lagovas:
close for remove noise of notifications

PR #113, opened 2018-03-02
direct connection to acraserver with ssl

Lagovas:
* dropped hardcoded path to tls certificates
* added handling SIGINT to correct closing sockets (@mozhmike maybe will be useful in webui to handle SIGINT and call server.Close())
* added test without acraproxy: app(tls) <-> acraserver(tls) <-> postgresql(tls)
* fixed tests and cleaning zombi processes

because official postgresql image doesn't support ssl, I changed image to non-official with such support and added server.[key|crt] from image to tests/ to use via acraserver.
and because postgresql support tls only over tcp (not over unix), and test with tls usage use TCP (with TCP TIME WAIT problem when port is binded even after killing process) I added changing ports between go versions


Lagovas:
added test for tls between acra and proxy and added more params to setup tls connection

PR #122, opened 2018-03-09
update tests

Lagovas:


Lagovas:
pass `sslmode=requir`e to acra_rollback if `TEST_TLS=on` and `sslmode=disable` if `TEST_TLS=off`

PR #123, opened 2018-03-09
Mozh/t500 graceful restart

mozhmike:


cmd/acra_configui/acra_configui.go
```diff
@@ -107,10 +115,29 @@ func SubmitSettings(w http.ResponseWriter, r *http.Request) {
...
+	tplPath, err = utils.AbsPath(tplPath)
+	if err != nil {
+		log.WithError(err).Errorf("no config file[%v]", tplPath)
+	}
+
+	var parsedTemplate *template.Template
+	parsedTemplate, err = template.ParseFiles(tplPath)
+	if err != nil {
+		log.WithError(err).Errorf("err while parsing template - %v", tplPath)
+		syscall.Exit(1)
```
Lagovas: Will be better if we will check template files on app start before running service. after app start check and go down if templates not exist. it will be better than go down at runtime

cmd/acra_configui/acra_configui.go
```diff
@@ -175,9 +202,9 @@ func main() {
 
 	http.HandleFunc("/index.html", index)
 	http.HandleFunc("/", index)
-	http.Handle("/static/", http.StripPrefix("/static/", http.FileServer(http.Dir("./static"))))
+	http.Handle("/static/", http.StripPrefix("/static/", http.FileServer(http.Dir("./cmd/acra_configui/static"))))
```
Lagovas: let's use path to static from cli args/config?

cmd/acraserver/acraserver.go
```diff
@@ -25,10 +25,15 @@ import (
...
 	"github.com/cossacklabs/acra/utils"
 	log "github.com/sirupsen/logrus"
+	"os/signal"
+	"time"
+	"errors"
 )
 
 // DEFAULT_CONFIG_PATH relative path to config which will be parsed as default
 var DEFAULT_CONFIG_PATH = utils.GetConfigPathByName("acraserver")
+var SignalsChannel = make(chan os.Signal)
```
Lagovas: https://github.com/cossacklabs/acra/blob/master/cmd/utils.go#L37
We has signal handler that used in acraproxy/acraserver (added before your PR:). Better to use one implementation instead many in different places. Better to extend existing in this case imho.

cmd/acraserver/acraserver.go
```diff
@@ -152,7 +158,13 @@ func main() {
 		}
 	}
 
-	server, err := NewServer(config, keyStore)
+	var server *SServer
+	if os.Getenv("_GRACEFUL_RESTART") == "true" {
+		server, err = NewFromFD(config, keyStore, 3, 4)
```
Lagovas: what means `3, 4`? Can we use here some constants? 
plus why not extend NewServer and add new function?

cmd/acraserver/acraserver.go
```diff
@@ -152,7 +158,13 @@ func main() {
 		}
 	}
 
-	server, err := NewServer(config, keyStore)
+	var server *SServer
+	if os.Getenv("_GRACEFUL_RESTART") == "true" {
```
Lagovas: What about to pass value from env variable through [string.ToLower](https://golang.org/pkg/strings/#ToLower) to support values `True`|`TRUE`?

cmd/acraserver/acraserver.go
```diff
@@ -175,7 +187,79 @@ func main() {
...
+			// Stop accepting new connections
+			server.Stop(server.socketACRA)
+			if *withZone && !*disableHTTPApi {
+				server.Stop(server.socketAPI)
+			}
+			// Wait a maximum of 10 seconds for existing connections to finish
+			err := server.WaitWithTimeout(10 * time.Second)
+			if err == WaitTimeoutError {
+				log.Debugf("Timeout when stopping server, %d active connections will be cut.\n", server.ConnectionsCounter())
+				os.Exit(-127)
```
Lagovas: why `-127`? What about to add some error message to output? For the case when app will work without debug output

cmd/acraserver/listener.go
```diff
@@ -131,6 +208,54 @@ func (server *SServer) Start() {
 	}
 }
 
+func (server *SServer) Stop(socket *net.TCPListener) {
+	socket.SetDeadline(time.Now())
+}
+
+func (server *SServer) ListenerFD(socket *net.TCPListener) (uintptr, error) {
```
Lagovas: What about to `ListenerFileDescriptor`? `uintptr` and `FD` not so obvious imho
Plus this method doesn't use any fields from `SServer` struct and better to implement it as function. Than we can use it in other places where will need to get descriptor from `TCPListener`
Plus we support not only `TCPListener`, `UnixListener` too (and connections). So what about to write something like that - https://play.golang.org/p/LuUqCB2YqyP ? Wrap it into functions `GetDescriptorFromListener[Connection]` and use where we need

cmd/acraserver/listener.go
```diff
@@ -131,6 +208,54 @@ func (server *SServer) Start() {
...
+
+	select {
+	case <-timeout.C:
+		return WaitTimeoutError
+	case <-wait:
+		return nil
+	}
+}
+
+func (server *SServer) Addr(socket *net.TCPListener) net.Addr {
```
Lagovas: the same, we need support unix too

cmd/acraserver/listener.go
```diff
@@ -152,12 +277,61 @@ func (server *SServer) handleCommandsConnection(connection net.Conn) {
...
 // start listening commands connections from proxy
-func (server *SServer) StartCommands() {
-	listener, err := network.Listen(server.config.GetAcraAPIConnectionString())
+func (server *SServer) StartCommands(graceful bool) {
+	var listener net.Listener
+	var listenerGraceful *net.TCPListener
+	var err error
+	var connection net.Conn
+	if graceful == true {
+		listenerGraceful, err = network.ListenTCP(server.config.GetAcraAPIConnectionString())
```
Lagovas: You can use this function - https://github.com/cossacklabs/acra/blob/master/network/utils.go#L22 to support unix/tcp

network/utils.go
```diff
@@ -31,6 +34,24 @@ func Listen(connectionString string) (net.Listener, error) {
 	}
 }
 
+func ListenTCP(connectionString string) (*net.TCPListener, error) {
```
Lagovas: https://github.com/cossacklabs/acra/blob/master/network/utils.go#L22 can we use it or extend/replace something of these 2?

cmd/acraserver/acraserver.go
vixentael: I would move this `if` to a separate function, like 

```
func isGracefulRestart() bool {
	return strings.ToLower(os.Getenv("_GRACEFUL_RESTART")) == "true"
}
```

because we are checking it's value in many places, better to declare it once (in case we will rename env variable)

cmd/acraserver/acraserver.go
```diff
@@ -175,7 +187,79 @@ func main() {
...
+	} else {
+		go server.Start(true)
 	}
-	server.Start()
+
+	signal.Notify(SignalsChannel, syscall.SIGHUP, syscall.SIGTERM)
+	var WaitTimeoutError = errors.New("timeout")
+	for sig := range SignalsChannel {
+		if sig == syscall.SIGTERM {
+			// Stop accepting new connections
```
vixentael: I believe that code would be much readable if we divide this laaarge `if` handlers into separate functions.

like

```
func handleSigTerm(...)
func handleSigHup(...)
```

smaller, more readable

cmd/acraserver/acraserver.go
```diff
@@ -175,7 +187,79 @@ func main() {
...
+			server.Stop(server.socketACRA)
+			if *withZone && !*disableHTTPApi {
+				server.Stop(server.socketAPI)
+			}
+			// Wait a maximum of 10 seconds for existing connections to finish
+			err := server.WaitWithTimeout(10 * time.Second)
```
vixentael: shall we define 10 as constant and move it to the top of file? for changing easily if needed?

like

```
var defaultConnectionTimeout int = 10
```

cmd/acraserver/acraserver.go
vixentael: maybe more `"_GRACEFUL_RESTART"` as constant as well?

```
var gracefulRestartSetting string = "_GRACEFUL_RESTART"
```

cmd/acraserver/listener.go
```diff
@@ -152,12 +277,61 @@ func (server *SServer) handleCommandsConnection(connection net.Conn) {
...
+			connection, err = listener.Accept()
+		}
+
+		if err != nil {
+			// log.WithError(err).Errorf("can't accept new connection (%v)", connection.RemoteAddr())
+			log.WithError(err).Errorf("can't accept new connection")
+			continue
+		}
+		// unix socket and value == '@'
+		if len(connection.RemoteAddr().String()) == 1 {
```
vixentael: is it a good idea to check unix socket value like string length?

cmd/acraserver/listener.go
vixentael: Instead of having `if` checks everywhere, can we create two separate functions, one for unix sockets, another for TCP, and use `if` once:

```
func (server *SServer) StartCommands(graceful bool) {

...

	if graceful == true {
	   handleGracefulSocket()
	} else {
	   handleSocket() 
	}
}

func handleGracefulSocket()
func handleSocket()
```

cmd/acraserver/listener.go
Lagovas: for unix socket RemoteAddr always == "@", LocalAddr has path. For TCP socket and RemoteAddr and LocalAddr has some path like "127.0.0.1:2345" and "127.0.0.1:5431". How you suggest to checK?

cmd/acra_configui/acra_configui.go
```diff
@@ -107,10 +115,29 @@ func SubmitSettings(w http.ResponseWriter, r *http.Request) {
 
 func index(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Security-Policy", "require-sri-for script style")
-	parsedTemplate, _ := template.ParseFiles(filepath.Join("static", "index.html"))
-	var outConfigParams configParamsYAML
-	configParamsYAML, err := ioutil.ReadFile("acraserver_config_vars.yaml")
-	check(err)
+	tplPath := fmt.Sprintf("./cmd/acra_configui/%s", filepath.Join("static", "index.html"))
```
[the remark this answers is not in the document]
mozhmike: done

cmd/acra_configui/acra_configui.go
mozhmike: done

cmd/acra_configui/acra_configui.go
mozhmike: Implemented cli-param

cmd/acraserver/acraserver.go
mozhmike: Refactored

cmd/acraserver/acraserver.go
mozhmike: Constants used

cmd/acraserver/acraserver.go
mozhmike: Refactored

cmd/acraserver/listener.go
mozhmike: Refactored

cmd/acraserver/listener.go
mozhmike: Refactored

cmd/acraserver/listener.go
mozhmike: Refactored

network/utils.go
mozhmike: done

cmd/acraserver/acraserver.go
mozhmike: Used our handler wrapper

cmd/acraserver/acraserver.go
mozhmike: Moved to constants

cmd/acraserver/listener.go
mozhmike: done

cmd/acra_configui/acra_configui.go
```diff
@@ -105,12 +116,37 @@ func SubmitSettings(w http.ResponseWriter, r *http.Request) {
...
+
+	parsedTemplate, err = template.ParseFiles(tplPath)
+	if err != nil {
+		log.WithError(err).Errorf("err while parsing template - %v", tplPath)
+		syscall.Exit(1)
+	}
+}
+
+func parseConfig() []byte {
+	configPath, err_ := utils.AbsPath(fmt.Sprintf("./%s", CONFIG_PATH))
```
Lagovas: 1. you can read file with relative path without converting to absolute. `CONFIG_PATH` constains relative path `config/blabla` and it can be read without transformations
2. what if user want to use config with custom path like `/var/acra/some.conf`? or how we can change path if we will distribute our package with binary and place config in different place? for example binary will be placed to /usr/local/bin and configs to `/etc/acra/*` or `/usr/local/etc` ?

cmd/acra_configui/acra_configui.go
```diff
@@ -105,12 +116,37 @@ func SubmitSettings(w http.ResponseWriter, r *http.Request) {
 	w.Write(jsonToServer)
 }
 
+func parseTemplate() {
+	tplPath := fmt.Sprintf(filepath.Join(*staticPath, "index.html"))
```
Lagovas: what about to use function parameter `staticPath` (`func parseTemplate(staticPath string)`) instead global variable?

cmd/acra_configui/acra_configui.go
```diff
@@ -105,12 +116,37 @@ func SubmitSettings(w http.ResponseWriter, r *http.Request) {
...
+	tplPath, err = utils.AbsPath(tplPath)
+	if err != nil {
+		log.WithError(err).Errorf("no config file[%v]", tplPath)
```
Lagovas: return? syscall.exit()?

cmd/acra_configui/acra_configui.go
```diff
@@ -105,12 +116,37 @@ func SubmitSettings(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	parsedTemplate, err = template.ParseFiles(tplPath)
+	if err != nil {
+		log.WithError(err).Errorf("err while parsing template - %v", tplPath)
+		syscall.Exit(1)
```
Lagovas: what about to return error object and handle it on up level? then we can easily test it if need. otherwise it is non-testable function

cmd/acra_configui/acra_configui.go
```diff
@@ -172,12 +210,13 @@ func main() {
...
+	parseTemplate()
+	configParamsBytes = parseConfig()
 	http.HandleFunc("/index.html", index)
 	http.HandleFunc("/", index)
-	http.Handle("/static/", http.StripPrefix("/static/", http.FileServer(http.Dir("./static"))))
+	http.Handle("/static/", http.StripPrefix("/static/", http.FileServer(http.Dir(*staticPath))))
 	http.HandleFunc("/acraserver/submit_setting", SubmitSettings)
-	log.Info(fmt.Sprintf("AcraConfigUI is listening @ :%d with PID %d", *port, os.Getpid()))
-	err := http.ListenAndServe(fmt.Sprintf(":%d", *port), nil)
+	log.Info(fmt.Sprintf("AcraConfigUI is listening @ %s:%d with PID %d", *host, *port, os.Getpid()))
```
Lagovas: you can use just `log.Infof` with placeholders without `fmt.Sprintf`

cmd/acraserver/acraserver.go
```diff
@@ -178,8 +188,96 @@ func main() {
...
+		if os.Getenv(GRACEFUL_ENV) == "true" {
+			go server.StartCommandsFromFileDescriptor(DESCRIPTOR_API)
+		} else {
+			go server.StartCommands()
+		}
+
+	}
+
+	log.Infof("PID: %v", os.Getpid())
+	if os.Getenv(GRACEFUL_ENV) == "true" {
```
Lagovas: what about to merge this block with if block above (where used os.Getenv condition)?

cmd/acraserver/listener.go
```diff
@@ -15,35 +15,73 @@ package main
...
-func (server *SServer) Close() error {
+func (server *SServer) Close() {
+	log.Debugln("Closing listeners")
 	var err error
 	for _, listener := range server.listeners {
-		if err_ := listener.Close(); err_ != nil && err == nil {
-			err = err_
+		switch listener.(type) {
+		case *net.TCPListener:
+			listener.(*net.TCPListener).Close()
```
Lagovas: `err = listener....` ?

cmd/acraserver/listener.go
```diff
@@ -15,35 +15,73 @@ package main
...
+		case *net.UnixListener:
+			err = listener.(*net.UnixListener).Close()
+			if err != nil {
+				log.WithError(err).Infoln("UnixListener.Close()")
+				continue
+			}
+			// TODO: find better way to remove unixsocket file
+			f, err1 :=listener.(*net.UnixListener).File()
```
Lagovas: you can assign like `_, err1 := ...`. then you don't need to assign `_ = f`
why here used new variable `err1`? if just to log then we can use such block:
```
if _, err := listener.(*net.UnixListener).File(); err != nil {
  log.WithError(err).Errorln("blabla");
}
```
this construction will create new local variable `err` only for `if` block: https://play.golang.org/p/pOw7Tvhzj7t
And will not override outer `err` variable (if you need it)

cmd/acraserver/listener.go
```diff
@@ -15,35 +15,73 @@ package main
...
+			_ = f
+			if err1 != nil {
+				log.WithError(err).Warningln("UnixListener.Close File()")
```
Lagovas: why not log.WithError(err**1**)? We will not know anything about error message from `err1`.

cmd/acraserver/listener.go
```diff
@@ -15,35 +15,73 @@ package main
...
+			}
+			url, err2 := url_.Parse(server.config.GetAcraConnectionString())
+			if err2 != nil {
+				log.WithError(err).Warningln("UnixListener.Close  url_.Parse")
```
Lagovas: log err2?

cmd/acraserver/listener.go
```diff
@@ -15,35 +15,73 @@ package main
...
+			}
+			err3 := os.Remove(url.Path)
+			if err3 != nil {
+				log.WithError(err).Warningf("UnixListener.Close  file.Remove(%s)", url.Path)
```
Lagovas: log err3?

cmd/acraserver/listener.go
```diff
@@ -107,30 +145,149 @@ func (server *SServer) handleConnection(connection net.Conn) {
...
+			log.WithError(err).Errorf("can't accept new connection (connection=%v)", connection)
+		}
+		// unix socket and value == '@'
+		if len(connection.RemoteAddr().String()) == 1 {
+			log.Infof("new connection to acraserver: <%v>", connection.LocalAddr())
+		} else {
+			log.Infof("new connection to acraserver: <%v>", connection.RemoteAddr())
+		}
+		go func() {
+			server.cmACRA.Add(1)
```
Lagovas: what mean `1`?

Lagovas: oh, it's incrementing counter. Would be more readable, if the call will look like `server.cmAcra.Incr()`

cmd/acraserver/listener.go
```diff
@@ -107,30 +145,149 @@ func (server *SServer) handleConnection(connection net.Conn) {
 
 // start listening connections from proxy
 func (server *SServer) Start() {
-	listener, err := network.Listen(server.config.GetAcraConnectionString())
+	var connection net.Conn
+	var listener, err = network.Listen(server.config.GetAcraConnectionString())
 	if err != nil {
 		log.WithError(err).Errorln("can't start listen connections")
+		ErrorSignalChannel <- syscall.SIGTERM
```
Lagovas: why not just return? listener is not created and the server hasn't any user connections.
And using global variable inside method looks a little bit strange. If we need it, better to get channel as function param or use struct field initialized somewhere else

cmd/acraserver/listener.go
```diff
@@ -107,30 +145,149 @@ func (server *SServer) handleConnection(connection net.Conn) {
...
+	}
+}
+
+func (server *SServer) StartFromFileDescriptor(fd uintptr) {
+	var connection net.Conn
+	file := os.NewFile(fd, "/tmp/acraserver")
+	listenerFile, err := net.FileListener(file)
+	if err != nil {
+		log.WithError(err).Errorln("can't start listen for file descriptor")
+		ErrorSignalChannel <- syscall.SIGTERM
```
Lagovas: same comments from `Start` method (above)

cmd/acraserver/listener.go
```diff
@@ -107,30 +145,149 @@ func (server *SServer) handleConnection(connection net.Conn) {
...
+			server.handleConnection(connection)
+			server.cmACRA.Done()
+		}()
+	}
+}
+
+func (server *SServer) StopListeners() {
+	var (
+		err  error
+		nerr *net.TCPListener
```
Lagovas: maybe better to name variable as `tcpListener`? when I read first, I thought that it `network error`... and wanted to say that you don't need to declare such variable here because you use it in `if block`. But then found that it not error)

cmd/acraserver/listener.go
```diff
@@ -107,30 +145,149 @@ func (server *SServer) handleConnection(connection net.Conn) {
...
+		ok   bool
+	)
+	if nerr, ok = server.listenerACRA.(*net.TCPListener); ok {
```
Lagovas: name nerr -> tcpListener and below you can call `tcpListener.SetDeadline()` without second type convertation

cmd/acraserver/listener.go
```diff
@@ -107,30 +145,149 @@ func (server *SServer) handleConnection(connection net.Conn) {
...
+		err = server.listenerACRA.(*net.TCPListener).SetDeadline(time.Now())
+		if err != nil {
+			if nerr, ok := err.(net.Error); ok && nerr.Timeout() {
```
Lagovas: will be good to add `else` branch where will be logged `err` too

cmd/acraserver/listener.go
```diff
@@ -107,30 +145,149 @@ func (server *SServer) handleConnection(connection net.Conn) {
...
+				log.WithError(err).Errorln("unable to SetDeadLine of acra-listener")
+			}
+		}
+	}
+	if nerr != nil {
+		log.Warningf("acra-interface assigment failed - %v", nerr)
```
Lagovas: `log.Warningf("acra-interface assigment failed - %v", nerr)` -> `log.WithError(nerr).Warningln("acra-interface assigment failed")`
Then we can easily filter error messages and our comments to such errors in logs.

cmd/acraserver/listener.go
```diff
@@ -107,30 +145,149 @@ func (server *SServer) handleConnection(connection net.Conn) {
...
+	}
+
+	if server.listenerAPI != nil {
+		if nerr, ok = server.listenerAPI.(*net.TCPListener); ok {
```
Lagovas: same as above about assigning listener after type assertion

cmd/acraserver/listener.go
```diff
@@ -107,30 +145,149 @@ func (server *SServer) handleConnection(connection net.Conn) {
...
+			err = server.listenerAPI.(*net.TCPListener).SetDeadline(time.Now())
+			if err != nil {
+				if nerr, ok := err.(net.Error); ok && nerr.Timeout() {
+					log.WithError(err).Errorln("unable to SetDeadLine of api-listener")
+				}
+			}
+		} else {
+			log.Warningf("api-interface assigment failed - %v", nerr)
```
Lagovas: same about logging error objects

cmd/acraserver/acraserver.go
```diff
@@ -178,8 +188,96 @@ func main() {
...
+		os.Exit(1)
+	}
+	go sigHandlerSIGTERM.Register()
+	sigHandlerSIGTERM.AddCallback(func() {
+		log.Infoln("Incoming SIGTERM or SIGINT")
+		// Stop accepting new connections
+		server.StopListeners()
+		// Wait a maximum of N seconds for existing connections to finish
+		err := server.WaitWithTimeout(ACRASERVER_WAIT_TIMEOUT * time.Second)
+		if err == errors.New("timeout") {
```
Lagovas: better to compare with `WaitTimeoutError`

cmd/acraserver/acraserver.go
```diff
@@ -178,8 +188,96 @@ func main() {
...
+		// Fork new process
+		var fork, err = syscall.ForkExec(os.Args[0], os.Args, execSpec)
+		if err != nil {
+			log.Fatalln("Fail to fork", err)
+		}
+		log.Infof("Server forked to PID: %v", fork)
+
+		// Wait a maximum of N seconds for existing connections to finish
+		err = server.WaitWithTimeout(ACRASERVER_WAIT_TIMEOUT * time.Second)
+		if err == errors.New("timeout") {
```
Lagovas: better to compare with `WaitTimeoutError`

cmd/acraserver/listener.go
```diff
@@ -107,30 +145,149 @@ func (server *SServer) handleConnection(connection net.Conn) {
...
+
+func (server *SServer) WaitConnections(duration time.Duration) {
+	log.Infof("waiting for %v connections to complete", server.ConnectionsCounter())
+	server.cmACRA.Wait()
+	if server.listenerAPI != nil {
+		server.cmAPI.Wait()
 	}
 }
 
+var WaitTimeoutError = errors.New("timeout")
```
Lagovas: go has convention to name all errors with prefix `Err` at start (ErrWaitTimeout)

cmd/acraserver/listener.go
Lagovas: do we use or need this function? where it used?

cmd/acraserver/listener.go
```diff
@@ -153,26 +310,77 @@ func (server *SServer) handleCommandsConnection(connection net.Conn) {
 
 // start listening commands connections from proxy
 func (server *SServer) StartCommands() {
-	listener, err := network.Listen(server.config.GetAcraAPIConnectionString())
+	var connection net.Conn
+	var listener, err = network.Listen(server.config.GetAcraAPIConnectionString())
 	if err != nil {
-		log.WithError(err).Errorln("can't start listen command connections")
+		log.WithError(err).Errorln("can't start listen command api connections")
+		ErrorSignalChannel <- syscall.SIGTERM
```
Lagovas: same about global variable and channel
and to not repeat, will be good to check/apply fixes from `Start` method in this method

cmd/constants.go
```diff
@@ -9,4 +9,7 @@ const (
 	DEFAULT_ACRA_PORT                 = 9393
 	DEFAULT_ACRA_API_PORT             = 9090
 	DEFAULT_ACRA_CONNECTION_PROTOCOL  = "tcp"
+	DEFAULT_ACRA_CONFIGUI_HOST        = "127.0.0.1"
+	DEFAULT_ACRA_CONFIGUI_PORT        = 8000
+	DEFAULT_ACRA_CONFIGUI_STATIC      = "./cmd/acra_configui/static"
```
Lagovas: relative path without `./` correctly handled in golang functions that operate with files and paths

cmd/acra_configui/acra_configui.go
mozhmike: As discussed - will move configs/acraserver_config_vars.yaml to cmd/acra_configui/acraserver_config_vars.go as it's internal thing.
TODO: make run-time config

cmd/acraserver/listener.go
mozhmike: Passing signal channels as server's property

cmd/acraserver/listener.go
mozhmike: done

cmd/acra_configui/acra_configui.go
mozhmike: done

cmd/acra_configui/acra_configui.go
mozhmike: done

cmd/acra_configui/acra_configui.go
mozhmike: done

cmd/acra_configui/acra_configui.go
mozhmike: done

cmd/acra_configui/acra_configui.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: refactored to Warningln as there is no err object available

cmd/acraserver/listener.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: done

cmd/acraserver/acraserver.go
mozhmike: done

cmd/acraserver/acraserver.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: done

cmd/acraserver/listener.go
mozhmike: done

cmd/constants.go
mozhmike: done

vixentael:
LGTM

PR #125, opened 2018-03-12
More flexible query analysis

storojs72:


firewall/firewall_test.go
```diff
@@ -0,0 +1,246 @@
...
+		"SELECT EMP_ID, LAST_NAME FROM EMPLOYEE WHERE CITY = 'Seattle' ORDER BY EMP_ID;",
+		"SELECT EMP_ID, LAST_NAME FROM EMPLOYEE, EMPLOYEE_TBL WHERE CITY = 'Seattle' ORDER BY EMP_ID;",
+		"SELECT EMP_ID, LAST_NAME FROM EMPLOYEE WHERE CITY = 'INDIANAPOLIS' ORDER BY EMP_ID asc;",
+		"INSERT INTO Customers (CustomerName, ContactName, Address, City, PostalCode, Country) VALUES ('Cardinal', 'Tom B. Erichsen', 'Skagen 21', 'Stavanger', '4006', 'Norway');",
+		"INSERT INTO Customers (CustomerName, City, Country) VALUES ('Cardinal', 'Stavanger', 'Norway');",
+	}
+
+	blacklistHandler.AddTables([]string{"EMPLOYEE_TBL", "Customers"})
+
+	err := firewall.HandleQuery(testQueries[0])
```
Lagovas: why not iterate over all queries in loop?

firewall/firewall_test.go
```diff
@@ -0,0 +1,246 @@
...
+		t.Fatal(err)
+	}
+
+	testBlacklistTables(t, firewall, blacklistHandler)
+}
+
+func testBlacklistTables(t *testing.T, firewall * Firewall, blacklistHandler * handlers.BlacklistHandler){
+
+	testQueries := []string {
+		"SELECT EMP_ID, LAST_NAME FROM EMPLOYEE_TBL WHERE CITY = 'Seattle' ORDER BY EMP_ID;",
```
Lagovas: can you add tests for queries with aliases? For example:
```
`"SELECT EMP_ID, LAST_NAME FROM EMPLOYEE_TBL AS EMPL_TBL WHERE CITY = 'Seattle' ORDER BY EMP_ID;"`
SELECT EMP_ID, LAST_NAME FROM EMPLOYEE_TBL EMPL_TBL WHERE CITY = 'Seattle' ORDER BY EMP_ID;"`
```
Below I see code `sqlparser.String(table.(*sqlparser.AliasedTableExpr)) == forbiddenTable` and not sure what actually checks - aliases or table names.

firewall/handlers/handlers_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+}
+
+func areEqual(a []string, b []string) bool {
+	if len(a) != len(b){
+		return false
+	}
+
+	for index := 0; index < len(a); index++{
+		if a[index] != b[index]{
```
Lagovas: [strings.EqualFold()](https://golang.org/pkg/strings/#EqualFold)

firewall/handlers/handlers_util.go
```diff
@@ -0,0 +1,26 @@
...
+		}
+	}
+	return result
+
+}
+
+func contains(queries []string, query string) (bool, int) {
+
+	for index, queryFromRange := range queries {
+		if queryFromRange == query {
```
Lagovas: [strings.EqualFold()](https://golang.org/pkg/strings/#EqualFold)

firewall/firewall_test.go
storojs72: This is for clarity. 0, 2, 4 and 5 queries should be blocked by firewall, while 1 and 3 should pass

firewall/firewall_test.go
storojs72: Actually, I used this code as in the tests of sqlparser (https://github.com/xwb1989/sqlparser/blob/master/analyzer_test.go#L125)

storojs72: Hmm, thanks. It seems it doesn't work for aliased tables

firewall/firewall_test.go
Lagovas: what about this?
```
queriesToBlock := []int{0,2,4,5}
for _, i := range queriesToBlock{
  err := firewall.HandleQuery(testQueries[i])
}
```
because when I read, I saw sequence 0,1,2,3,4 and missed that some of them compared as  `==` and some of them `!=` )

firewall/handlers/handlers_util.go
```diff
@@ -1,28 +1,46 @@
 package handlers
 
-import "strings"
+import (
+	"testing"
+	"strings"
+)
 
-func removeDuplicates(input []string) []string {
+func TestUtilities(t *testing.T){
```
Lagovas: maybe test function better to move to *_test.go file?

PR #136, opened 2018-03-27
use firewall in query processing

Lagovas:


decryptor/mysql/error.go
```diff
@@ -0,0 +1,46 @@
...
+		data = make([]byte, 0, 3+len(mysqlError.Message))
+	} else {
+		// 1 byte ERR_PACKET flag + 2 bytes of error code + 6 bytes of state (protocol41) = 9
+		data = make([]byte, 0, 9+len(mysqlError.Message))
+	}
+
+	data = append(data, ERR_PACKET)
+	data = append(data, byte(mysqlError.Code), byte(mysqlError.Code>>8))
+
+	if isProtocol41 {
```
storojs72: What is "Protocol41"?

Lagovas: mysql has several protocol versions and structs depends on version (https://dev.mysql.com/doc/internals/en/packet-ERR_Packet.html). 41 mean 4.1+ version

tests/test.py
```diff
@@ -205,27 +207,51 @@ def socket_path_from_connection_string(connection_string):
...
     for i in BINARIES:
         try:
-            os.remove(i)
+            os.remove(i.name)
         except:
             pass
 
 PROCESS_CALL_TIMEOUT = 120
 
+def get_go_version():
```
vixentael: so this function changes `1.7` to `1.7.0`?

PR #138, opened 2018-03-28
Updating firewall

storojs72:
Add ability to configure firewall from configuration file


firewall/firewall_configuration_provider.go
```diff
@@ -0,0 +1,106 @@
...
+type FirewallConfig struct {
+	Handlers []struct {
+		Handler string
+		Queries []string
+		Tables  []string
+		Rules   []string
+	}
+}
+
+func (firewall *Firewall) SetFirewallConfiguration(configuration []byte) error {
```
Lagovas: what about `firewall.SetFirewallConfiguration` -> `firewall.LoadConfiguration`?

firewall/firewall_configuration_provider.go
```diff
@@ -0,0 +1,106 @@
...
+
+	err := updateFirewall(firewall, configuration)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func updateFirewall(firewall *Firewall, configuration []byte) error {
```
Lagovas: what about `updateFirewall(firewall *Firewall, ...)` -> `func (firewall *Firewall) update(...)`?

firewall/firewall_configuration_provider.go
```diff
@@ -0,0 +1,106 @@
...
+
+	//fmt.Println(firewallConfiguration)
+
+	var firewallCheckers []QueryHandlerInterface
+	for _, handlerConfiguration := range firewallConfiguration.Handlers {
+		switch handlerConfiguration.Handler{
+		case WhitelistConfigStr:
+			whitelistHandler := &handlers.WhitelistHandler{}
+
+			whitelistHandler.AddQueries(handlerConfiguration.Queries)
```
Lagovas: better to check input params in AddQueries and return error here. Then we can drop extra function `testConfigurationSyntax` that in a future will grow with each new handler type or rules

firewall/firewall_configuration_provider.go
```diff
@@ -0,0 +1,106 @@
...
+			whitelistHandler.AddTables(handlerConfiguration.Tables)
+			whitelistHandler.AddRules(handlerConfiguration.Rules)
+
+			firewallCheckers = append(firewallCheckers, whitelistHandler)
```
Lagovas: and then we can call `firewall.AddHandler(whitelistHandler)` without need to extra check at end of loop

firewall/firewall_interfaces.go
```diff
@@ -4,11 +4,20 @@ package firewall
 type QueryHandlerInterface interface {
 	CheckQuery(sqlQuery string) error
 
+	Refresh()
```
Lagovas: what this method should do in handlers?

Lagovas: as I see below it reset state of handler. Maybe better to rename to `Reset()`?

firewall/firewall_test.go
```diff
@@ -109,7 +112,7 @@ func testWhitelistTables(t *testing.T, firewall * Firewall, whitelistHandler * h
 	for _, i := range queryIndexesToBlock {
 		err := firewall.HandleQuery(testQueries[i])
 		if err != nil {
```
Lagovas: we can simplify to
```
if err != handlers.ErrAccessToForbiddenTableWhitelist {
 t.Fatal(err)
```

firewall/firewall_test.go
```diff
@@ -144,7 +147,7 @@ func testWhitelistTables(t *testing.T, firewall * Firewall, whitelistHandler * h
 	//firewall should block this query
 	if err == nil {
 		if err != nil {
```
Lagovas: err == nil here 100%

firewall/firewall_test.go
```diff
@@ -434,11 +435,135 @@ func testBlacklistByRules(t *testing.T, firewall * Firewall, blacklistHandler *
...
 			}
 		} else {
 			t.Fatal(err)
 		}
 	}
+
+
+	testSyntax(t)
 }
+func testSyntax(t *testing.T) {
```
Lagovas: better to rename `TestLoadConfiguration` because when I read first, I thought that we test here correct parsing of yaml format. But we don't need it, because it tested by yaml parser itself

firewall/handlers/blacklist_handler.go
```diff
@@ -24,45 +24,43 @@ func(handler * BlacklistHandler) CheckQuery(query string) error {
 		}
 	}
 
-
 	//Check tables
 	if len(handler.tables) != 0 {
 		parsedQuery, err := sqlparser.Parse(query)
 		if err != nil {
-			return err
+			return errors.New(err.Error() + " | blacklist (tables)")
```
Lagovas: bad pattern to concatenate strings via "+" operator. And bad way to create error in such way. Better to add new type of error. Then we can easily check is it correct error type in unit-tests without string comparison.

firewall/firewall_interfaces.go
```diff
@@ -4,11 +4,20 @@ package firewall
...
 	AddQueries(queries []string)
 	RemoveQueries(queries []string)
 
 	AddTables(tables []string)
 	RemoveTables(tables []string)
+
+	AddRules(rules []string)
+	RemoveRules(rules []string)
+
+	GetActiveQueries() []string
```
Lagovas: Our handler interface grow and grow and in a future will be very big monolitic interface. But at start it was designed like simple abstract interface for different handlers with simple function `CheckQuery` and encapsulated logic in custom handler itself. 
Anyway what mean `Active` in name? Now it's just added queries/tables/rules to handler. So will be enough to name `GetQueries`. Because `active` tell us that handler has some different queries too

firewall/handlers/blacklist_handler.go
```diff
@@ -24,45 +24,43 @@ func(handler * BlacklistHandler) CheckQuery(query string) error {
...
 			for _, forbiddenTable := range handler.tables {
 				if strings.EqualFold(parsedQuery.Table.Name.String(), forbiddenTable) {
-					return ErrAccessToForbiddenTable
+					return ErrAccessToForbiddenTableBlacklist
 				}
 			}
 
 		case *sqlparser.Update:
-
+			return errors.New("not implemented yet")
```
Lagovas: add new error like `ErrNotImplemented` and then in a future we can use it in another handlers

firewall/handlers/handlers_util.go
```diff
@@ -7,8 +7,12 @@ import (
 
 var ErrQueryNotInWhitelist = errors.New("query not in whitelist")
 var ErrQueryInBlacklist = errors.New("query in blacklist")
-var ErrAccessToForbiddenTable = errors.New("query tries to access forbidden table")
-var ErrForbiddenSqlStructure = errors.New("query's structure is forbidden")
+
+var ErrAccessToForbiddenTableBlacklist = errors.New("query tries to access forbidden table | blacklist")
```
Lagovas: no need to add `| blacklist` at end of message for ErrAccessToForbiddenTable**Blacklist** error

firewall/test_firewall_rules
```diff
@@ -0,0 +1,19 @@
+[handler]
```
Lagovas: do we need this file?

firewall/firewall_configuration_provider.go
```diff
@@ -29,65 +32,44 @@ func (firewall *Firewall) SetFirewallConfiguration(configuration string) error {
 	return nil
 }
 
-func updateFirewall(firewall *Firewall, configuration string) error {
+func updateFirewall(firewall *Firewall, configuration []byte) error {
```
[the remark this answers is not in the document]
storojs72: Agree

firewall/firewall_implementation.go
```diff
@@ -27,3 +29,9 @@ func (firewall *Firewall) HandleQuery(query string) error{
 	}
 	return nil
 }
+
+func (firewall *Firewall) PrintStatus() {
```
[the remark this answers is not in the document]
storojs72: This is for debug purposes. Removed

firewall/firewall_configuration_provider.go
```diff
@@ -29,65 +32,44 @@ func (firewall *Firewall) SetFirewallConfiguration(configuration string) error {
...
-			configLineType = queryConfigHeader
-			break;
-		case "[tables]":
-			configLineType = tableConfigHeader
-			break;
-		case "[structures]":
-			configLineType = structureConfigHeader
-			break;
-		case "\n":
+			whitelistHandler.AddQueries(handlerConfiguration.Queries)
```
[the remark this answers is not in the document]
storojs72: If we have separate function to test syntax, we are able to modify it if necessary. And we can't do it if logic of syntax checking will be scattered over multiply functions

firewall/firewall_configuration_provider.go
storojs72: Agree. Changed

firewall/firewall_interfaces.go
storojs72: Agree. Changed

firewall/firewall_test.go
storojs72: Oh. Agree.

firewall/firewall_test.go
storojs72: Agree. Refactored

firewall/firewall_test.go
storojs72: No. yaml parser can't check those error inputs : "SLECT AVG(Price) FROM Products;" or "SELECT * ROM EMPLOYEE WHERE CITY='Seattle';". These syntax errors can be checked only by sqlparser

firewall/handlers/blacklist_handler.go
storojs72: Agree. Refactored

firewall/handlers/blacklist_handler.go
storojs72: Ok

firewall/test_firewall_rules
storojs72: No. We don't.

firewall/firewall_interfaces.go
storojs72: Ok. Renamed

firewall/firewall_configuration_provider.go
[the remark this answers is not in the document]
storojs72: Finally, I agree. It also helps to reduce handler public interface

firewall/handlers/whitelist_handler.go
```diff
@@ -8,124 +8,134 @@ import (
...
+		}
 	}
 
 	handler.rules = removeDuplicates(handler.rules)
+
+	return nil
 }
 
-func (handler *WhitelistHandler) RemoveRules(rules []string){
+func (handler * WhitelistHandler) RemoveRules(rules []string){
```
Lagovas: `gofmt` will not agree with your space between `*` and type :)
please, run `go fmt ./...` in repo folder before pushing or we will not so excellent golang repo as now (https://goreportcard.com/report/github.com/cossacklabs/acra)  :((

firewall/firewall_configuration_provider.go
```diff
@@ -56,7 +56,7 @@ func (firewall *Firewall) update(configuration []byte) error {
 			}
 
 			firewallCheckers = append(firewallCheckers, whitelistHandler)
```
Lagovas: why not add handler here? instead of second loop after this

storojs72: Thanks. Agree

PR #140, opened 2018-04-02
mysql tls support

Lagovas:


cmd/acraserver/acraserver.go
```diff
@@ -169,14 +178,23 @@ func main() {
...
+	var tlsConfig *tls.Config
+	if *useTls || *tlsKey != "" {
+		tlsConfig, err = network.NewTLSConfig(*tlsSNI, *tlsCA, *tlsKey, *tlsCert)
 		if err != nil {
 			log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorTransportConfiguration).
 				Errorln("Configuration error: can't get config for TLS")
 			os.Exit(1)
 		}
+		// need for testing with mysql docker container that always generate new certificates
+		if Test == TEST_MODE {
```
vixentael: Won't it be better to rename `Test` into `SkipVerifyingCertificated`?

Lagovas: thought to name it more common to use in future tests if need

vixentael: what about 'IsRunningTest'? 'TestOnly'?

vixentael: by the way, why do you check 
`if Test == TEST_MODE`

instead of `if Test == True`?

Lagovas: on compile time we can change only string values

Lagovas: will rename to TestOnly

PR #143, opened 2018-04-03
encrypt private keys

Lagovas:
added encryption of private (only, public keys stored as is) keys via keystore. Now it load master key from env var `ACRA_MASTER_KEY` and it is requirement (acraserver/acraproxy must start with correct base64 value in this emv variable)


Lagovas:
I will add check for key length (32 bytes) and then will merge to avoid pass bad key by users

PR #145, opened 2018-04-04
Decrease amount of connection errors during tests

vixentael:
This workaround possibly decreases amount of unstable connection errors during integration tests.

I can't be sure that it will help to eliminate all connection issues, but according to my last N tests, it reduces connection errors on CircleCI significantly (I got no connection errors during testing in my branch).

However, due to delay between tests, overall time for testing increases in ~1.5x.

I think we should merge this workaround to master branch. However, if we won't get positive results, just put `0` as value of `BETWEEN_TESTS_SLEEP` to remove extra delays.


vixentael:
@Lagovas will implement idea of closing cursor better than me. PR will come!

PR #149, opened 2018-04-05
check tls config before handshake

Lagovas:
now it's panic with nil dereference when you didn't pass tls path to key/cert/ca cert and when you initialize tls connection with db
this PR log error message about it and tell that you need to provide such files if you want to run acraserver with tls support app<->db


decryptor/mysql/response_proxy.go
```diff
@@ -174,6 +175,20 @@ func (handler *MysqlHandler) ClientToDbProxy(errCh chan<- error) {
 			firstPacket = false
 			handler.clientProtocol41 = packet.ClientSupportProtocol41()
 			if packet.IsSSLRequest() {
+				if handler.tlsConfig == nil {
+					log.Errorln("To support TLS connections you must pass TLS key and certificate for AcraServer that will be used" +
```
shadinua: Looks like the space was missed at the end of this line (and below in the same block).

vixentael:
Is there any easy way how we can test it?

tests/test.py
```diff
@@ -1548,8 +1548,7 @@ def test_only_alpha_client_id(self):
         # call with directory separator in key name
         self.assertEqual(create_client_keypair(POISON_KEY_PATH), 1)
 
-
-class SSLPostgresqlConnectionTest(AcraCatchLogsMixin, HexFormatTest):
+class SSLPostgresqlMixin(AcraCatchLogsMixin):
     ACRA2_PORT = BaseTestCase.ACRA_PORT+1000
```
storojs72: Why did you decide to plus 1000? Why not 2000 or 10000?

Lagovas: chosen manually. we ran tests on ~5-10 go versions and for each we use ACRA_PORT+1. Thought that 100 may have intersection with any PROXY_PORT

storojs72: There is no probability that port will be greater than max value - 65535?

Lagovas: No, because it may be overrided with TEST_ACRA_PORT env variable that in tests use 10003 port as default

PR #150, opened 2018-04-05
retry connect on failure

Lagovas:
added retrying to handle cases when db not up yet or overloaded in test environment


vixentael:
Hope it helps 🤞

PR #151, opened 2018-04-06
Add logger-handler to firewall

storojs72:


cmd/acraserver/config.go
```diff
@@ -70,8 +71,26 @@ func NewConfig() *Config {
 
 var ErrTwoDBSetup = errors.New("only one db supported at one time")
 
-func (config *Config) SetFirewall(fw firewall.FirewallInterface) {
-	config.firewall = fw
+func (config *Config) SetFirewall(censorConfigPath string) error {
+	firewall := &firewall.Firewall{}
```
Lagovas: better to assign one time `config.firewall = &firewall.Firewall{}` instead 5 extra lines

cmd/acraserver/acraserver.go
```diff
@@ -148,6 +149,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := config.SetFirewall(*firewallConfig); err != nil {
```
vixentael: I think we shouldn't return error if no config file.
users should be able not to start firewall if they don't want to.

i would check if `firewallConfig` exists first, then `setFirewall`

firewall/handlers/logging_handler.go
```diff
@@ -0,0 +1,89 @@
...
+	return queries
+}
+
+func (handler *LoggingHandler) MarkQueryAsForbidden(query string) {
+	for index, queryInfo := range handler.Queries {
+		if strings.EqualFold(query, queryInfo.RawQuery) {
+			handler.Queries[index].IsForbidden = true
+		}
+	}
+	handler.Serialize()
```
Lagovas: what if error?

firewall/handlers/logging_handler.go
```diff
@@ -0,0 +1,89 @@
...
+		if queryInfo.IsForbidden == true{
+			forbiddenQueries = append(forbiddenQueries, queryInfo.RawQuery)
+		}
+	}
+	return forbiddenQueries
+}
+
+func (handler *LoggingHandler) Serialize() error {
+	jsonFile, err := json.Marshal(handler.Queries)
+	err = ioutil.WriteFile(handler.filePath, jsonFile, 0600)
```
Lagovas: you can return result because it err itself

firewall/handlers/logging_handler.go
```diff
@@ -0,0 +1,89 @@
...
+	return nil
+}
+
+func (handler *LoggingHandler) Deserialize() error {
+	var bufferBytes []byte
+	bufferBytes, err := ioutil.ReadFile(handler.filePath)
+	if err != nil {
+		return err
+	}
+	json.Unmarshal(bufferBytes, &handler.Queries)
```
Lagovas: `err := json.Unmarshal(...)`

firewall/handlers/logging_handler.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
```
Lagovas: here you check ReadFile call

vixentael:
Please add info log to see if query was allowed/filtered here:
https://github.com/cossacklabs/acra/blob/master/firewall/firewall_implementation.go#L23

And show example of json output

acracensor/acracensor_implementation.go
```diff
@@ -0,0 +1,31 @@
...
+func (acraCensor *AcraCensor) RemoveHandler(handler QueryHandlerInterface) {
+	for index, handlerFromRange := range acraCensor.handlers {
+		if handlerFromRange == handler {
+			acraCensor.handlers = append(acraCensor.handlers[:index], acraCensor.handlers[index+1:]...)
+		}
+	}
+}
+
+func (acraCensor *AcraCensor) HandleQuery(query string) error {
+	log.Infof("Censor works")
```
vixentael: we will receive this log message on every query. which is not very useful, i'd remove it

acracensor/acracensor_implementation.go
```diff
@@ -0,0 +1,31 @@
...
+	for _, handler := range acraCensor.handlers {
+		log.Infof("Handler: %s", handler.GetName())
+		if err := handler.CheckQuery(query); err != nil {
+			return err
```
vixentael: `log.WithFields(log.Fields{"acracensor" : "forbidden query", "handler" : handler.GetName()}).Infof("Query: %s", query)`

acracensor/acracensor_implementation.go
```diff
@@ -0,0 +1,31 @@
...
+		if handlerFromRange == handler {
+			acraCensor.handlers = append(acraCensor.handlers[:index], acraCensor.handlers[index+1:]...)
+		}
+	}
+}
+
+func (acraCensor *AcraCensor) HandleQuery(query string) error {
+	log.Infof("Censor works")
+	for _, handler := range acraCensor.handlers {
+		log.Infof("Handler: %s", handler.GetName())
```
vixentael: i think we should remove this

acracensor/acracensor_implementation.go
```diff
@@ -0,0 +1,31 @@
...
+		if err := handler.CheckQuery(query); err != nil {
+			return err
+		}
```
vixentael: `log.WithFields(log.Fields{"acracensor" : handler.GetName()}).Infof("Allowed query: %s", query)`

cmd/acraserver/config.go
```diff
@@ -50,7 +51,7 @@ type Config struct {
 	postgresql              bool
 	configPath              string
 	debug                   bool
-	firewall                firewall.FirewallInterface
+	firewall                acracensor.AcracensorInterface
```
vixentael: `censor` instead of `firewall`?

logging/event_codes.go
```diff
@@ -52,8 +52,9 @@ const (
 	EventCodeErrorCantParseAuthData       = 557
 	EventCodeErrorCantDumpConfig          = 558
 
-	// firewall
+	// acracensor
 	EventCodeErrorFirewallQueryIsNotAllowed = 560
```
vixentael: censor instead of firewall?

storojs72: Agree

cmd/acraserver/config.go
storojs72: Agree

cmd/acraserver/config.go
```diff
@@ -70,11 +71,26 @@ func NewConfig() *Config {
...
+	}
+	configuration, err := ioutil.ReadFile(censorConfigPath)
+	if err != nil {
+		return err
+	}
+	err = acraCensor.LoadConfiguration(configuration)
+	if err != nil {
+		return err
+	}
+	config.censor = acraCensor
```
Lagovas: unnecessary

acracensor/acracensor_implementation.go
```diff
@@ -0,0 +1,30 @@
...
+		if handlerFromRange == handler {
+			acraCensor.handlers = append(acraCensor.handlers[:index], acraCensor.handlers[index+1:]...)
+		}
+	}
+}
+
+func (acraCensor *AcraCensor) HandleQuery(query string) error {
+	for _, handler := range acraCensor.handlers {
+		if err := handler.CheckQuery(query); err != nil {
+			return err
```
Lagovas: log error. will be good to log handler name too

PR #153, opened 2018-04-11
Fix for acra-configui auth_off mode

mozhmike:


cmd/acra_configui/acra_configui.go
```diff
@@ -399,9 +402,14 @@ func main() {
 		os.Exit(1)
 	}
 
-	err = loadAuthData()
-	if err != nil {
-		os.Exit(1)
+	if *authMode == "auth_off" {
+		log.Info("HTTP Basic Auth is turned off")
```
vixentael: warning?

mozhmike: true

cmd/acra_configui/acra_configui.go
```diff
@@ -399,9 +402,14 @@ func main() {
 		os.Exit(1)
```
Lagovas: print about error for user

cmd/acra_configui/acra_configui.go
```diff
@@ -399,9 +402,14 @@ func main() {
...
-	err = loadAuthData()
-	if err != nil {
-		os.Exit(1)
+	if *authMode == "auth_off" {
+		log.Info("HTTP Basic Auth is turned off")
+	} else {
+		log.Infof("HTTP Basic Auth mode: %v", *authMode)
+		err = loadAuthData()
+		if err != nil {
+			os.Exit(1)
```
Lagovas: print something about error

PR #161, opened 2018-04-13
Acra ConfigUI integration tests

mozhmike:
Tests for:
- set, remove basic auth user with acra_genauth
- basic auth (valid and wrong credentials) at configui's web
- get acraserver's settings
- update acraserver's setttings and apply after graceful restart


tests/test.py
```diff
@@ -1540,6 +1597,77 @@ def test_only_alpha_client_id(self):
...
+        try:
+            os.unlink('configs/acraserver.yaml')
+        except Exception as e:
+            print(e)
+        stop_process([self.configui])
+        args = ['killall', '--signal=SIGTERM', 'acraserver']
+        try:
+            subprocess.call(args, cwd=os.getcwd(), timeout=PROCESS_CALL_TIMEOUT)
+        except Exception as e:
+            print(e)
```
Lagovas: on error send sigkill signal to stop process at all because sigterm may not finished and may be freezed

mozhmike: done

PR #167, opened 2018-04-24
check netstat/awk dependencies

Lagovas:
check that system has installed netstat\awk that acraproxy uses to check that connection come from another user on localhost network
now it incorrectly handle situation when netstat not installed


Lagovas:
is it common case and it really need to be handled? they may do it like `PATH=$PATH:/path/to/dir/with/tool go run cmd/acraproxy/acraproxy.go`.

shadinua:
Of course, simply adding paths to `$PATH` when starting the command will solve this. I have mentioned only rare cases when adding an entire directory to `$PATH` may be unacceptable. I think we can keep the current code.

PR #169, opened 2018-05-03
renaming: acrawriter -> acra-writer / AcraWriter & acraserver -> acra-server / AcraServer

vixentael:
Continue in renaming.


cmd/acra-connector/acra-connector.go
```diff
@@ -149,23 +149,23 @@ func main() {
 	keysDir := flag.String("keys_dir", keystore.DEFAULT_KEY_DIR_SHORT, "Folder from which will be loaded keys")
 	clientId := flag.String("client_id", "", "Client id")
 	acraHost := flag.String("acra_host", "", "IP or domain to acra daemon")
-	acraCommandsPort := flag.Int("acra_commands_port", cmd.DEFAULT_ACRA_API_PORT, "Port of acra http api")
+	acraCommandsPort := flag.Int("acra_commands_port", cmd.DEFAULT_ACRA_API_PORT, "Port of Acra HTTP api")
```
shadinua: Here and further we use the term `Acra port` for different instances: AcraServer and AcraConnector. It would be good to explicitly specify application. In this line, I think, should be AcraServer.

cmd/acra-connector/acra-connector.go
```diff
@@ -149,23 +149,23 @@ func main() {
...
 	acraPort := flag.Int("acra_port", cmd.DEFAULT_ACRA_PORT, "Port of acra daemon")
-	acraId := flag.String("acra_id", "acra_server", "Expected id from acraserver for Secure Session")
+	acraId := flag.String("acra_id", "acra_server", "Expected id from AcraServer for Secure Session")
 	verbose := flag.Bool("v", false, "Log to stderr")
 	port := flag.Int("port", cmd.DEFAULT_CONNECTOR_PORT, "Port fo acra-connector")
```
shadinua: Please, correct: "Port `to` AcraConnector".
Also, I suggest to use the same way to describe components in messages: `AcraServer` or `acra-server`. If we're talking about application, let's use `AcraServer`. In case we mean the package or binary file, let's use `acra-server`.

cmd/acra-connector/acra-connector.go
```diff
@@ -149,23 +149,23 @@ func main() {
 	keysDir := flag.String("keys_dir", keystore.DEFAULT_KEY_DIR_SHORT, "Folder from which will be loaded keys")
 	clientId := flag.String("client_id", "", "Client id")
 	acraHost := flag.String("acra_host", "", "IP or domain to acra daemon")
-	acraCommandsPort := flag.Int("acra_commands_port", cmd.DEFAULT_ACRA_API_PORT, "Port of acra http api")
+	acraCommandsPort := flag.Int("acra_commands_port", cmd.DEFAULT_ACRA_API_PORT, "Port of Acra HTTP api")
 	acraPort := flag.Int("acra_port", cmd.DEFAULT_ACRA_PORT, "Port of acra daemon")
```
shadinua: Probably, we can rename `cmd.DEFAULT_ACRA_PORT` → `cmd.DEFAULT_ACRASERVER_PORT` or `cmd.DEFAULT_SERVER_PORT` and so on?

vixentael: renamed constants, see in last commit

vixentael: we will rename arguments a bit later in a separate chunk of refactoring

PR #171, opened 2018-05-04
add auth mode param

Lagovas:
add settings that change tls authentication mode. in AcraConnector it set how it will authenticate AcraServer in tls connection, on AcraServer side it setup how it will authenticate database in tls connection
plus updated tls configuration to provide other less strict authentication methods (not authenticate client from server, require any certificate, require and check) as it implemented in other places (for examples in dbs)

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -1255,7 +1255,7 @@ class AcraCatchLogsMixin(object):
         with open(self.log_files[process].name, 'r', errors='replace',
                   encoding='utf-8') as f:
             log = f.read()
-            print(log)
+            print(log.encode('utf-8'))
             return log
 
     def fork_acra(self, popen_kwargs: dict=None, **acra_kwargs: dict):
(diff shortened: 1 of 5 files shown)


tests/test.py
```diff
@@ -429,6 +428,11 @@ def fork_configui(self, proxy_port: int, http_port: int):
         process = self.fork(lambda: subprocess.Popen(args))
         return process
 
+    def get_connector_tls_params(self):
+        return [
+            '--tls',
+            '--tls_sni=acraserver',
```
vixentael: we should not forget to rename this to `acra-server` in `rename` branch

Lagovas: it doesn't matter here. at first - it is tests. and it's depends on values in certificate that used in tests

vixentael: oh right, it's a value inside certificate 🙈

PR #172, opened 2018-05-04
Add goroutine to censor for background logging 

storojs72:


acracensor/acracensor_interfaces.go
```diff
@@ -3,7 +3,7 @@ package acracensor
 type QueryHandlerInterface interface {
 	CheckQuery(sqlQuery string) error
 	Reset()
-	GetName() string
+	//GetName() string
```
vixentael: remove?

storojs72: Yeah. It is redundant method - censor doesn't use it

acracensor/handlers/querycapture_handler.go
```diff
@@ -0,0 +1,153 @@
...
+	return queries
+}
+
+func (handler *QueryCaptureHandler) MarkQueryAsForbidden(query string) error {
+	for index, queryInfo := range handler.Queries {
+		if strings.EqualFold(query, queryInfo.RawQuery) {
+			handler.Queries[index].IsForbidden = true
+		}
+	}
+	return handler.Serialize()
```
Lagovas: serialization better to do in separate goroutine/thread

acracensor/handlers/querycapture_handler.go
```diff
@@ -0,0 +1,153 @@
...
+	queryInfo.IsForbidden = false
+	handler.Queries = append(handler.Queries, *queryInfo)
+
+	handler.logChannel <- *queryInfo
+
+	return nil
+}
+
+func (handler *QueryCaptureHandler) Reset() {
+	handler.Queries = nil
```
Lagovas: reset `err` too?

acracensor/handlers/querycapture_handler.go
```diff
@@ -5,89 +5,127 @@ import (
...
-
 func NewQueryCaptureHandler(filePath string) (*QueryCaptureHandler, error) {
-	file, err := os.OpenFile(filePath, os.O_RDONLY|os.O_CREATE, 0600)
-	if err != nil {
-		return nil, err
+
+	var _, err = os.Stat(filePath)
+
+	// create file if not exists
+	if os.IsNotExist(err) {
```
Lagovas: you don't check other errors than IsNotExist. As example user hasn't permission to this file, or device is busy or anything else
```
if err != nil {
   if os.IsNotExist(err) {
   } else {
      // handle other errors
   }
}
```

acracensor/handlers/querycapture_handler.go
```diff
@@ -5,89 +5,127 @@ import (
...
+	signalToSerialize := make(chan bool)
+	signalShutdown := make(chan os.Signal, 2)
+	signal.Notify(signalShutdown, os.Interrupt, syscall.SIGTERM)
+
+	//serialization signal
+	go func() {
+		for {
+			time.Sleep(TimeoutSecondsToSerialize * time.Second)
+			//timer finished. Close channel to inform that serialization should be performed
+			signalToSerialize <- true
```
Lagovas: each goroutine with infinite loop should give possibility to stop it. for example when we will tests and use at start this handler then this goroutine will leave whole time of test even when it doesn't needed. If you will initialize 100 time in tests this handler then during test will leave 100 background goroutines. Better to add some way to stop it, for example by `Reset()` method:
```
go func () {
  for {
    select {
      case <- time.NewTimer(TimeoutSecondsToSerialize * time.Second).C:
          signalToSerialize <- true
      case <- signalShutdown:
            return
    }
  }
}
```

acracensor/handlers/querycapture_handler.go
```diff
@@ -5,89 +5,127 @@ import (
...
+	handler.signalToSerialize = signalToSerialize
+	handler.logChannel = logChannel
+
+	//handling goroutine
 	go func (){
 		for {
 			select {
-			case queryInfo, gotQuery := <-logChannel_:
-				if gotQuery {
+			case <-signalToSerialize:
```
Lagovas: we don't need second goroutine to take signal to serialize by some interval. Use `<- NewTimer().C`
Plus as above, add way to stop this infinite goroutine too by processing signal to stop processing.

acracensor/handlers/querycapture_handler.go
```diff
@@ -5,89 +5,127 @@ import (
...
-						occuredError.query = &queryInfo
-						occuredError.err = err
+						log.WithError(ErrSingleQueryCaptureError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSecurityError)
 					}
 
 					f, err := os.OpenFile(filePath, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
 					if err != nil {
-						occuredError.query = &queryInfo
-						occuredError.err = err
+						log.WithError(ErrSingleQueryCaptureError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSecurityError)
```
Lagovas: same there. we don't interrupt processing and in next commands we will try to write to not-existed file.

acracensor/handlers/querycapture_handler.go
```diff
@@ -5,89 +5,127 @@ import (
...
-					//channel is closed
-					return
+					//channel is unexpectedly closed
+					log.WithError(ErrUnexpectedCaptureChannelClose).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSecurityError)
+				}
+			case <-signalShutdown:
+				err := handler.Serialize()
+				if err != nil {
+					log.WithError(ErrComplexSerializationError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSecurityError)
 				}
```
Lagovas: add return to stop goroutine

storojs72: Do we need to return? As we discussed with @shadinua and @vixentael if error in background occurred, we should inform about this and continue working. Am I missing something?

acracensor/handlers/querycapture_handler.go
```diff
@@ -5,89 +5,127 @@ import (
...
 
+	logChannel := make(chan QueryInfo, MaxQueriesInChannel)
+	signalToSerialize := make(chan bool)
+	signalShutdown := make(chan os.Signal, 2)
+	signal.Notify(signalShutdown, os.Interrupt, syscall.SIGTERM)
+
+	//serialization signal
+	go func() {
+		for {
+			time.Sleep(TimeoutSecondsToSerialize * time.Second)
```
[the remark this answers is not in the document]
storojs72: If I add timeout parameter to constructor, It will be necessary to add this parameter to yaml config. Is it good? May be it is ok to initialise it with default value (1 second) and give possibility to change via getter/setter?

acracensor/handlers/querycapture_handler.go
Lagovas: it was releated to standard processing (`case queryInfo, ok := <-handler.logChannel`). That when you took any errors, you should continue with some delay. But you should add some way to stop this goroutine. For example for tests, or restarting (for example on changes from webconfig), or something else.

acracensor/handlers/querycapture_handler.go
```diff
@@ -192,7 +196,13 @@ func (handler *QueryCaptureHandler) GetForbiddenQueries() []string{
 	}
 	return forbiddenQueries
 }
+func (handler *QueryCaptureHandler) SetSerializationTimeout(timeoutInSeconds time.Duration){
+	handler.serializationTimeoutSeconds = timeoutInSeconds
```
Lagovas: Better to expect time.Duration without assumptions of type of value like second/minute/hour. We expect duration and use it as is. If user will want to use seconds, he will pass `time.Second * 2`, or `time.Millisecond * 50`. And we will pass it as is to `time.Sleep`. 
For me timeoutInSeconds tells that I should pass value correct "timeout" as seconds in golang - `time.Second * myValue`. But if we will look above, we expect value without converting and convert it itself by `time.Sleep(handler.serializationTimeoutSeconds * time.Second)`

acracensor/acracensor_test.go
```diff
@@ -479,8 +479,9 @@ func TestSerialization(t *testing.T){
 		}
 	}
 
+	loggingHandler.SetSerializationTimeout(1)
```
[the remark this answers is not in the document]
Lagovas: or time.Millisecond*50 after comments below about duration and converting time.

acracensor/handlers/querycapture_handler.go
```diff
@@ -0,0 +1,216 @@
...
+			//timer finished. Close channel to inform that serialization should be performed
+			signalToSerialize <- true
+		}
+	}()
+
+	//handling goroutine
+	go func (){
+		for {
+			select {
+			case <-signalToSerialize:
```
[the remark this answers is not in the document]
storojs72: I tried as you proposed (`<- time.NewTimer(...).C`) but it didn't work as expected. Test ran infinitely, so I left this unchanged

acracensor/handlers/querycapture_handler.go
```diff
@@ -85,11 +85,15 @@ func NewQueryCaptureHandler(filePath string) (*QueryCaptureHandler, error) {
...
 			}
 			time.Sleep(handler.serializationTimeout)
-			//timer finished. Close channel to inform that serialization should be performed
 			signalToSerialize <- true
 		}
 	}()
 
+	f, err := os.OpenFile(filePath, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
+	if err != nil {
+		log.WithError(ErrSingleQueryCaptureError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSecurityError)
```
Lagovas: what if you can't open file? here we should return error and stop processing. this handler initialized at start of program and should be handled by user before any requests.

acracensor/handlers/querycapture_handler.go
```diff
@@ -156,7 +156,12 @@ func (handler *QueryCaptureHandler) CheckQuery(query string) error {
 	queryInfo.IsForbidden = false
 	handler.Queries = append(handler.Queries, *queryInfo)
 
-	handler.logChannel <- *queryInfo
+	select {
```
Lagovas: wow good, I didn't know about such construction to avoid block of write to channel )

acracensor/handlers/querycapture_handler.go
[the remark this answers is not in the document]
Lagovas: it will works as work now. your goroutine above do the same - infinitely send signal to `signalToSerialize` and stop when took `signalShutdown`
When you will use `NewTimer`, you will infinitely receive signal to serialize at constant interval and will stop on `signalShutdown` (this signal we handle here too)
https://play.golang.org/p/l8uYMdPePbC - example
This is last change for it PR that we need and will simplify processing. Replace signalToSerialize to Ticker

PR #178, opened 2018-05-25
test stable versions of golang

Lagovas:


Lagovas:
was fixed in golang repo, we doesn't need it more

PR #179, opened 2018-05-25
log errors in api handler

Lagovas:


cmd/acra-server/client_commands_session.go
```diff
@@ -85,9 +87,13 @@ func (clientSession *ClientCommandsSession) HandleSession() {
 	case "/getNewZone":
 		log.Debugln("Got /getNewZone request")
 		id, publicKey, err := clientSession.keystorage.GenerateZoneKey()
-		if err == nil {
+		if err != nil {
+			log.WithError(err).Errorln("can't generate zone key")
```
vixentael: I would start logs from capital letter.

vixentael: Can we create separate error code here? Or you think it's not useful?

PR #184, opened 2018-05-30
Basic integration test for censor

storojs72:


acra-censor/acra-censor_configuration_provider.go
```diff
@@ -58,16 +59,23 @@ func (acraCensor *AcraCensor) LoadConfiguration(configuration []byte) ([]QueryHa
...
 			}
-			logger, err := handlers.NewQueryCaptureHandler(handlerConfiguration.Filepath)
+			queryCaptureHandler, err := handlers.NewQueryCaptureHandler(handlerConfiguration.Filepath)
 			if err != nil {
 				return nil, err
 			}
-			acraCensor.AddHandler(logger)
-			handlers_ = append(handlers_, logger)
+			acraCensor.AddHandler(queryCaptureHandler)
+			handlers_ = append(handlers_, queryCaptureHandler)
```
Lagovas: do we really need second slice with all handlers? I found one place when it used - in tests. And used to iterate and check length of slice. But in tests we have access to private fields if tests inside the same package. So we don't need this return value, imo

acra-censor/handlers/whitelist_handler.go
```diff
@@ -82,6 +84,10 @@ func (handler *WhitelistHandler) Release() {
 	handler.Reset()
 }
 
+func (handler *WhitelistHandler) Priority() int {
+	return 2
```
Lagovas: imo this value shouldn't be hardcoded and should be configurable by user

configs/acra-censor.example.yaml
```diff
@@ -10,6 +15,6 @@ handlers:
       - Customers
     rules:
       - SELECT * FROM EMPLOYEE WHERE CITY='Seattle';
-  - handler: whitelist
-    tables:
-      - EMPLOYEE
\ No newline at end of file
+#  - handler: whitelist
```
Lagovas: drop it?

acra-censor/acra-censor_implementation.go
```diff
@@ -10,6 +11,9 @@ type AcraCensor struct {
 
 func (acraCensor *AcraCensor) AddHandler(handler QueryHandlerInterface) {
 	acraCensor.handlers = append(acraCensor.handlers, handler)
+	sort.Slice(acraCensor.handlers, func(i, j int) bool {
```
Lagovas: as we discussed, priority should be fetched from order in configuration. if rule is on top of list then it has higher priority than next rules. or we should change rule structure and add priority field to provide ability to configure it by user
```
handlers:
  - handler: query_ignore
    priority: 1
    queries:
      - ROLLBACK
      - COMMIT
      - BEGIN
  - handler: query_capture
    priority: 3
    filepath: censor_log
  - handler: blacklist
    priority: 2
    queries:
      - INSERT INTO SalesStaff1 VALUES (1, 'Stephen', 'Jiang');
```
if without priority field, than `query_ignore` than rules should executed in next order `query_ignore` -> `query_capture` -> `blacklist`. If user want to run `blacklist` first then he should place it first in config.

acra-censor/acra-censor_implementation.go
```diff
@@ -27,19 +34,18 @@ func (acraCensor *AcraCensor) ReleaseAll() {
...
-		for _, handler := range acraCensor.handlers {
-			if err := handler.CheckQuery(query); err != nil {
-				log.Errorf("Forbidden query: '%s'", query)
-				return err
+	for _, handler := range acraCensor.handlers {
+		continueHandling, err := handler.CheckQuery(query)
+		if err != nil {
+			log.Errorf("Forbidden query: '%s'", query)
+			return err
+		} else{
```
Lagovas: `else if !continueHandling`
plus run gofmt before pushing

acra-censor/acra-censor_test.go
```diff
@@ -712,12 +783,16 @@ func TestConfigurationProvider(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	if len(handlers_) != 3{
```
Lagovas: `acraCensor.handlers` and then we don't need to return all handlers just for tests

acra-censor/handlers/queryignore_handler.go
```diff
@@ -0,0 +1,44 @@
...
+
+	if handler.ignoredQueries[query]{
+		//do not continue query handling
+		return false, nil
+	}
+	return true, nil
+}
+
+func (handler *QueryIgnoreHandler) Reset() {
+	handler.ignoredQueries = nil
```
Lagovas: `handler.ignoredQueries = make(map[string]bool)` will be better to avoid nil dereference

acra-censor/handlers/blacklist_handler.go
```diff
@@ -25,36 +29,50 @@ func (handler *BlacklistHandler) CheckQuery(query string) (bool, error) {
...
 		}
-
 		switch parsedQuery := parsedQuery.(type) {
 		case *sqlparser.Select:
 			for _, forbiddenTable := range handler.tables {
-				for _, table := range parsedQuery.From {
-					if strings.EqualFold(sqlparser.String(table.(*sqlparser.AliasedTableExpr).Expr), forbiddenTable) {
-						return false, ErrAccessToForbiddenTableBlacklist
+				for _, fromStatement := range parsedQuery.From {
+					_, ok := fromStatement.(*sqlparser.AliasedTableExpr)
```
Lagovas: you can use switch statement to handle several types:
```
switch fromStatement.(type){
case *sqlparser.JoinTableExpr:
  break
case *sqlparser.JoinTableExpr:
  break
default:
  break
}
```

acra-censor/handlers/blacklist_handler.go
```diff
@@ -245,4 +243,31 @@ func (handler *BlacklistHandler) isForbiddenColumnAccess(columnsToEvaluate sqlpa
 		}
 	}
 	return false
-}
\ No newline at end of file
+}
+
+//
+//func (handler *BlacklistHandler) handleParenTable(fromStatement *sqlparser.ParenTableExpr, forbiddenTable string) (bool, error){
```
Lagovas: drop it?

acra-censor/handlers/queryignore_handler.go
```diff
@@ -2,43 +2,48 @@ package handlers
 
 type QueryIgnoreHandler struct {
 	ignoredQueries map[string]bool
+	priority       int
 }
 
 func NewQueryIgnoreHandler() *QueryIgnoreHandler {
 	handler := &QueryIgnoreHandler{}
 	handler.ignoredQueries = make(map[string]bool)
+	handler.priority = 2
```
Lagovas: can we avoid hardcoded priorities?

tests/test.py
```diff
@@ -750,23 +750,53 @@ def testReadAcrastructInAcrastruct(self):
...
+                result = self.engine1.execute(sa.text("select data_raw from test"))
+
+            with self.assertRaises(sa.exc.ProgrammingError):
+                result = self.engine1.execute(sa.text("select * from acrarollback_output"))
+
+class CensorWhitelistTest(BaseCensorTest):
+    CENSOR_CONFIG_FILE = 'tests/acra-censor_configs/acra-censor_whitelist.yaml'
+    def testWhitelist(self):
+        if TEST_MYSQL:
+            with self.assertRaises(sa.exc.OperationalError):
```
Lagovas: to avoid duplicated queries, we can do something like:
```
expectedException = None
if TEST_MYSQL:
  expectedException = sa.exc.OperationalError
else if TEST_POSTGRESQL:
  expectedException = sa.exc.ProgrammingError
with self.assertRaises(expectedException):
  ...
```

acra-censor/handlers/blacklist_handler.go
```diff
@@ -32,27 +32,22 @@ func (handler *BlacklistHandler) CheckQuery(query string) (bool, error) {
...
-					_, ok = fromStatement.(*sqlparser.ParenTableExpr)
-					if ok {
+						break
+					case *sqlparser.ParenTableExpr:
 						return false, ErrNotImplemented
-						//continueHandling, err := handler.handleParenTable(fromStatement.(*sqlparser.ParenTableExpr), forbiddenTable)
-						//if err != nil {
-						//
-						//}
+						break
```
Lagovas: after return we doesn't need `break` because function stop execution

tests/test.py
```diff
@@ -361,7 +361,7 @@ def terminate(self, *args, **kwargs):
     def poll(self, *args, **kwargs):
         pass
 
-
+'''
```
Lagovas: drop commented test?

PR #189, opened 2018-06-13
ruby mysql examples [T633]

Lagovas:
PR fixes mysql protocol parsing, and updates ruby example to work with both MySQL and PostgreSQL.

### DIFF
--- a/acra-censor/handlers/blacklist_handler.go
+++ b/acra-censor/handlers/blacklist_handler.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"errors"
+	"github.com/cossacklabs/acra/logging"
 	log "github.com/sirupsen/logrus"
 	"github.com/xwb1989/sqlparser"
 	"reflect"
@@ -82,7 +83,7 @@ func (handler *BlacklistHandler) AddQueries(queries []string) error {
 	for _, query := range queries {
 		_, err := sqlparser.Parse(query)
 		if err != nil {
-			log.WithError(err).Errorln("can't parse query")
+			log.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorQueryParseError).WithError(err).Errorln("can't parse query")
 			return ErrQuerySyntaxError
 		}
 		handler.queries = append(handler.queries, query)
--- a/acra-censor/handlers/blacklist_handler.go
+++ b/acra-censor/handlers/blacklist_handler.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"errors"
+	log "github.com/sirupsen/logrus"
 	"github.com/xwb1989/sqlparser"
 	"reflect"
 	"strings"
@@ -81,6 +82,7 @@ func (handler *BlacklistHandler) AddQueries(queries []string) error {
 	for _, query := range queries {
 		_, err := sqlparser.Parse(query)
 		if err != nil {
+			log.WithError(err).Errorln("can't parse query")
 			return ErrQuerySyntaxError
 		}
 		handler.queries = append(handler.queries, query)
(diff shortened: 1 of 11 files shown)


acra-censor/handlers/blacklist_handler.go
```diff
@@ -81,6 +82,7 @@ func (handler *BlacklistHandler) AddQueries(queries []string) error {
 	for _, query := range queries {
 		_, err := sqlparser.Parse(query)
 		if err != nil {
+			log.WithError(err).Errorln("can't parse query")
```
vixentael: error code?

acra-censor/handlers/whitelist_handler.go
```diff
@@ -87,6 +88,7 @@ func (handler *WhitelistHandler) AddQueries(queries []string) error {
 		handler.queries = append(handler.queries, query)
 		_, err := sqlparser.Parse(query)
 		if err != nil {
+			log.WithError(err).Errorln("can't parse query")
```
vixentael: error code as well?

decryptor/mysql/column_field.go
```diff
@@ -103,11 +108,13 @@ func ParseResultField(data []byte) (*ColumnDescription, error) {
 		//length of default value lenenc-int
 		field.DefaultValueLength, _, n, err = LengthEncodedInt(data[pos:])
 		if err != nil {
+			log.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorProtocolProcessing).WithError(err).Errorln("can't get length encoded integer of default value length")
```
vixentael: I would change wording to:

"SQL parsing error: can't get length encoded integer of default value length"

decryptor/mysql/column_field.go
```diff
@@ -103,11 +108,13 @@ func ParseResultField(data []byte) (*ColumnDescription, error) {
...
 			return nil, err
 		}
 		pos += n
 
 		if pos+int(field.DefaultValueLength) > len(data) {
+			log.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorProtocolProcessing).Errorln("incorrect position, malformed packet")
```
vixentael: similar as above,

"SQL parsing error: incorrect position, malformed packet"

decryptor/mysql/packet.go
```diff
@@ -15,7 +15,7 @@ const (
 	// SSL_REQUEST - https://dev.mysql.com/doc/internals/en/capability-flags.html#flag-CLIENT_SSL
 	SSL_REQUEST = 0x00000800
 	// https://dev.mysql.com/doc/internals/en/capability-flags.html#flag-CLIENT_DEPRECATE_EOF - 0x1000000
-	CLIENT_DEPRECATE_EOF = 1 << 6
+	CLIENT_DEPRECATE_EOF = 0x01000000
```
vixentael: why do you change it?

vixentael:
Basically this PR fixes mysql protocol parsing, and updates ruby example to work with both MySQL and PostgreSQL.

From my opinion we should start log messages from capital :)

decryptor/mysql/column_field.go
Lagovas: Here we parse packet of mysql protocol, not sql.

decryptor/mysql/packet.go
Lagovas: because 1<<6 it's `0b1000000` (correct was 1<<8 for 2 byte comparison) and `0x01000000` it's `0b1000000000000000000000000` and I changed comparison. before, we compared flag with 2 byte integer value. now used full compatibility flag that has 4 byte size.

Lagovas:
`From my opinion we should start log messages from capital, but it's not critical`
agree. better when all message has one format. I will fix it

Lagovas:
> What is the proper way of running this example?
> Acra should be on and running on localhost machine? As well as database?

actually, I remembered that we explain python examples in our wiki - https://github.com/cossacklabs/acra/wiki/Using-Acra-to-Protect-Your-Django-App#if-you-want-to-try-acra-in-containers-first
And all examples are the same. They differ only by running program (python/ruby/golang). Other are the same actually.

PR #190, opened 2018-06-13
golang mysql examples [T633]

Lagovas:


examples/golang/src/example_with_zone/example_with_zone.go
```diff
@@ -18,103 +18,128 @@ package main
...
-	}
-	var zone, data []byte
-	var rawData string
-	fmt.Println("zone, data - raw_data")
-	for rows.Next() {
-		err := rows.Scan(&zone, &data, &rawData)
+	if *data != "" {
+		rand.Seed(time.Now().UnixNano())
+		//get new zone over http
+		resp, err := http.Get("http://127.0.0.1:9191/getNewZone")
```
vixentael: is it the address of acraconnector? shall we put it into variable?

vixentael:
What is the proper way of running this example?
Acra should be on and running on localhost machine? As well as database?

examples/golang/src/example_with_zone/example_with_zone.go
Lagovas: yes. it's acraconnector's address and it used once, so I didn't assign it to variable and used default value

Lagovas:
yes. this is simple example that was created when Acra could work only with AcraConnector in the pair and also required db) can be used any our docker config where used AcraConnector + AcraServer + Database

Lagovas:
> What is the proper way of running this example?
> Acra should be on and running on localhost machine? As well as database?

actually, I remembered that we explain python examples in our wiki - https://github.com/cossacklabs/acra/wiki/Using-Acra-to-Protect-Your-Django-App#if-you-want-to-try-acra-in-containers-first
And all examples are the same. They differ only by running program (python/ruby/golang). Other are the same actually.

PR #191, opened 2018-06-22
Joinable and parenthesized tables handling

storojs72:


acra-censor/acra-censor_test.go
```diff
@@ -855,3 +853,36 @@ func testSyntax(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestDifferentTablesParsing(t *testing.T) {
+	testQuery := "SELECT Orders.OrderID, Customers.CustomerName, Shippers.ShipperName FROM ((Orders INNER JOIN Customers ON Orders.CustomerID = Customers.CustomerID) INNER JOIN Shippers ON Orders.ShipperID = Shippers.ShipperID);"
```
Lagovas: hard to read such query, first time see parenthesized joins in such way ;). can you reformat to smth like:
```
testQuery := `
    SELECT Orders.OrderID, Customers.CustomerName, Shippers.ShipperName 
    FROM Orders 
    INNER JOIN Customers ON Orders.CustomerID = Customers.CustomerID
    INNER JOIN Shippers ON Orders.ShipperID = Shippers.ShipperID;
`
```
some sql style guides:
* https://gist.github.com/fredbenenson/7bb92718e19138c20591
* https://www.sqlstyle.guide/
* https://medium.com/@lenguyenthedat/sql-style-guide-9d2d1ae805d2

acra-censor/handlers/blacklist_handler.go
```diff
@@ -71,6 +82,69 @@ func (handler *BlacklistHandler) CheckQuery(query string) (bool, error) {
...
+	}
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (handler *BlacklistHandler) handleParenTables(statement *sqlparser.ParenTableExpr, forbiddenTable string) error {
```
Lagovas: handleParenTables -> handleParen**t**Tables?

acra-censor/handlers/whitelist_handler.go
```diff
@@ -29,14 +29,30 @@ func (handler *WhitelistHandler) CheckQuery(query string) (bool, error) {
...
+					break
+				case *sqlparser.JoinTableExpr:
+					err = handler.handleJoinedTables(fromStatement.(*sqlparser.JoinTableExpr))
+					if err != nil {
+						return false, ErrAccessToForbiddenTableWhitelist
+					}
+				case *sqlparser.ParenTableExpr:
+					err = handler.handleParenTables(fromStatement.(*sqlparser.ParenTableExpr))
+					if err != nil {
+						return false, ErrAccessToForbiddenTableWhitelist
```
Lagovas: can you add logging all errors that we wrap with our errors with debug level?
```
log.WithError(err).Debugln("error from WhitlistHandler.handleParentTables")
```
I remember one situation when I had need something like that when debugged censor blocks and need original error to understand - it's parsing error or correct block

acra-censor/handlers/whitelist_handler.go
```diff
@@ -65,6 +81,107 @@ func (handler *WhitelistHandler) CheckQuery(query string) (bool, error) {
...
+			}
+
+			if err != nil {
+				return ErrAccessToForbiddenTableWhitelist
+			}
+		}
+	}
+
+	if allowedTablesCounter != len(parsedQuery) {
+		return ErrAccessToForbiddenTableWhitelist
```
Lagovas: can you add unit test for this method where `allowedTablesCounter != len(parsedQuery)`
i'm not sure that it works correctly because you use local `allowedTablesCounter` variable that will not incremented with other types of expressions than `AliasedTableExpr`. And in case when query will contain not only aliased expressions you always will have error

acra-censor/handlers/blacklist_handler.go
storojs72: No, it's 'handleParenTables' - like it is used in '*sqlparser.ParenTableExpr'

acra-censor/handlers/whitelist_handler.go
storojs72: Can you please provide example of sql query for this unit-test? I thought that existing unit tests are enough (`testWhitelistTables` and `TestDifferentTablesParsing`)

storojs72: It seems that `allowedTablesCounter` is global for inner functions that are called recursively. This behaviour is similar: https://play.golang.org/p/KoVkUb5ydrt

acra-censor/acra-censor_test.go
storojs72: Done

acra-censor/handlers/whitelist_handler.go
Lagovas: yes, it's global for inner functions. but here we don't create inner function with closure. this situation similar to https://play.golang.org/p/j7-hUWDxcu9

Lagovas: plus better to rewrite with map usage instead arrays for table validation and exclude extra outer for-loop

PR #192, opened 2018-07-04
mysql prepared statements

Lagovas:
* support prepared statements with [binary](https://dev.mysql.com/doc/internals/en/binary-protocol-resultset.html) and [text](https://dev.mysql.com/doc/internals/en/com-query-response.html#packet-ProtocolText::Resultset) formats of response
* test support of prepared statements with postgresql/mysql (two formats)
* fix race condition (when one goroutine close connection with secure session and second try to write/encrypt)
* increase wait time in acra-censor tests for correct running on slow machines (need refactoring in a future)
* small refactoring

### DIFF
--- a/cmd/acra-server/listener.go
+++ b/cmd/acra-server/listener.go
@@ -150,7 +150,7 @@ func (server *SServer) handleConnection(connection net.Conn) {
 	if err != nil {
 		log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCantInitClientSession).
 			Errorln("Can't initialize client session")
-		if closeErr := connection.Close(); closeErr != nil {
+		if closeErr := wrappedConnection.Close(); closeErr != nil {
 			log.WithError(closeErr).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCantCloseConnection).
 				Errorln("Can't close connection")
 		}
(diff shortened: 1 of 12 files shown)


cmd/acra-server/listener.go
```diff
@@ -402,6 +410,11 @@ func (server *SServer) StartCommands() {
 func (server *SServer) StartCommandsFromFileDescriptor(fd uintptr) {
 	var connection net.Conn
 	file := os.NewFile(fd, "/tmp/acra-server_http_api")
+	if file == nil {
+		log.Errorln("can't create new file from descriptor for api listener")
```
vixentael: usually we start logs with capital letter :)

just saying

cmd/utils.go
```diff
@@ -58,10 +58,10 @@ func (handler *SignalHandler) AddCallback(callback SignalCallback) {
 
 // Register should be called as goroutine
 func (handler *SignalHandler) Register() {
-	for _, osSignal := range handler.signals {
-		signal.Notify(handler.ch, osSignal)
-	}
+	signal.Notify(handler.ch, handler.signals...)
```
vixentael: so it notifies about many signals, right?

decryptor/mysql/response_proxy.go
```diff
@@ -325,7 +326,28 @@ func (handler *MysqlHandler) processBinaryDataRow(rowData []byte, fields []*Colu
...
+	// no data in response
+	if rowData[0] == EOF_PACKET {
+		return rowData, nil
+	}
+
+	if rowData[0] != OK_PACKET {
+		return nil, ErrMalformPacket
+	}
+
+	pos = 1 + ((len(fields) + 7 + 2) >> 3)
```
vixentael: 7? 2?

can you please add some comment or save this variables as constants?

cmd/utils.go
Lagovas: yes

cmd/acra-server/listener.go
Lagovas: will update. forgot

PR #193, opened 2018-07-05
Using JSON line format for QueryCapture

vixentael:
As we discussed previously having large JSON file as Query blob might be not the best option [T605]

- format changed from large json to multiline json
- on each parsed query remember query in the buffer
- dump buffer into file once a while (instead of writing into the file on each query)

### DIFF
--- a/acra-censor/handlers/querycapture_handler.go
+++ b/acra-censor/handlers/querycapture_handler.go
@@ -24,8 +24,8 @@ func trimToN(query string, n int) string {
 }
 
 type QueryCaptureHandler struct {
-	Queries              []QueryInfo
-	BufferedQueries      []QueryInfo
+	Queries              []*QueryInfo
+	BufferedQueries      []*QueryInfo
 	filePath             string
 	signalBackgroundExit chan bool
 	serializationTimeout time.Duration
@@ -40,7 +40,7 @@ func NewQueryCaptureHandler(filePath string) (*QueryCaptureHandler, error) {
 	// open or create file, APPEND MODE
 	openedFile, err := os.OpenFile(filePath, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0600)
 	if err != nil {
-		log.WithError(ErrSingleQueryCaptureError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSecurityError)
+		log.WithError(ErrCantReadQueriesFromFileError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError)
 		return nil, err
 	}
 
@@ -58,8 +58,12 @@ func NewQueryCaptureHandler(filePath string) (*QueryCaptureHandler, error) {
 	handler.serializationTicker = time.NewTicker(DefaultSerializationTimeout)
 
 	// read existing queries from file
-	handler.ReadAllQueriesFromFile()
-
+	err = handler.ReadAllQueriesFromFile()
+	if err != nil {
+		log.WithError(ErrCantReadQueriesFromFileError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError)
+		openedFile.Close()
+		return nil, err
+	}
 
 	//handling goroutine
 	go func() {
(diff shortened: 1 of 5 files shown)


acra-censor/handlers/querycapture_handler.go
```diff
@@ -38,99 +37,59 @@ type QueryInfo struct {
...
 	handler.serializationTimeout = DefaultSerializationTimeout
 	handler.serializationTicker = time.NewTicker(DefaultSerializationTimeout)
 
-	f, err := os.OpenFile(filePath, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
-	if err != nil {
-		log.WithError(ErrSingleQueryCaptureError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSecurityError)
-		return nil, err
-	}
+	// read existing queries from file
+	handler.ReadAllQueriesFromFile()
```
Lagovas: handle error and close file if `err != nil`

acra-censor/handlers/querycapture_handler.go
```diff
@@ -38,99 +37,59 @@ type QueryInfo struct {
...
-					if _, err = f.Write(bytes); err != nil {
-						log.WithError(ErrSingleQueryCaptureError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSecurityError)
-					}
-
-				} else {
-					//channel is unexpectedly closed
-					log.WithError(ErrUnexpectedCaptureChannelClose).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSecurityError)
-				}
-
+			// TODO: how to remove duplicate code?
```
Lagovas: Golang doesn't provide way to write two cases at once like in switch with `fallthrough` becase any read from channel block goroutine and for `fallthrough` it need to read second channel :) So we can hack it like:
1. move logic to some function
```
func (handler *QueryCaptureHandler) finish(){
  handler.serializationTicker.Stop()
  err := handler.DumpAllQueriesToFile()
  if err != nil {
      log.WithError(ErrComplexSerializationError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSecurityError)
  }
  openedFile.Close()
}
```
and call it in two cases
```
case <-someCh:
....
   handler.finish()
   return
....
```

2. use one more channel that will flag that we need to finish
```
exitCh := make(chan bool)
for {
  select {
     case <-signalBackgroundExit:
       exitCh <- true
     case <-signalShutdown:
       exitCh <- true
  }
  select {
    case <-exitCh:
       // exit code
    default:
      // do nothing, continue
  }
}
```

acra-censor/handlers/querycapture_handler.go
```diff
@@ -152,18 +111,15 @@ func (handler *QueryCaptureHandler) CheckQuery(query string) (bool, error) {
 	queryInfo.RawQuery = query
 	queryInfo.IsForbidden = false
 	handler.Queries = append(handler.Queries, *queryInfo)
-
-	select {
-	case handler.logChannel <- *queryInfo: // channel is ok
-	default: //channel is full
-		log.Errorf("Can't process too many queries. Skip query: %s", trimToN(query, LogQueryLength))
-	}
+	handler.BufferedQueries = append(handler.BufferedQueries, *queryInfo)
```
Lagovas: better to append pointers instead copy of struct. change type of BufferedQueries from `QueryInfo` to `*QueryInfo` and then append only pointers.

acra-censor/handlers/querycapture_handler.go
```diff
@@ -192,25 +148,112 @@ func (handler *QueryCaptureHandler) GetForbiddenQueries() []string {
...
 }
+
 func (handler *QueryCaptureHandler) GetSerializationTimeout() time.Duration {
 	return handler.serializationTimeout
 }
 
-func (handler *QueryCaptureHandler) Serialize() error {
-	jsonFile, err := json.Marshal(handler.Queries)
+func (handler *QueryCaptureHandler) DumpAllQueriesToFile() error {
+	// open or create file, NO APPEND
```
Lagovas: what about to truncate file and add `O_TRUNC` flag too?

acra-censor/handlers/querycapture_handler.go
```diff
@@ -192,25 +148,112 @@ func (handler *QueryCaptureHandler) GetForbiddenQueries() []string {
...
+
+	// nothing to dump
+	if len(buf) == 0 {
+		return nil
+	}
+
+
+	// clean buffered queries
+	handler.BufferedQueries = nil
+	err := AppendQueries(buf, openedFile)
```
Lagovas: we can just `return AppendQueries(buf, openedFile)` 
or `return err` below without `if`

acra-censor/handlers/querycapture_handler.go
```diff
@@ -192,25 +148,112 @@ func (handler *QueryCaptureHandler) GetForbiddenQueries() []string {
...
+
+func SerializeQueries(queries []QueryInfo) ([]byte, error) {
+	var linesToAppend []byte
+	for _, queryInfo := range queries {
+		jsonQueryInfo, err := json.Marshal(queryInfo)
+		if err != nil {
+			return nil, err
+		}
+		if len(jsonQueryInfo) > 0 {
+			jsonQueryInfo = append(jsonQueryInfo, "\n"...)
```
Lagovas: can we change to `append(jsonQueryInfo, '\n')` (single quotes are one character literal, double quotes are string literal that iterable)? after `"\n"... I think about _"is it only 1 character or it 2 characters and therefore we use ... ?"_ )

vixentael:
@Lagovas thank you for your comments

take a look!

PR #200, opened 2018-07-12
grpc handler

Lagovas:


cmd/acra-reader/server.go
```diff
@@ -167,7 +190,6 @@ func (server *ReaderServer) processGRPCConnection(parentContext context.Context,
 	logger.Debugln("grpc handler")
```
vixentael: `processGRPCConnection ` not needed now?

cmd/acra-reader/api/decryptor.go
```diff
@@ -0,0 +1,37 @@
...
+	if len(request.ZoneId) != 0 {
+		privateKey, err = service.keystorage.GetZonePrivateKey(request.ZoneId)
+		decryptionContext = request.ZoneId
+	} else {
+		privateKey, err = service.keystorage.GetServerDecryptionPrivateKey(request.ClientId)
+	}
+	if err != nil {
+		return nil, err
+	}
+	data, err := base.DecryptAcrastruct(request.Acrastruct, privateKey, decryptionContext)
```
vixentael: fill private keys with zeros?

PR #206, opened 2018-07-16
Update AcraConnector to connect with AcraTranslator

vixentael:
AcraConnector now has required parameter `mode` (could be "AcraServer" or "AcraTranslator") which defines what connection string / host / port / secure-session-id to use.

By default mode is set to AcraServer for backwards compatibility.

Setting mode to AcraTranslator allows to point AcraConnector to AcraTranslator.

Usage (using all params from configuration file):

```
$GOPATH/bin/acra-connector --mode=acratranslator
```

Overriding config params:
```
 --client_id=client --acratranslator_securesession_id=acra_translator --incoming_connection_string=tcp://127.0.0.1:8000 --acratranslator_connection_string=tcp://127.0.0.1:9595 -v 
```

Keymaker now can generate `client_translator` keypair that is used for decrypting SecureSession traffic between AC and AT.

Important: this PR is using updated naming, should be merged after #205


cmd/acra-connector/acra-connector.go
```diff
@@ -156,20 +158,27 @@ func main() {
...
 	useTls := flag.Bool("acraserver_tls_transport_enable", false, "Use tls to encrypt transport between AcraServer and AcraConnector/client")
 	tlsCA := flag.String("tls_ca", "", "Path to root certificate which will be used with system root certificates to validate AcraServer's certificate")
 	tlsKey := flag.String("tls_key", "", "Path to private key that will be used in TLS handshake with AcraServer")
 	tlsCert := flag.String("tls_cert", "", "Path to certificate")
 	tlsAcraserverSNI := flag.String("tls_acraserver_sni", "", "Expected Server Name (SNI) from AcraServer")
-	tlsAuthType := flag.Int("tls_auth", int(tls.RequireAndVerifyClientCert), "Set authentication mode that will be used in TLS connection with Postgresql. Values in range 0-4 that set auth type (https://golang.org/pkg/crypto/tls/#ClientAuthType). Default is tls.RequireAndVerifyClientCert")
+	tlsAuthType := flag.Int("tls_auth", int(tls.RequireAndVerifyClientCert), "Set authentication mode that will be used in TLS connection with AcraServer/AcraReader. Values in range 0-4 that set auth type (https://golang.org/pkg/crypto/tls/#ClientAuthType). Default is tls.RequireAndVerifyClientCert")
```
Lagovas: `AcraReader` -> `AcraTranslator`

cmd/acra-connector/acra-connector.go
```diff
@@ -186,37 +195,69 @@ func main() {
...
+			if *acraServerHost != "" {
+				*acraServerApiConnectionString = network.BuildConnectionString(cmd.DEFAULT_ACRA_CONNECTION_PROTOCOL, *acraServerHost, *acraServerApiPort, "")
+			}
+		}
+
+		outgoingConnectionString = *acraServerConnectionString
+		outgoingSecureSessionId = *acraServerId
 	}
 
+	// --------- check client id and client key  -----------
```
Lagovas: why hyphens?)
in a future to edit such comments we will need to move cursor through hyphens and only then update comment
or delete pre/post hyphens and edit text. instead just edit comment text)
Anyway, I think any developer will understand what we do when calling `ValidateClientId` and will not need any comment to understand :)

cmd/acra-connector/acra-connector.go
```diff
@@ -228,22 +269,50 @@ func main() {
...
-	if !exists {
-		log.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorWrongConfiguration).
-			Errorf("Configuration error: AcraServer public key %s doesn't exists", serverPublicKey)
-		os.Exit(1)
+	log.Infof("Client id and client key is OK")
+
+
+	// --------- check public key  -----------
+	if connectorMode == connector_mode.AcraTranslatorMode {
+		translatorPublicKey := fmt.Sprintf("%v%v%v_translator.pub", *keysDir, string(os.PathSeparator), *clientId)
```
Lagovas: before we did check public key incorrectly. what about to fix it with new changes where we will use keystore for it and  cal `keystore.GetPeerPublicKey([]byte(*OutgoingServiceId))` to verify that we can load this key and encapsulate knowledge about file naming?

cmd/acra-connector/acra-connector.go
```diff
@@ -228,22 +269,50 @@ func main() {
...
+		}
+		log.Infof("AcraTranslator public key is OK")
 	}
-	if err != nil {
-		log.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorWrongConfiguration).
-			Errorf("Configuration error: can't check is exists AcraServer public key %v, got error - %v", serverPublicKey, err)
-		os.Exit(1)
+
+	if connectorMode == connector_mode.AcraServerMode {
+		serverPublicKey := fmt.Sprintf("%v%v%v_server.pub", *keysDir, string(os.PathSeparator), *clientId)
```
Lagovas: will be great to check key via keystore here too (or drop two checks by one with GetPeerPublicKey, because keystore now parameterized with mode)

cmd/acra-connector/acra-connector.go
vixentael: updated

PR #207, opened 2018-07-18
acra reader integration tests [T663]

Lagovas:


tests/test.py
```diff
@@ -2160,5 +2164,214 @@ def executePreparedStatement(self, query):
...
+            stop_process(translator)
+            raise
+        return translator
+
+    def fork_connector(self, connector_port: int, server_port: int, client_id: str, check_connection: bool=True):
+        server_connection = get_tcp_connection_string(server_port)
+        connector_connection = get_tcp_connection_string(connector_port)
+        args = [
+            './acra-connector',
+            '-acraserver_connection_string={}'.format(server_connection),
```
vixentael: `-acratranslator_connection_string`
`mode=AcraTranslator`

tests/test.py
```diff
@@ -2160,5 +2164,214 @@ def executePreparedStatement(self, query):
...
+                self.assertIn("Can't decrypt AcraStruct".lower(),
+                              response.text.lower())
+                self.assertEqual(response.headers['Content-Type'], 'text/plain')
+
+
+                # test with correct acrastruct
+                response = requests.post(api_url, data=acrastruct)
+                self.assertEqual(data, response.content)
+                self.assertEqual(response.status_code, http.HTTPStatus.OK)
+                self.assertEqual(response.headers['Content-Type'],
```

network/secure_session_wrapper.go
```diff
@@ -152,20 +153,28 @@ func (err *ConnectionWrapError) Timeout() bool {
...
 	return true
-	//netErr, ok := err.error.(net.Error)
-	//if ok {
-	//	return netErr.Temporary()
-	//}
-	//return false
 }
 
+// SECURE_SESSION_ESTABLISHING_TIMEOUT timeout for secure session handshake that should be enough
+const SECURE_SESSION_ESTABLISHING_TIMEOUT = time.Second * 10
```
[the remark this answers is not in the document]
Lagovas: no. first value was 2 sec but then decide to leave 10 sec for slow connections. It's rare case when it will need. only on network errors when it will not fail on earlier stages (wrap/unwrap, searching public key key)

PR #211, opened 2018-07-20
Hide values while query logging + some refactoring

storojs72:


acra-censor/handlers/querycapture_handler.go
```diff
@@ -219,8 +220,15 @@ func AppendQueries(queries []*QueryInfo, openedFile *os.File) error {
 
 func SerializeQueries(queries []*QueryInfo) ([]byte, error) {
 	var linesToAppend []byte
+	var tempQueryInfo = &QueryInfo{}
 	for _, queryInfo := range queries {
-		jsonQueryInfo, err := json.Marshal(queryInfo)
+		queryWithHiddenValues, err := sqlparser.RedactSQLQuery(queryInfo.RawQuery)
```
Lagovas: may we copy implementation and refactor it to accept placeholder as param? I think in a future we will want to change prefix to something else than "redactedN". Plus tests will not be so hardcoded as now (if I wouldn't know about exact value for placeholder then I wouldn't understand why we expect "redacted1" after function call).

acra-censor/handlers/querycapture_handler.go
```diff
@@ -219,8 +221,15 @@ func AppendQueries(queries []*QueryInfo, openedFile *os.File) error {
 
 func SerializeQueries(queries []*QueryInfo) ([]byte, error) {
 	var linesToAppend []byte
+	var tempQueryInfo = &QueryInfo{}
 	for _, queryInfo := range queries {
-		jsonQueryInfo, err := json.Marshal(queryInfo)
+		queryWithHiddenValues, err := RedactSQLQuery(queryInfo.RawQuery)
+		if err != nil {
+			return nil, ErrQuerySyntaxError
```
vixentael: I believe we shouldn't return error if query can't be redacted, just skip it during writing.

I think it's better to skip writing one query to QueryCaptureLog, then not to write all of them due to one error.

acra-censor/handlers/querycapture_handler.go
```diff
@@ -12,9 +12,12 @@ import (
 	"bytes"
 	"github.com/cossacklabs/acra/logging"
 	log "github.com/sirupsen/logrus"
+	"github.com/xwb1989/sqlparser"
+	"github.com/xwb1989/sqlparser/dependency/querypb"
 )
 
 const DefaultSerializationTimeout = time.Second
+const ValuePlaceholder = "X"
```
Lagovas: better to change to some more obvious as placeholder, something like `::replaced::`
because I think sometimes in the future even me will be surprised during debugging when will see this value when it can be by logic ))

acra-censor/handlers/querycapture_handler.go
Lagovas: @vixentael it will be ignored if IgnoreParseError setting will be set in acra-censor config (acra-censor/acra-censor_implementation.go:35, configs/acra-censor.example.yaml:1)

But I think better to add here some log `log.WithError(err).Errorln("Can't hide values")` to understand what and where it happened

vixentael: I think we should add log for sure.

But I'm concerned about the fact that if masking at least one SQL query fails, the code exits from `for` loop, and returns error instead of json lines. I think code should log error, and skip this line (continue parsing other queries).

PR #214, opened 2018-07-23
log who decrypt [T664]

Lagovas:
* pass client_id to `decryptor` objects to use in internal loggers
* client_id is required in grpc
* refactor some log messages
* add check to ci that all code formatted with gofmt


vixentael:
wow, lots of changes!

Lagovas:
> wow, lots of changes!

gofmt and Ctrl+Shift+R do magic

PR #218, opened 2018-07-25
Some golint fixes

vixentael:
Style fixes according to golint checks.

Decreased amount of errors on ~25% (from 1035 to 788).

I don't want to make super-huge PR, so I'll add changes step-by-step.

### DIFF
--- a/cmd/acra-translator/grpc_api/decryptor.go
+++ b/cmd/acra-translator/grpc_api/decryptor.go
@@ -11,25 +11,31 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// DecryptGRPCService represents decryptor for decrypting AcraStructs from gRPC requests.
 type DecryptGRPCService struct {
 	*common.TranslatorData
 }
 
+// NewDecryptGRPCService creates new DecryptGRPCService.
 func NewDecryptGRPCService(data *common.TranslatorData) (*DecryptGRPCService, error) {
 	return &DecryptGRPCService{TranslatorData: data}, nil
 }
 
-var ErrCantDecrypt = errors.New("can't decrypt data")
-var ErrClientIdRequired = errors.New("ClientId is empty")
+// Errors possible during decrypting AcraStructs.
+var (
+	ErrCantDecrypt      = errors.New("can't decrypt data")
+	ErrClientIDRequired = errors.New("ClientId is empty")
+)
 
+// Decrypt decrypts AcraStruct from gRPC request and returns decrypted data or error.
 func (service *DecryptGRPCService) Decrypt(ctx context.Context, request *DecryptRequest) (*DecryptResponse, error) {
 	var privateKey *keys.PrivateKey
 	var err error
-	var decryptionContext []byte = nil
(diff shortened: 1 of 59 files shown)


cmd/acra-translator/grpc_api/decryptor.go
```diff
@@ -11,25 +11,31 @@ import (
...
 func NewDecryptGRPCService(data *common.TranslatorData) (*DecryptGRPCService, error) {
 	return &DecryptGRPCService{TranslatorData: data}, nil
 }
 
-var ErrCantDecrypt = errors.New("can't decrypt data")
-var ErrClientIdRequired = errors.New("ClientId is empty")
+// Errors possible during decrypting AcraStructs.
+var (
+	ErrCantDecrypt      = errors.New("can't decrypt data")
+	ErrClientIDRequired = errors.New("ClientID is empty")
```
Lagovas: `CliendId` was used because it's name of generated filed with protobuf compiler (cmd/acra-translator/grpc_api/api.pb.go:27). I think better to use field name as is

cmd/acra-translator/grpc_api/decryptor.go
```diff
@@ -11,25 +11,31 @@ import (
...
 func (service *DecryptGRPCService) Decrypt(ctx context.Context, request *DecryptRequest) (*DecryptResponse, error) {
 	var privateKey *keys.PrivateKey
 	var err error
-	var decryptionContext []byte = nil
+	var decryptionContext []byte
 	logger := logrus.WithFields(logrus.Fields{"client_id": string(request.ClientId), "zone_id": string(request.ZoneId), "translator": "grpc"})
 	if len(request.ClientId) == 0 {
-		logrus.Errorln("Grpc request without ClientId not allowed")
-		return nil, ErrClientIdRequired
+		logrus.Errorln("GRPC request without ClientID not allowed")
```
Lagovas: same. and linter doesn't care about how written `clientId` in logs :) it will help in the future if we will need to find a place in code by text from logs (with case sensitive search^)

decryptor/mysql/response_proxy.go
```diff
@@ -295,8 +295,8 @@ func (handler *MysqlHandler) isFieldToDecrypt(field *ColumnDescription) bool {
 func (handler *MysqlHandler) processTextDataRow(rowData []byte, fields []*ColumnDescription) ([]byte, error) {
 	var err error
 	var value []byte
-	var pos int = 0
-	var n int = 0
```
Lagovas: hate golang for preference of the implicit explicit ))

vixentael: agree 100%

cmd/acra-translator/grpc_api/decryptor.go
vixentael: currently "clientId" is used only in grpc field, because name is auto-generated from "client_id". I would prefer to change style to use one naming everywhere, "ClientID".

Is there any way how to change name of grpc-generated field to "ClientID"?

PR #221, opened 2018-07-25
Continue fixing golint errors [T708]

vixentael:
Fixes include:
- added package headers here and there.
- added many comments to public functions/ types.
- fixed "if block ends with a return statement, so drop this else and outdent its block".
- changed "n += 1" to "n++".
- fixed naming for some fields (serverId -> serverID).

Errors number decreased from 788 to 512.


cmd/acra-authmanager/acra_authmanager.go
```diff
@@ -83,7 +91,7 @@ func (hp HashedPasswords) SetPassword(name, password string) (err error) {
 	return nil
 }
 
-func ParseHtpasswdFile(file string, keystore *filesystem.FilesystemKeyStore) (passwords HashedPasswords, err error) {
+func parseHtpasswdFile(file string, keystore *filesystem.FilesystemKeyStore) (passwords HashedPasswords, err error) {
```
Lagovas: change to private because used only in this module?

decryptor/base/callbacks_test.go
```diff
@@ -8,7 +8,7 @@ import (
 type TestCallback struct{ CallCount *int }
 
 func (callback *TestCallback) Call() error {
-	*callback.CallCount += 1
```
Lagovas: python style vs golang style))

cmd/acra-authmanager/acra_authmanager.go
vixentael: yes

vixentael:
> check and update variable name of main.TestOnly in tests/test.py

that's might be the reason of failing tests. thank you

Lagovas:
awesome results) well done :)

PR #223, opened 2018-07-26
kill forked process

Lagovas:
I think some tests fail because I dropped `killall` command (by mistake *( ) that kill all instances of acra-server (thought that it's bad way and in other code it clean up processes already, but nope )
Here I more accurately kill forked process by parsing log about fork and fetch correct pid to kill (instead all in system)


vixentael:
Great, I hope it will help :))

PR #224, opened 2018-07-27
More fixing of golint issues + add check to CircleCI

vixentael:
- Add comments.
- Decrease golint issues to 421.
- Add checks in CircleCI, use "450" issues as magic constant now. After fixing most of issues, I'll update this constant.


vixentael:
I noticed that `golint` package for Ubuntu has different versions from the one installed through `go get`. Number of issues vary in unpredictable manner.

Will disable golint check on CircleCI for now, I'll think what's the best way to turn it on.

PR #229, opened 2018-08-01
remove duplicated package comments

vixentael:
In addition to #228 it happens that only one package file should contain package comment 🙄

Removed extra.


vixentael:
yeah, didn't realize, that they have so strict code style

PR #230, opened 2018-08-01
check poison records after decryption [T685]

Lagovas:
* check poison records on AcraStruct decryption failure
* refactor processing column's data and move logic to separate object PacketHandler
* drop redundant ExtendedBufferedReader
* changed processing errors from pass errCh to return error and handle it in one place (pgproxy)
* use separate logger instances
* move processing some if blocks from PgDecrypt to functions to simplify reading and understanding
* drop testing golang < 1.8


decryptor/base/decryptor.go
```diff
@@ -49,57 +48,33 @@ hex   char dec  bin
 */
 
 //var TAG_BEGIN = []byte{133, 32, 251}
+
+// Constants that setup which symbol would be used at start in AcraStruct to simplify recognizing from other binary data
+// Double-quote was chosen because it's printable symbol (help in debugging when we can see in console that it's start of
+// AcraStruct) and rarely used sequentially
+// Tag length was chosen
```
vixentael: > Tag length was chosen

_dramatic pause..._

I will never know why Tag length was chose!

decryptor/postgresql/pg_decryptor.go
```diff
@@ -294,408 +92,401 @@ func NewPgProxy(clientConnection, dbConnection net.Conn) (*PgProxy, error) {
...
 				return
 			}
+			continue
+		}
+
+		logger.Debugf("Process columns data")
+		for i := 0; i < packetHandler.columnCount; i++ {
+			column := packetHandler.Columns[i]
+
 			// TODO check poison record before zone matching in two modes.
```
vixentael: i think we can remove TODO now

vixentael:
wow, thank you, that's a great (but large) PR. 
awesome job!

P.S. please update [golint-checker](https://github.com/cossacklabs/acra/blob/master/.circleci/check_golint.sh) with new value.

--------

Short summary, correct me if I'm wrong:

for users:
- we decreased decryption time for AcraServer with poison records checks.

for us:
- split some large files into smaller ones
- improve logging / error handling
- improve tests for graceful restart
- rename lots of constants to decrease golint issues.

PR #231, opened 2018-08-02
Add patterns functionality to whitelist

storojs72:


acra-censor/acra-censor_test.go
```diff
@@ -103,68 +107,322 @@ func testWhitelistTables(t *testing.T, acraCensor *AcraCensor, whitelistHandler
...
+	whitelist := handlers.NewWhitelistHandler()
+	censor.AddHandler(whitelist)
+
+	pattern := "SELECT a, b, c FROM z %%WHERE%%"
+	err = whitelist.AddPatterns([]string{pattern})
+	if err != nil {
+		t.Fatal(err)
+	}
+	acceptableQueries := []string{
+		"SELECT a, b, c FROM z WHERE a = 'someValue'",
```
Lagovas: can we test here queries that we use for star pattern too (https://github.com/cossacklabs/acra/blob/acracensor/acra-censor/acra-censor_test.go#L545) ? to test different `where` clauses without star and columns

acra-censor/handlers/handlers_util.go
```diff
@@ -151,6 +151,9 @@ func handleSelectWherePattern(queryNodes, patternNodes []sqlparser.SQLNode) bool
 	patternWhereDetected := false
 	queryWhereDetected := false
 	for index, patternNode := range patternNodes {
+		if index >= len(queryNodes) {
```
Lagovas: maybe it makes sense to compare slice's lengths before loop?
```
if len(queryNodes) != len(patternNodes)
```

vixentael:
i noticed in many places in code and logs we use `parenTable`, you mean `parentTable`?

vixentael:
ahh, `parenthesized`.. ok, got it

acra-censor/handlers/handlers_util.go
storojs72: Yes, actually we can perform this comparison before loop. But instead  `!=` it's better `<`, because length of query is allowed to be equal or greater than pattern. The intention here was to prevent case when pattern is greater than query, and index out of range when use `queryNodes[index]`

acra-censor/handlers/handlers_util.go
```diff
@@ -78,3 +68,189 @@ func RedactSQLQuery(sql string) (string, error) {
...
+		if index+patternNodeOffset >= len(patternNodes) || index+queryNodeOffset >= len(queryNodes) {
+			return false
+		}
+		if reflect.DeepEqual(patternNodes[index+patternNodeOffset], queryNodes[index+queryNodeOffset]) {
+			continue
+		}
+		if patternSelectExpr, ok := patternNodes[index+patternNodeOffset].(sqlparser.SelectExprs); ok && starFound(patternSelectExpr) {
+			if _, ok := queryNodes[index+queryNodeOffset].(sqlparser.SelectExprs); ok {
+				i := index
+				for i < len(queryNodes) {
```
Lagovas: `for ;i<len(queryNodes); i++`

acra-censor/handlers/handlers_util.go
```diff
@@ -78,3 +68,189 @@ func RedactSQLQuery(sql string) (string, error) {
...
+	queryNodeOffset := 0
+	for index := 1; index < len(patternNodes); index++ {
+		if index+patternNodeOffset >= len(patternNodes) || index+queryNodeOffset >= len(queryNodes) {
+			return false
+		}
+		if reflect.DeepEqual(patternNodes[index+patternNodeOffset], queryNodes[index+queryNodeOffset]) {
+			continue
+		}
+		if patternSelectExpr, ok := patternNodes[index+patternNodeOffset].(sqlparser.SelectExprs); ok && starFound(patternSelectExpr) {
+			if _, ok := queryNodes[index+queryNodeOffset].(sqlparser.SelectExprs); ok {
```
Lagovas: what if not ok?

acra-censor/handlers/handlers_util.go
storojs72: Wow, didn't know this loop syntax. Thanks

PR #237, opened 2018-08-09
Censor refactoring

storojs72:
Improve %%VALUE%% pattern to handle specific WHERE structures:
- BETWEEN %%VALUE%% AND %%VALUE%%
- LIKE %%VALUE%%

Add '*' case handling to %%VALUE%% pattern


.circleci/config.yml
```diff
@@ -20,14 +20,14 @@ jobs:
     environment:
       GOTHEMIS_IMPORT: github.com/cossacklabs/themis/gothemis
       FILEPATH_ERROR_FLAG: /tmp/test_fail
-      VERSIONS: 1.7 1.7.6 1.8 1.8.7 1.9.4 1.10
+      VERSIONS: 1.8 1.8.7 1.9.4 1.10 1.10.3
```
shadinua: It would be better to lead testing versions to a system. 
Do we want to test only latest subversions of each version? If so than `VERSIONS: 1.8.7 1.9.7 1.10.3`.
If we also want to test first release of each version, than `VERSIONS: 1.8 1.8.7 1.9 1.9.4 1.10 1.10.3`.

acra-censor/acra-censor_test.go
```diff
@@ -805,18 +827,29 @@ func testBlacklistValuePattern(t *testing.T) {
...
+		"SELECT * FROM t WHERE ID LIKE 'Pa%'",
+		"SELECT * FROM t WHERE ID IN ( 25, 27 )",
+		"SELECT * FROM t WHERE ID NOT IN ( 25, 27 )",
 	}
 	blockableQueries := []string{
 		"SELECT a, b FROM t WHERE ID = 'someValue_testValue_1234567890'",
-		"SELECT a, b FROM t WHERE ID = 'someValue'",
+		"SELECT * FROM t WHERE ID = 'someValue'",
+		"SELECT * FROM t WHERE ID = 999",
 	}
```
shadinua: Perhaps we could also add invalid queries to the test list?

vixentael:
@storojs72 why did you merge `master` into `acracensor` on this step?

I think we should close this PR and open new two:

1. merge `storojs72:censor_refactoring` into `cossacklabs:acracensor` to see and comment your latest features. when approved -- merge to `acracensor`.

2. after that, open separate PR from `acracensor` into `master` to add new AcraCensor features into master, merging master into `acracensor` first (like you did here -> https://github.com/cossacklabs/acra/pull/237/commits/652c8bcd73fb04110bb042978138e6a13857d77b)

.circleci/config.yml
vixentael: makes sense for me to test only latest (first option). @Lagovas what do u think?

@shadinua will you do a separate PR to master with proposed change?

PR #238, opened 2018-08-09
Improve %%VALUE%% pattern

storojs72:
1) Add ' * ' support in pattern
2) Update handling of *ComparedExprs values
3) Add tests for whitelist and blacklist

### DIFF
--- a/acra-censor/acra-censor_test.go
+++ b/acra-censor/acra-censor_test.go
@@ -842,7 +842,33 @@ func testBlacklistValuePattern(t *testing.T) {
 	for _, query := range acceptableQueries {
 		err = censor.HandleQuery(query)
 		if err != nil {
+			t.Fatal(err, "Blacklist pattern blocked query. \nPattern:", blacklistPattern, "\nQuery:", query)
+		}
+	}
+	for _, query := range blockableQueries {
+		err = censor.HandleQuery(query)
+		if err != handlers.ErrBlacklistPatternMatch {
 			t.Fatal(err, "Blacklist pattern passed query. \nPattern:", blacklistPattern, "\nQuery:", query)
+		}
+	}
+
+	blacklist.Reset()
+	blacklistPattern = "SELECT * from t where ID = %%VALUE%%"
+	err = blacklist.AddPatterns([]string{blacklistPattern})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	blockableQueries = []string{
+		"SELECT a, b FROM t WHERE ID = 'someValue_testValue_1234567890'",
+		"SELECT a, b FROM t WHERE ID = 'someValue'",
+		"SELECT a, b, c, d FROM t WHERE ID = 'someValue'",
+	}
+
+	for _, query := range acceptableQueries {
+		err = censor.HandleQuery(query)
+		if err != nil {
+			t.Fatal(err, "Blacklist pattern blocked query. \nPattern:", blacklistPattern, "\nQuery:", query)
 		}
 	}
 	for _, query := range blockableQueries {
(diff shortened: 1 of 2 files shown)


acra-censor/acra-censor_test.go
```diff
@@ -384,6 +384,32 @@ func testWhitelistValuePattern(t *testing.T) {
...
+	whitelist.Reset()
+	pattern = "SELECT * from t where ID = %%VALUE%%"
+	err = whitelist.AddPatterns([]string{pattern})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	acceptableQueries = []string{
+		"SELECT a, b FROM t WHERE ID = 'someValue_testValue_1234567890'",
+		"SELECT a, b FROM t WHERE ID = 'someValue'",
```
Lagovas: better to test different type of values instead same with different length/value. can we check boolean/integer/float/nullable values too? `where a=1 or b=1.0 or c=TRUE or d=NULL

acra-censor/acra-censor_test.go
```diff
@@ -816,9 +842,35 @@ func testBlacklistValuePattern(t *testing.T) {
...
+	blacklist.Reset()
+	blacklistPattern = "SELECT * from t where ID = %%VALUE%%"
+	err = blacklist.AddPatterns([]string{blacklistPattern})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	blockableQueries = []string{
+		"SELECT a, b FROM t WHERE ID = 'someValue_testValue_1234567890'",
+		"SELECT a, b FROM t WHERE ID = 'someValue'",
```
Lagovas: same as above about different types of value

acra-censor/handlers/handlers_util.go
```diff
@@ -174,17 +174,49 @@ func handleSelectWherePattern(queryNodes, patternNodes []sqlparser.SQLNode) bool
...
+						patternNodeOffset++
+					}
+					continue
+				}
+			}
+		}
+
+		if patternNodeComparison, ok := patternNodes[index + patternNodeOffset].(*sqlparser.ComparisonExpr); ok && patternNodeComparison != nil {
+			if queryNodeComparison, ok := queryNodes[index + queryNodeOffset].(*sqlparser.ComparisonExpr); ok && queryNodeComparison != nil {
+				if reflect.DeepEqual(queryNodeComparison.Left, patternNodeComparison.Left) &&
```
Lagovas: reflect.DeepEqual don't compare all fields? why not just `reflect.DeepEqual(queryNodeComparison, patternNodeComparison)`? If it didn't, so I suggest to move this logic to some small function and use it to increase readability and understanding:
```
func IsEqualComparisonNode(node1, node2) bool {
  ....
}
```
then there use just
```
  if IsEqualComparisonNode(queryNodeComparison, patternNodeComparison) {
     ....
  } else {
     ....
  }
```
After this refactoring other people (like me) wouldn't need to know about node implementations and their structure to understand what actually we expect and check in this context .

acra-censor/acra-censor_test.go
Lagovas: plus we should check that it works with >1 conditions - `where a=1 and b=2`, not only with one.

acra-censor/acra-censor_test.go
```diff
@@ -860,9 +864,10 @@ func testBlacklistValuePattern(t *testing.T) {
 	}
 
 	blockableQueries = []string{
-		"SELECT a, b FROM t WHERE ID = 'someValue_testValue_1234567890'",
-		"SELECT a, b FROM t WHERE ID = 'someValue'",
+		"SELECT a, b FROM t WHERE ID = 1",
+		"SELECT a, b FROM t WHERE ID = 1.0",
```
Lagovas: more important to check cases with >1 blockable values %%value%%
can we check queries like `where a= %%value%% and b=%%value%% or c is Null` ?

PR #239, opened 2018-08-09
rotate acrastructs in files [T672]

Lagovas:
* added new acra-rotate utility
* extended KeyStore interface with new method RotateZoneKey()(NewPublic []byte, err error)
* updated existed and test key stores
* extended `tests/generate_acrastruct.py` utility to support AcraStruct generation with zones (was need for manual testing:)
* refactored integration tests to use a separate folder for keys in tests instead default value `.acrakeys` (tired when it override my local test .acrakeys folder)
* added force connection.Close() on acra-server to fix test failure on buildbot (when count of expected connections different from actual count)

Acra-Rotate now support only rotation of AcraStructs that stored in files and encrypted with zones
It has 2 args:
* `--keys_dir=<path_to_folder>`- folder with private keys of zones
* `--file_map_config=<path_to_json_file>` - json config in format
```
{
    "zone_id_1": "/path/to/file",
    "zone_id_2": [
        "/path/to/file1",
        "/path/to/file2"
    ]
}
```
It support value as string (to simplify case when always only one file encrypted per zone key without need to wrap it into array) and as list of strings (when >1 files was encrypted with one key)
On success acra-rotate output result in next json format:
```
{
    "zone_id_1": {
        "new_public_key": "base64 public key",
        "file_paths": [
            "/path/to/only/one/file"
        ]
    },
    "zone_id_2": {
        "new_public_key": "base64 public key",
        "file_paths": [
            "/path/to/file1",
            "/path/to/file2"
        ]
    }
}
```
P.S. Now it is the user's responsibility to backup keys and files for case of failure


vixentael:
> refactored integration tests to use a separate folder for keys in tests instead default value .acrakeys (tired when it override my local test .acrakeys folder)

So true.

vixentael:
Basically, AcraRotate knows nothing about the database, and works with files, right?

vixentael:
Ok, makes sense

vixentael:
I think it will be useful to add backup option in future

PR #242, opened 2018-08-10
AcraTranslator startup, and logs

vixentael:
- AT: start only gRPC connection by default
- AT, AS, AC: unify start up logs


vixentael:
@shadinua log examples can be found here ->

https://ph.cossacklabs.com/T673#11482

PR #243, opened 2018-08-10
AcraCensor: normalize queries and fix QueryCaptureHandler

vixentael:
1. **QueryCaptureHandler** was storing in the `query_capture` log file redacted queries (which is good), but was comparing non-redacted queries during `CheckQuery` (which is bad and leads to duplication of the same redacted queries in the `query_capture` log file). 

**Fixed:** now QueryCaptureHandler uses redacted queries for checking and writing/reading from `query_capture` log file.

2. **QueryCaptureHandler** used `IsForbidden` as keyword for queries-that-should-be-blocked-some-day-via-WebConfig, which lead to users confusion (because `IsForbidden` doesn't indicate state of query after passing the AcraCensor).

Numerous tests were updated to handle multiple cases and redacted queries.renamed to `_blacklisted_by_web_config ` which indicates that this is private field and points on its future use.

Numerous tests were updated to redacted queries and new param name.

3. ⚠️**Whitelist handler, blacklist handler and query ignore handler have case-sensitive strings checks**, which leads to numerous configuration mistakes.

F.e. blacklisting query `SELECT * FROM Z` will result in allowing its lower-case version `select * from Z`.

Moreover, the semicolon at the end of the query matters.

F.e. blacklisting query `SELECT * FROM Z;` _(semicolon)_ will result in allowing its lower-case version `SELECT * FROM Z` _(no semicolon)_.

Because SQL queries contains values and table/column names that can be case-sensitive, we can't just "lowercase/uppercase" entire SQL statement. 

**Fixed**: I decided to use "normalizedQuery" as query-to-operate-with for `CheckQuery`/`AddQuery`/`RemoveQuery` functions of W, B, I handlers. I stripped comments while normalizing

Numerous tests were updated to handle different query cases.

### DIFF
--- a/acra-censor/acra-censor_test.go
+++ b/acra-censor/acra-censor_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	"fmt"
 	"github.com/cossacklabs/acra/acra-censor/handlers"
 	"github.com/cossacklabs/acra/utils"
 )
@@ -67,6 +68,25 @@ func TestWhitelistQueries(t *testing.T) {
 	}
 	//ditto
 	err = acraCensor.HandleQuery("INSERT INTO SalesStaff1 VALUES (1, 'Stephen', 'Jiang');")
+	if err != handlers.ErrQueryNotInWhitelist {
+		t.Fatal(err)
+	}
+
+	//acracensor should NOT block this query because its the same as in whitelist, but lower-cased and without ";"
+	lowerCaseWhiteListedQuery := "select * from STUDENT"
+	err = acraCensor.HandleQuery(lowerCaseWhiteListedQuery)
+	if err != nil {
+		t.Fatal(err)
+	}
+	//acracensor should NOT block this query because its the same as in whitelist, but lower-cased and without ;
+	err = acraCensor.HandleQuery("select EMP_ID, LAST_NAME from EMPLOYEE where CITY = 'Seattle' order BY EMP_ID")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	whitelistHandler.RemoveQueries([]string{lowerCaseWhiteListedQuery})
+	err = acraCensor.HandleQuery(lowerCaseWhiteListedQuery)
+	//now acracensor should block this query because it is not in whitelist anymore
 	if err != handlers.ErrQueryNotInWhitelist {
 		t.Fatal(err)
 	}
@@ -523,7 +543,8 @@ func TestBlacklistQueries(t *testing.T) {
 			t.Fatal(err)
 		}
 	}
-	testQuery := "INSERT INTO Customers (CustomerName, City, Country) VALUES ('Cardinal', 'Stavanger', 'Norway');"
+
+	testQuery := "insert INTO Customers (CustomerName, City, Country) VALUES ('Cardinal', 'Stavanger', 'Norway');"
 	blacklist.AddQueries([]string{testQuery})
 	err = acraCensor.HandleQuery(testQuery)
 	//acracensor should block this query because it's in blacklist
@@ -996,7 +1017,7 @@ func TestQueryIgnoring(t *testing.T) {
 		}
 	}
 }
-func TestSerialization(t *testing.T) {
+func TestSerializationOnUniqueQueries(t *testing.T) {
 	testQueries := []string{
 		"SELECT Student_ID FROM STUDENT;",
 		"SELECT * FROM STUDENT;",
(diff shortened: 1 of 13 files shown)


acra-censor/handlers/querycapture_handler.go
```diff
@@ -29,10 +29,10 @@ type QueryCaptureHandler struct {
 	logger               *log.Entry
 }
 
-// QueryInfo describes Query and its status (was forbidden or allowed).
+// QueryInfo describes Query and if it was blocked by external source.
 type QueryInfo struct {
-	RawQuery    string
-	IsForbidden bool
+	RawQuery    string `json:"RawQuery"`
```
Lagovas: in json usually used field names in lowercase. so will be better if we will use `raw_query`/`_blocked_by_web_config` names

acra-censor/acra-censor_test.go
```diff
@@ -1070,30 +1090,78 @@ func TestSerialization(t *testing.T) {
...
+	}
+	if len(handler.GetAllInputQueries()) != numOfUniqueQueries {
+		t.Fatal("Expected to have " + fmt.Sprint(numOfUniqueQueries) + " unique queries. \n Got:" + strings.Join(handler.GetAllInputQueries(), " | "))
+	}
+	for index, query := range handler.GetAllInputQueries() {
+		if strings.EqualFold(testQueries[index], query) {
+			t.Fatal("Expected: " + testQueries[index] + "\nGot: " + query)
+		}
+	}
+	if err = os.Remove(tmpFile.Name()); err != nil {
```
Lagovas: better to defer file removing after creation because it will exists after any `t.Fatal` call above

acra-censor/handlers/blacklist_handler.go
```diff
@@ -188,17 +188,25 @@ func (handler *BlacklistHandler) Release() {
 	handler.Reset()
 }
 
-// AddQueries adds queries to the list that should be blacklisted
+// AddQueries normalizes and adds queries to the list that should be blacklisted
 func (handler *BlacklistHandler) AddQueries(queries []string) {
 	for _, query := range queries {
-		handler.queries[query] = true
+		normalizedQuery, _, err := NormalizeAndRedactSQLQuery(query)
```
Lagovas: why we use redacted queries here? what if we will need block all queries `where user_id=1` or `where is_superuser=True`? it will add `where user_id=::replaced::` as I understand

acra-censor/handlers/whitelist_handler.go
```diff
@@ -216,17 +216,25 @@ func (handler *WhitelistHandler) Release() {
 	handler.Reset()
 }
 
-// AddQueries adds queries to the list that should be whitelisted
+// AddQueries normalizes and adds queries to the list that should be whitelisted
 func (handler *WhitelistHandler) AddQueries(queries []string) {
 	for _, query := range queries {
-		handler.queries[query] = true
+		normalizedQuery, _, err := NormalizeAndRedactSQLQuery(query)
```
Lagovas: same as with blacklist. I think we should use non-redacted query

acra-censor/handlers/blacklist_handler.go
vixentael: it's not redacted, it's normalized.

to make `where user_id=1` same as `WHERE user_id=1`

acra-censor/handlers/whitelist_handler.go
vixentael: it's not redacted, it's normalized ;)

acra-censor/acra-censor_test.go
vixentael: totally makes sense, added everywhere

acra-censor/handlers/querycapture_handler.go
vixentael: changed

vixentael:
@Lagovas 
to fix circleCI tests, I merged `acra-server` from master to this branch
https://github.com/cossacklabs/acra/pull/243/files#diff-64a96ce100f1538851ad657696905b9e

and updated acra-censor config files with more database-specific queries
https://github.com/cossacklabs/acra/pull/243/files#diff-371471da67ad823fdcaf465ede92cc49

and attempted to fix (i hope it worked) encoding issue with
https://github.com/cossacklabs/acra/pull/243/files#diff-b131fe2646b3e4980d06f90692dc2401

```
======================================================================
ERROR: testDetect (__main__.TestCheckLogPoisonRecord)
----------------------------------------------------------------------
  File "tests/test.py", line 1615, in testDetect
    self.assertIn('Check poison records', self.read_log(self.acra))
  File "tests/test.py", line 1527, in read_log
    print(log)
UnicodeEncodeError: 'ascii' codec can't encode character '\ufffd' in position 75907: ordinal not in range(128)
```

PR #245, opened 2018-08-13
fix %%value%% and %%column%% patterns

Lagovas:


vixentael:
Thank you for improving AcraCensor and tests.

P.S. I believe that merging with master was a bad idea :D

PR #246, opened 2018-08-13
Remove repetitive and redundant logs

vixentael:
Log queries during decryption by AcraCensor if enabled, or only in Debug mode.

Actually, this is an attempt to optimize too often usage of `NormalizeAndRedact` function.

### DIFF
--- a/acra-censor/acra-censor_implementation.go
+++ b/acra-censor/acra-censor_implementation.go
@@ -65,6 +65,7 @@ func (acraCensor *AcraCensor) ReleaseAll() {
 // HandleQuery processes every query through each handler.
 func (acraCensor *AcraCensor) HandleQuery(query string) error {
 	if len(acraCensor.handlers) == 0 {
+		// no handlers, AcraCensor won't work
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


acra-censor/acra-censor_implementation.go
```diff
@@ -79,8 +80,8 @@ func (acraCensor *AcraCensor) HandleQuery(query string) error {
 		}
 		continueHandling, err := handler.CheckQuery(normalizedQuery)
 		if err != nil {
+			// continue to next handler
 			if err == handlers.ErrQuerySyntaxError && acraCensor.ignoreParseError {
-				acraCensor.logger.WithError(err).Infof("Parsing error on query (first %v symbols): %s", handlers.LogQueryLength, handlers.TrimStringToN(queryWithHiddenValues, handlers.LogQueryLength))
```
Lagovas: how we will now that some handler can't process query? why not log error?

Lagovas: agree. didn't notice `ErrQuerySyntaxError` error type

PR #248, opened 2018-08-14
extended acracensor

Lagovas:


vixentael:
thank you @storojs72 and @Lagovas for a great job!

PR #250, opened 2018-08-14
refactoring utils and deadlines [T696]

Lagovas:
* drop utils.AbsPath and use filepath.Abs
* drop utils.FindTag and use bytes.Index
* use SetDeadline where it possible
* use uint64 instead int in decryptors


vixentael:
i suggest merging, if @secumod will have time, he will look at this

PR #251, opened 2018-08-17
metrics for acratranslator [T760, T662]

Lagovas:
* added metrics to translator (time of connection life, time of processing requests, acrastruct decryption counters). for it was added listener/connection wrappers that wrap accept method (for grpc requests when we have only one way to inject our logic) that wrap connections to track their lifetime because then we have not access to them. maybe in a future will be better to move these wrappers from `cmd/acra-translator/network.go` to `network` and use in other places...
* added tests for metrics and check that all expected metric names are exported from services
* added logging metrics in tests (by environment variable TEST_LOG_METRICS or LOG_METRICS config in tests). default `off` to not increase time of tests without need (can be useful in debugging)
* don't turn on prometheus http handler if address wasn't set
* renamed metrics from `<metric_name>_bucket` to `<metric_name>` if it histogram because prometheus client do itself (for example `acraconnector_connections_processing_seconds_bucket` -> `acraconnector_connections_processing_seconds`)
* some golint fixes and reduce limit to 150


decryptor/mysql/response_proxy.go
```diff
@@ -101,41 +101,22 @@ const (
 
 // MySQL types
 const (
-	MYSQL_TYPE_NEWDECIMAL byte = iota + 0xf6
```
vixentael: we don't use these constants?

Lagovas: yes. maybe I copied them earlier for the future)) but now we don't need them and too lazy was to rename all of them to go style)

decryptor/postgresql/pg_decryptor.go
```diff
@@ -205,15 +205,15 @@ func checkInlinePoisonRecordInBlock(block []byte, decryptor base.Decryptor, logg
 		logger.Debugln("Check poison records")
 		currentIndex := 0
 		for {
-			if index, _ := decryptor.BeginTagIndex(block[currentIndex:]); index == utils.NotFound {
-				return nil
-			} else {
+			if index, _ := decryptor.BeginTagIndex(block[currentIndex:]); index != utils.NotFound {
```
vixentael: i would change it

```
if ... {
  return nil
}
...
currentIndex++
```

decryptor/mysql/response_proxy.go
vixentael: ok :))

decryptor/postgresql/pg_decryptor.go
Lagovas: we need `index` value that added to `currentIndex`. Now it in `if expression`. To do like you want we need to move assignment from if block to outer. then we can use this value
```
index, _ := decryptor.BeginTagIndex(block[currentIndex:])
if index == utils.NotFound {
   return nil 
}
currentIndex += index
.....
```
this variant is better?

vixentael: yes, i prefer this updated version

PR #253, opened 2018-08-30
validate clientid packet length [T696]

Lagovas:
added validation of first packet from acra-connector to acra-server that send before secure session packets and has limited length that we can check (client id can have 255 symbols as length plus some overhead with length of packet). set 1kb limit

p.s. plus found old local code with extended tests for acra-censor that forget to add to some previous pr. added to this pr because not so much and simple changes


vixentael:
AcraCensor now can handle delete patterns as well?

Lagovas:
it support any `where` expressions. it uses in `select`/`delete`/`update` queries. tests just check and prove it)

PR #254, opened 2018-08-31
update default tls config

Lagovas:
* use tlsv1.2 (highest supported version in golang now) and limited cipher suit that are recommended by https://www.iana.org/assignments/tls-parameters/tls-parameters.xml and https://www.owasp.org/index.php/TLS_Cipher_String_Cheat_Sheet, and that used in tls1.3
* use tlsv1.0 and extended cipher suit list for mysql tls connections due to limited support from mysql db and mysql drivers (for example mysql-connector-python support only tls1.0)


vixentael:
It affects AcraServer, AcraTranslator, AcraConnector, right?

Lagovas:
only server and connector. translator now support only secure session transport protocol and doesn't use tls (only internally in grpc that not affected by this pr)

network/tls_wrapper.go
```diff
@@ -19,15 +19,24 @@ package network
...
 
+// allowedCipherSuits that set in default tls config
+var allowedCipherSuits = []uint16{
+	tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
+	tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
+	tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+	tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+	tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+	tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+}
```
secumod: Is this list sorted by priority? If yes, then I would put it a bit differently:
* TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
* TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
* TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305
* TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
* TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
* TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305

Justification:
- elliptic curves are more secure, in general faster than RSA and has smaller key sizes (hence the number of transferred bytes on the network is also lower), so they come before RSA
- AES comes before CHACHA - CHACHA is mostly useful to communicate with low-end mobile devices, which do not have AES crypto accelerators in their CPUs (non-hardware CHACHA is generally cheaper than AES and is mostly there for battery savings)
- AES-128 with SHA-256 come before AES-256 with SHA-384 (they have smaller key sizes and are usually better optimised, because they are more popular)

network/tls_wrapper.go
```diff
@@ -107,5 +116,41 @@ func NewTLSConfig(serverName string, caPath, keyPath, crtPath string, authType t
...
+
+		tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,
+		tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
+
+		tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
+		tls.TLS_RSA_WITH_AES_128_GCM_SHA256,
+		tls.TLS_RSA_WITH_AES_128_CBC_SHA256,
+		tls.TLS_RSA_WITH_AES_256_CBC_SHA,
+		tls.TLS_RSA_WITH_AES_128_CBC_SHA,
+	}
```
secumod: Same here as above

network/tls_wrapper.go
Lagovas: agree, changed order in both cases (in second I copied order from first list and placed ecdsa before rsa in other cases)

PR #255, opened 2018-08-31
format custom fields before output cef logs

Lagovas:
when read logs noticed that more readable when fields are sorted


vixentael:
shouldn't we sort fields for json as well? :)

Lagovas:
maybe in a future. we use third-party implementation and for it we need write own or override that we use...
I added it for myself and decide add it to repo too ) now it help read logs from tests)

PR #257, opened 2018-09-11
add sqlalchemy type wrappers for python acrawriter [T712]

Lagovas:
copy wrappers from example without zone to acrawriter to provide way to just use sqlalchemy types which will implicitly encrypt with `acrawriter` instead explicit call `acrawriter.create_acrastruct` (as we did it for django orm)

p.s. now github has some issues (https://status.github.com/messages) and tests failed. just look on code. I will merge after green tests


vixentael:
Good catch!

How AcraString is working? Is it string utf8 representation of binary data?

vixentael:
great, thank you!

PR #259, opened 2018-09-17
refactor processing clients packets

Lagovas:
postgresql has 3 special message types that has format another than all other message: startup, ssl request and cancelation request. 
before we try to parse only first packet as special and all other as general
but was case when client send ssl request (first special), then if it continue after db response it send startup message. we correctly handle case when db accept ssl request and didn't handle deny from db.
Now processing re-writed for client's side of processing and we try to recognize these special messages or other known general messages and doesn't depend on their order (first, second etc). Plus in a future can process them separately. 

plus added handling columns with null values (it fixes panics [T790])

### DIFF
--- a/cmd/acra-connector/acra-connector.go
+++ b/cmd/acra-connector/acra-connector.go
@@ -160,10 +160,11 @@ func handleConnection(config *Config, connection net.Conn) {
 	}
 	log.Infoln("Close wrapped connection with AcraServer")
 	if err := acraConnWrapped.Close(); err != nil {
-		log.WithError(err).Errorln("Error on closing wrapped connection with AcraServer")
+		log.WithError(err).Errorf("Error on closing wrapped connection with %s", connector_mode.ModeToServiceName(config.Mode))
+
 	}
 	if err := acraConn.Close(); err != nil {
-		log.WithError(err).Errorln("Error on closing connection with AcraServer")
+		log.WithError(err).Errorln("Error on closing connection with %s", connector_mode.ModeToServiceName(config.Mode))
 	}
 }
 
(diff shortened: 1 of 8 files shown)


cmd/acra-connector/acra-connector.go
```diff
@@ -122,15 +127,16 @@ func handleConnection(config *Config, connection net.Conn) {
...
 		return
 	}
-	defer acraConn.Close()
 
 	acraConnWrapped, err := config.ConnectionWrapper.WrapClient(config.ClientID, acraConn)
 	if err != nil {
 		log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCantWrapConnection).
 			Errorln("Can't wrap connection")
+		if err := acraConn.Close(); err != nil {
+			log.WithError(err).Errorln("Error on closing connection with AcraServer")
```
vixentael: It might be AcraServer or AcraTranslator.

Shall we change log message to "AcraServer/AcraTranslator" or (better) check AcraConnector's `mode` and use appropriate naming.

cmd/acra-connector/acra-connector.go
```diff
@@ -151,7 +157,13 @@ func handleConnection(config *Config, connection net.Conn) {
 			log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCantStartConnection).
 				Errorln("Connector error")
 		}
-		return
+	}
+	log.Infoln("Close wrapped connection with AcraServer")
+	if err := acraConnWrapped.Close(); err != nil {
+		log.WithError(err).Errorln("Error on closing wrapped connection with AcraServer")
```
vixentael: log warning – same as above

vixentael:
Cool that you've added unit tests to catch this!

PR #261, opened 2018-09-19
Add parameter to generate CLI agrs in markdown

Lagovas:
in a past was added logic to generate description of cli args in markdown format as table and was leaved as is with hardcoded paths. It was used for our documentation in a past

I refactored it. Let discuss do we need it now and do we want to add it to repo

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -40,8 +40,9 @@ import (
 )
 
 var (
-	config     = flag_.String("config_file", "", "path to config")
-	dumpconfig = flag_.Bool("dump_config", false, "dump config")
+	config                   = flag_.String("config_file", "", "path to config")
+	dumpconfig               = flag_.Bool("dump_config", false, "dump config")
+	generateMarkdownArgTable = flag_.Bool("generate_markdown_args_table", false, "Generate with yaml config markdown text file with descriptions of all args")
 )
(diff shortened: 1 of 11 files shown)


cmd/utils.go
```diff
@@ -189,8 +190,10 @@ func GenerateMarkdownDoc(output io.Writer, serviceName string) {
 	escapeColumn := func(text string) string {
 		return strings.Replace(text, "|", "\\|", -1)
 	}
-	// serviceName | arg name | rename to | default value | description
-	fmt.Fprintf(output, "|%v|||||\n", serviceName)
+	// table header with service name
+	// |serviceName | arg name | rename to | default value | description|
+	// |:-:         |:-:       |:-:        |:-:            |:-:         |
+	fmt.Fprintf(output, "|%v|||||\n|:-:|:-:|:-:|:-:|:-:|\n", serviceName)
 	flag_.CommandLine.VisitAll(func(flag *flag_.Flag) {
```
vixentael: "rename to" looks un-needed anymore

vixentael: just realized, that we can use this column in future, after next renaming.

vixentael:
Thank you, let's keep this flag for now (we can remove it anytime).

PR #265, opened 2018-09-25
T816 Add OrderBy, GroupBy (plus Having) and Limit / Offset expressions support

storojs72:


acra-censor/handlers/handlers_util.go
```diff
@@ -171,46 +171,6 @@ func handleSelectPattern(queryNodes, patternNodes []sqlparser.SQLNode) bool {
 	return false
 }
 
-// isColumnPattern return true if this SelectExpr is our %%COLUMN%% pattern
```
vixentael: why are these three functions moved?

acra-censor/handlers/handlers_util.go
```diff
@@ -251,6 +210,13 @@ func handleSelectColumnPattern(queryNodes, patternNodes []sqlparser.SQLNode) boo
 			}
 			// check other nodes on equal (except SelectExprs)
 		} else if !reflect.DeepEqual(patternNode, queryNode) {
+			if patternOrderBy, ok := patternNode.(sqlparser.OrderBy); ok {
```
vixentael: can we rewrite this piece into one function? something like

`func ifPatternOrderBy(patternNode, queryNode)`

acra-censor/handlers/handlers_util.go
```diff
@@ -283,6 +249,105 @@ func handleSelectWherePattern(queryNodes, patternNodes []sqlparser.SQLNode) bool
 	return true
 }
 
+// handleStarPattern handles SELECT * FROM table %%WHERE%% pattern
+func handleStarPattern(queryNodes, patternNodes []sqlparser.SQLNode) bool {
```
vixentael: why this function is moved?

acra-censor/handlers/handlers_util.go
```diff
@@ -283,6 +249,105 @@ func handleSelectWherePattern(queryNodes, patternNodes []sqlparser.SQLNode) bool
...
+			return true
+		}
+		return false
+	}
+	//this is a case when pattern == query
+	return true
+}
+
+// handleWherePatterns try to match all WHERE conditions with supported patterns
+func handleWherePatterns(queryNodes, patternNodes []sqlparser.SQLNode) bool {
```
vixentael: why this function is moved?

acra-censor/handlers/handlers_util.go
vixentael: actually we can move first and second `if` checks into `matchOrderBy` function

vixentael:
Thank you, overall I like it, it's a great addition to AcraCensor logic.

However, I didn't get the reason of moving functions inside file. Can you explain please?

If I understood correctly, the "real" changes are way smaller than 400 lines, but rather ~100 lines, most changes are caused by moving functions.

acra-censor/handlers/handlers_util.go
storojs72: It's just only move without any logic change. For better readability. Now it will be:
// pattern handlers
- handleSelectPattern
- handleSelectColumnPattern
- handleSelectWherePattern
- handleStarPattern
- handleWherePatterns

// helper methods
- isColumnPattern
- isValueReplacer
- isValuePattern
- isListOfValuesPattern
- IsEqualComparisonNodes

// matchers
- matchValuePattern
- matchSubqueryPattern
- matchQueryNodeWithValuePattern
- matchRangeCondition
- matchOrderBy

And then other utilility functions

acra-censor/handlers/handlers_util.go
storojs72: It seems that OrderBy is not a separate pattern - it's part of %%VALUE%%. Yes we can move type assertions to matchOrderBy - but we process types of nodes inside pattern handlers - not in comparers

acra-censor/handlers/handlers_util.go
storojs72: just for readability, ditto

acra-censor/handlers/handlers_util.go
storojs72: ditto

acra-censor/handlers/handlers_util_test.go
```diff
@@ -536,5 +536,86 @@ func TestPatternsInWhereClauses(t *testing.T) {
...
+	parsedPatterns, err := ParsePatterns(patterns)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	matchableQueries := [][]string{
+		[]string{
+			"SELECT a1 FROM table1 ORDER BY column1",
+			"SELECT a1 FROM table1 ORDER BY 1",
+			"SELECT a1 FROM table1 ORDER BY (select priority from ordering o where o.val = e.name)",
```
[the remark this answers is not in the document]
storojs72: Yeah - its rare but possible. (http://www.java2s.com/Code/SQL/Select-Clause/Orderbyvaluefromsubquery.htm)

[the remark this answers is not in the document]
storojs72: 1 - is also not a column, but value that means first column

acra-censor/handlers/handlers_util.go
vixentael: got it!

acra-censor/handlers/handlers_util.go
vixentael: It's up to you then.

I'd prefer hiding if statements somewhere inside function.

Lagovas: use switch instead if else and then it should be
```
switch patternNode.(type){
case *sqlparser.SelectExprs:
...
case sqlparser.OrderBy:
...
default: reflect.DeepEqual(...)
 }
```
we don't need to compare via deepequal before type check on OrderBy struct

acra-censor/handlers/handlers_util.go
```diff
@@ -358,6 +491,66 @@ func handleRangeCondition(patternNode, queryNode *sqlparser.RangeCond) bool {
...
+
+	if len(patternNode) != len(queryNode) {
+		return false
+	}
+	for index, _ := range patternNode {
+		if !strings.EqualFold(patternNode[index].Direction, queryNode[index].Direction) {
+			return false
+		}
+		if !strings.EqualFold(sqlparser.String(patternNode[index].Expr), ColumnConfigPlaceholderReplacer) {
+			return false
```
Lagovas: we doesn't need sqlparser.String, better if we will get value from node with explicit cast to correct type

acra-censor/handlers/handlers_util.go
storojs72: Thanks. Usefully.

acra-censor/handlers/handlers_util.go
storojs72: Thanks. Agree

acra-censor/handlers/handlers_util.go
```diff
@@ -283,76 +283,125 @@ func handleSelectWherePattern(queryNodes, patternNodes []sqlparser.SQLNode) bool
...
+		patternNode := patternTopNodes[i]
+		queryNode := queryTopNodes[i]
+
+		switch patternNode.(type) {
+		case *sqlparser.Limit:
+			if !matchLimit(querySelect.Limit, patternSelect.Limit) {
+				return false
+			}
+		default:
+			if !reflect.DeepEqual(queryNode, patternNode) {
```
Lagovas: we don't need to validate other fields of "select" strict here, only limit. we can just accept "sqlparser.Limit" type at function and work only with them and pass them from top function

acra-censor/handlers/handlers_util.go
```diff
@@ -548,51 +596,274 @@ func IsEqualComparisonNodes(patternNode, queryNode *sqlparser.ComparisonExpr) bo
...
+}
+
+// matchOrderBy handles order by construction
+// return true if match otherwise false
+func matchOrderBy(patternNode, queryNode sqlparser.OrderBy) bool {
+	if reflect.DeepEqual(patternNode, queryNode) {
+		return true
+	}
+
+	if len(patternNode) != len(queryNode) {
```
Lagovas: we don't need the check of length because in a future here may be list of columns or values pattern and length of pattern nodes will not equal to length of query nodes

storojs72: Lets remove it when we infroduce such placeholder. Now it seems correct to check that amounts of column placeholders equals to actual amount of columns in query

acra-censor/handlers/handlers_util.go
storojs72: In such case it will accept any expressions except limit

Lagovas: yes and we don't need to check twice other fields that can be checked in top function that call this. this function should expect only Limit and work only with this. and caller should pass correct type and work with it. it should not pass anything except limit, because other fields will be checked before (like with handleWherePatterns or something else)

Lagovas:
I looked to our code of pattern processing and i think we need refactoring in a future because do a lot of unnecessary work between checks

acra-censor/handlers/handlers_util_test.go
```diff
@@ -536,5 +536,287 @@ func TestPatternsInWhereClauses(t *testing.T) {
 			}
 		}
 	}
+}
+
+func TestOrderByWithColumnPattern(t *testing.T) {
+	patterns := []string{
+		// ORDER BY with column placeholder
+		fmt.Sprintf("SELECT a1 FROM table1 ORDER BY %s", ColumnConfigPlaceholder),
```
Lagovas: please add one more pattern:
select a from b order by a, %%column%%, 1, %%column%%

match queries:
1. select a from b order by a, 2, 1, ABC

not match:
1. select a from b order by 1, abc, 1, abs
2. select a from b order by a, abc, a, ABC
3. select a from b order by a, abc, 1, ABC, ABC

I think it will not pass now because in processing order by we doesn't use reflect DeepEqual after pattern mismatch in a loop

storojs72:
Yes. Agree. It's good that we have a lot of unit tests now

acra-censor/handlers/handlers_util_test.go
storojs72: Agree. Thanks. It seems that the same it true for GroupBy

storojs72:
Thanks @Lagovas for very careful look at this PR, even being in vacation!

PR #268, opened 2018-10-08
T824 censor refactoring

storojs72:


acra-censor/handlers/handlers_util.go
```diff
@@ -125,711 +171,1558 @@ func NormalizeAndRedactSQLQuery(sql string) (normalizedQuery string, redactedQue
...
-	return false
-}
-
-// handleSelectPattern handles %%SELECT%% pattern
-func handleSelectPattern(queryNodes, patternNodes []sqlparser.SQLNode) bool {
-	if reflect.TypeOf(queryNodes[0]) == reflect.TypeOf(patternNodes[0]) {
-		if patternNodeSelect, ok := patternNodes[0].(*sqlparser.Select); ok && strings.EqualFold(sqlparser.String(patternNodeSelect.SelectExprs), SelectConfigPlaceholderReplacerPart2) {
-			return true
-		}
+	match = handleSetStatement(query, pattern)
```
Lagovas: 1. the same as above with abstract type for methods, better to use loop with abstract type like `type handleFunction func(query, pattern sqlparser.SQLNode) bool`

2. i think we don't need to add specific function for each type of statement or struct field, only for which we process specially. all other statements and fields should processed with some default logic. if we add method for each statement and process it as 20 other same statements, we add a lot of unnecessary code that we should read and maintain. now we have specific functions only for our custom patterns. for example if i read and see something like `processWhereClause(query, pattern)` - I think that we process it with some specific logic. when i see like `reflect.DeepEqual(query, pattern) - I understand that we doesn't support or haven't special logic for such statements and no matter which values it has

acra-censor/handlers/handlers_util.go
```diff
@@ -125,711 +171,1558 @@ func NormalizeAndRedactSQLQuery(sql string) (normalizedQuery string, redactedQue
...
 	}
-	queryTopNodes, err := getTopNodes(queryNodes[0])
-	if err != nil {
+	match = matchSelectDistinct(querySelectNode.Distinct, patternSelectNode.Distinct)
+	if !match {
 		return false
 	}
-	patternTopNodes, err := getTopNodes(patternNodes[0])
-	if err != nil {
+	match = matchSelectHints(querySelectNode.Hints, patternSelectNode.Hints)
```
Lagovas: as said above I prefer to add some general function that handle statements that we don't care like
```
func defaultMatch (query, pattern sqlparser.SQLNode) bool{
   return reflect.DeepEqual(query, pattern)
}
```
and then write like
```
type comparePair struct {
  pattern sqlparser.SQLNode, 
  query sqlparser.SQLNode
}
for _, pair := range []comparePair{{patternSelectNode.Distinct, querySelectNode.Distinct}, {patternSelectNode.Hints, querySelectNode.Hints}, {patternSelectNode.Comments, querySelectNode.Comments}}{
  if !defaultMatch(pair.pattern, pair.query){ return false}
}
```
and then call specific functions for correct fields as below for `SelectExpr`, `Where`, `From`

acra-censor/handlers/handlers_util.go
```diff
@@ -125,711 +171,1558 @@ func NormalizeAndRedactSQLQuery(sql string) (normalizedQuery string, redactedQue
...
+		return false
+	}
+	match = matchSelectFrom(querySelectNode.From, patternSelectNode.From)
+	if !match {
+		return false
+	}
+	match = matchSelectWhere(querySelectNode.Where, patternSelectNode.Where)
+	if !match {
+		// Check %%WHERE%% pattern
+		if isWherePattern(patternSelectNode.Where) {
```
Lagovas: imo better to check most general pattern first instead concrete. it's easier to check simple pattern that avoid other checks of conditions in `where`. after that, it has sense to check more concrete and complex patterns

acra-censor/acra-censor_test.go
```diff
@@ -414,7 +414,7 @@ func testWhitelistColumnsPattern(t *testing.T) {
 	acceptableQueries = []string{
 		"SELECT a1 FROM table1 GROUP BY a2 HAVING COUNT(a3) > 0",
 		"SELECT a1 FROM table1 GROUP BY a2 HAVING COUNT(a2) > 1000",
-		"SELECT a1 FROM table1 GROUP BY a2 HAVING COUNT(a1) > (select 1)",
```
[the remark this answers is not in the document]
storojs72: Because we decided some time ago, that we do not consider query as value (https://github.com/cossacklabs/acra/blob/master/acra-censor/handlers/handlers_util.go#L677)

acra-censor/handlers/blacklist_handler.go
```diff
@@ -35,7 +35,7 @@ func NewBlacklistHandler() *BlacklistHandler {
 	handler := &BlacklistHandler{}
 	handler.queries = make(map[string]bool)
 	handler.tables = make(map[string]bool)
-	handler.patterns = make([][]sqlparser.SQLNode, 0)
+	handler.patterns = make([]sqlparser.Statement, 0)
```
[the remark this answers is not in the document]
storojs72: because sqlparser.Parse - returns Statement which is itself SQLNode.

acra-censor/acra-censor_test.go
```diff
@@ -339,7 +339,7 @@ func testWhitelistColumnsPattern(t *testing.T) {
 		// different table
 		"SELECT * FROM testTable1 ORDER BY 1",
 		// different columns in select expressions
-		"SELECT A FROM testTable ORDER BY (case when f1 then 1 when f1 is null then 2 else 3 end)",
+		//"SELECT A FROM testTable ORDER BY (case when f1 then 1 when f1 is null then 2 else 3 end)",
```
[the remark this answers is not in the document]
storojs72: Because this query shouldn't be blocked by whitelist (and same is true for other queries with defined columns when pattern contains '*' in SelectExprs). We have:

pattern - "SELECT * FROM testTable ORDER BY %%COLUMN%%"
query - "SELECT A FROM testTable ORDER BY (case when f1 then 1 when f1 is null then 2 else 3 end)"

and * means that we accept any column in SelectExprs. So if query matches pattern, whitelist should pass it. Please, look at this: https://github.com/cossacklabs/acra/blob/master/acra-censor/acra-censor_test.go#L548 - pattern with * matches query with a, b columns and whitelist passes it

acra-censor/handlers/common.go
```diff
@@ -24,28 +24,52 @@ import (
 )
 
 // ParsePatterns replace placeholders with our values which used to match patterns and parse them with sqlparser
-func ParsePatterns(patterns []string) ([][]sqlparser.SQLNode, error) {
-	placeholders := []string{SelectConfigPlaceholder, ColumnConfigPlaceholder, WhereConfigPlaceholder, ValueConfigPlaceholder, ListOfValuesConfigPlaceholder, SubqueryConfigPlaceholder}
-	replacers := []string{SelectConfigPlaceholderReplacer, ColumnConfigPlaceholderReplacer, WhereConfigPlaceholderReplacer, ValueConfigPlaceholderReplacer, ListOfValuesConfigPlaceholderReplacer, SubqueryConfigPlaceholderReplacer}
+func ParsePatterns(patterns []string) ([]sqlparser.Statement, error) {
+	placeholders := []string{
```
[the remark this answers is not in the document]
storojs72: Agree. Thanks

acra-censor/handlers/handlers_util.go
storojs72: if %%WHERE%% true - we should stop further checks and return true

acra-censor/handlers/handlers_util.go
storojs72: I deliberately go out from reflect.DeepEqual - because you said me somehow that it is very cost system call. Currently many false cases are detected by matching types (when types are not equal) - and It seems I used reflect.DeepEqual only two-three times in code. Also such skeleton logic covers all sql cases that sqlparser library can suggest.

storojs72: And also it seems it's very simple to add specific checks, patterns if we will need

PR #269, opened 2018-10-10
dry run rotate [T813]

Lagovas:
`acra-rotate` with flag `--dry-run` will fetch acrastructs (from files or database), decrypt, rotate in memory keys, encrypt with new public keys and print result json with new public keys without saving rotated keys and acrastructs (to filesystem or database)

### DIFF
--- a/cmd/acra-rotate/dbRotation.go
+++ b/cmd/acra-rotate/dbRotation.go
@@ -36,13 +36,13 @@ func rotateDb(selectQuery, updateQuery string, db *sql.DB, keystore keystore.Key
 
 	rows, err := db.Query(selectQuery)
 	if err != nil {
-		log.WithError(err).Errorf("Can't fetch result")
+		log.WithError(err).Errorf("Can't fetch result with sql_select query")
 		return false
 	}
 	defer rows.Close()
 	columns, err := rows.Columns()
 	if err != nil {
-		log.WithError(err).Errorln("Can't fetch info about result columns")
+		log.WithError(err).Errorln("Can't fetch metadata for result columns")
 		return false
 	}
 	if len(columns) < 2 {
@@ -97,7 +97,7 @@ func rotateDb(selectQuery, updateQuery string, db *sql.DB, keystore keystore.Key
 		if !dryRun {
 			_, err = db.Exec(updateQuery, extraArgs...)
 			if err != nil {
-				logger.WithError(err).Errorln("Can't update data in db via update query")
+				logger.WithError(err).Errorln("Can't update data in db via sql_update query")
 				return false
 			}
 		}
--- a/cmd/acra-rotate/dbRotation.go
+++ b/cmd/acra-rotate/dbRotation.go
@@ -19,90 +19,21 @@ package main
 import (
 	"database/sql"
 	"encoding/hex"
-	"encoding/json"
 	"fmt"
-	"github.com/cossacklabs/acra/acra-writer"
-	"github.com/cossacklabs/acra/decryptor/base"
 	"github.com/cossacklabs/acra/keystore"
 	"github.com/cossacklabs/acra/utils"
-	"github.com/cossacklabs/themis/gothemis/keys"
 	log "github.com/sirupsen/logrus"
 	"reflect"
 )
 
-type keyPair struct {
-	oldPrivatekey *keys.PrivateKey
-	NewPublicKey  *keys.PublicKey
-}
-
-// keyRotationStore store previous privateKeys and new public keys after rotation to use same private/public keys during rotation
-type keyRotationStore struct {
-	keys     map[string]*keyPair
-	keystore keystore.KeyStore
-}
-
-// Marshal encode new public keys for zones to json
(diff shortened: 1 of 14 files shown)


cmd/acra-rotate/dbRotation.go
```diff
@@ -152,43 +79,36 @@ func rotateDb(selectQuery, updateQuery string, db *sql.DB, keystore keystore.Key
 			return false
 		}
 		logger := log.WithFields(log.Fields{"ZoneId": string(acraStructID)})
-		logger.Infof("Rotate AcraStruct with ZoneId=%s", string(acraStructID))
+		logger.Infof("Rotate AcraStruct")
```
vixentael: why we don't add zone id anymore?

vixentael: I think we should log smth like `Rotating in dry-run mode` if `dry-run` is `true`

cmd/acra-rotate/fileRotation.go
```diff
@@ -86,26 +79,33 @@ func rotateFiles(fileMap ZoneIDFileMap, keyStore keystore.KeyStore) (ZoneRotateR
...
 }
 
 // runFileRotation read map zones to files, re-generate zone key pairs and re-encrypt files
-func runFileRotation(fileMapConfigPath string, keystorage keystore.KeyStore) {
+func runFileRotation(fileMapConfigPath string, keystorage keystore.KeyStore, dryRun bool) {
 	fileMap, err := loadFileMap(fileMapConfigPath)
 	if err != nil {
 		log.WithError(err).Errorln("Can't load config with map <ZoneId>: <FilePath>")
 		os.Exit(1)
 	}
```
vixentael: Same, I think we should log "Starting in .. dry-run mode"

cmd/acra-rotate/rotator.go
```diff
@@ -0,0 +1,103 @@
...
+		if err := rotator.keystore.SaveZoneKeypair([]byte(zoneID), keypair); err != nil {
+			log.WithField("zoneID", zoneID).WithError(err).Errorln("Can't save rotated keypair")
+			return err
+		}
+	}
+	return nil
+}
+
+func (rotator *keyRotator) clearKeys() {
+	for _, keypair := range rotator.newKeypairs {
```
vixentael: 👍

cmd/acra-rotate/rotator.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't get private key")
+		return nil, err
+	}
+	decrypted, err := base.DecryptAcrastruct(acrastruct, privateKey, zoneID)
+	if err != nil {
+		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't decrypt AcraStruct")
+		return nil, err
+	}
+	utils.FillSlice(0, privateKey.Value)
```
vixentael: 👍

cmd/acra-rotate/rotator.go
```diff
@@ -0,0 +1,103 @@
...
+	logger.Infof("Rotate AcraStruct")
+	// rotate
+	privateKey, err := rotator.keystore.GetZonePrivateKey(zoneID)
+	if err != nil {
+		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't get private key")
+		return nil, err
+	}
+	decrypted, err := base.DecryptAcrastruct(acrastruct, privateKey, zoneID)
+	if err != nil {
+		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't decrypt AcraStruct")
```
vixentael: maybe it worth zero-ing private key in case of error too?

keystore/filesystem/server_keystore_test.go
```diff
@@ -105,26 +106,27 @@ func testGeneratingDataEncryptionKeys(store *FilesystemKeyStore, t *testing.T) {
 	}
 }
 
+func checkPath(path string, t *testing.T) {
```
vixentael: 👍

cmd/acra-rotate/dbRotation.go
Lagovas: > why we don't log zone id anymore?
because row above create logger with zone as field that always will be included in log messages

> I think we should log smth like Rotating in dry-run mode if dry-run is true
agree

cmd/acra-rotate/rotator.go
```diff
@@ -57,12 +57,13 @@ func (rotator *keyRotator) rotateAcrastruct(zoneID, acrastruct []byte) ([]byte,
 		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't get private key")
 		return nil, err
 	}
+	defer utils.FillSlice(0, privateKey.Value)
```
vixentael: 👍

vixentael:
Looks awesome!

PR #270, opened 2018-10-11
AcraWriter c++ with tests and examples

vixentael:
1. AcraWriter c++ library is a single `.hpp` file in `wrappers/cpp`. Written in C++14 standard.
2. Example is CLion project in `examples/cpp`. Example has two entry points: `main.cpp` and `main_tests.cpp`. Example uses `catch` for testing and `cppcodec` for base64/hex formats encoding.

Any suggestions how to run acrastruct tests automated?


wrappers/cpp/acrawriter.hpp
```diff
@@ -0,0 +1,124 @@
...
+          uint8_t em_length_array[8];
+          memcpy(em_length_array, &em_length, sizeof(em_length));
+
+          // convert encrypted data length to little endian
+          if (is_big_endian()) {
+            reverse(begin(em_length_array), end(em_length_array));
+          }
+
+          // zeroing symmertic key
+          fill(symmertic_key.begin(), symmertic_key.end(), 0);
```
Lagovas: maybe would be good to fill with zeroes `temp_private_key` too?

vixentael: agree

PR #272, opened 2018-10-16
add concurrent access in lru cache

Lagovas:
* noticed that lru cache doesn't guarantee safe concurrent access. added mutex https://github.com/golang/groupcache/blob/master/lru/lru.go#L22
* use formatted logging functions when used placeholders in text


vixentael:
awesome catch!

PR #274, opened 2018-10-22
add tracing with opencensus

Lagovas:
* extend connection wrapper with new context.Context parameter
* use specific logger instead global when it exists
* send trace data from connector to server/translator
* receive trace data on translator for http requests (grpc now unsupported because need deeper integration)
* receive trace data on server side and use it to create next spans
* add log exporter of traces
* add parameters for jaeger exporter

### DIFF
--- a/cmd/acra-connector/acra-connector.go
+++ b/cmd/acra-connector/acra-connector.go
@@ -151,7 +151,7 @@ func handleConnection(config *Config, connection net.Conn) {
 			Errorln("Can't wrap connection")
 		span.SetStatus(trace.Status{Code: trace.StatusCodeUnknown})
 		if err := acraConn.Close(); err != nil {
-			logger.WithError(err).Errorln("Error on closing connection with AcraServer")
+			logger.WithError(err).Errorf("Error on closing connection with %v", connector_mode.ModeToServiceName(config.Mode))
 		}
 		wrapSpan.End()
 		return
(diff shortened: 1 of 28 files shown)


cmd/acra-connector/acra-connector.go
```diff
@@ -77,44 +80,56 @@ func handleApiConnection(config *Config, connection net.Conn) {
 }
 
 func handleConnection(config *Config, connection net.Conn) {
+	options := []trace.StartOption{trace.WithSpanKind(trace.SpanKindClient)}
```
vixentael: what is "span kind client"?

Lagovas: just metadata that indicate client/server side of span
https://opencensus.io/core-concepts/tracing/#spankind
it does not affect anything but show it in some trace collectors

cmd/acra-server/config.go
```diff
@@ -74,12 +77,32 @@ type UIEditableConfig struct {
...
 func NewConfig() *Config {
-	return &Config{withZone: false, stopOnPoison: false, wholeMatch: true, mysql: false, postgresql: false}
+	return &Config{withZone: false, stopOnPoison: false, wholeMatch: true, mysql: false, postgresql: false, withConnector: true}
 }
 
 // ErrTwoDBSetup shows that AcraServer can connects only to one database at the same time
 var ErrTwoDBSetup = errors.New("only one db supported at one time")
 
+// WithConnector show that acra-server will or not accept connections from acra-connector. Default: true
+func (config *Config) WithConnector() bool {
```
vixentael: .. AcraServer expects connections from AcraConnector

PR #275, opened 2018-10-25
refactor logs [T836]

Lagovas:
* use pre-generated data for tests. previously generating data took 0.x ms, now 0.000x-0.0000x ms, but actually it not decrease time of tests so much as expected :(
* log pid of process on start Acra[Server|Connector|Translator|WebConfig]. was need to test restart and decided that it may be useful in a future
* use our wrapper of net.Dial/Listenen from network package to add ability globally wrap all connections in project with our logic (added wrapper that ensure that conn.Close will called only once to avoid "i/o operation on closed connection" error)
* use debug log level for all errors that depends on network (closed connection, unavailable network, etc) that will catched and handled on high level function of processing connection (instead lower level with packet parsing/decryption/acra checks). Use error log level only in cases when sure that it unexpected behavior.
* stop accepting connections after error and stop AcraServer instead infinite loop with trying to accept connections even when listening socket was closed
* handle COM_QUIT(postgresql)/TerminatePacket(mysql) packets separately to correctly handle closing connections from client app
* increase timeout for sql statements in tests from 1s to 5s (some tests fail due to timeout to execute queries directly to db)

after these changes we should not to see error messages after simple actions like connect/disconnect with `psql`/`mysql` cli commands or `pg_isready` (in debug mode will see)


cmd/acra-server/client_session.go
```diff
@@ -120,20 +120,29 @@ func (clientSession *ClientSession) HandleClientConnection(clientID []byte, decr
 		go pgProxy.PgDecryptStream(clientSession.config.censor, decryptorImpl, clientSession.config.GetTLSConfig(), clientSession.connectionToDb, clientSession.connection, dbProxyErrorCh)
 	}
 	var channelToWait chan error
+	const (
+		acraDbSide     = "AcraServer<->Database"
```
vixentael: 👍

cmd/acra-server/client_session.go
```diff
@@ -120,20 +120,29 @@ func (clientSession *ClientSession) HandleClientConnection(clientID []byte, decr
...
+		clientAcraSide = "Client/Connector<->Database"
+	)
+	var interruptSide string
 	for {
 		select {
 		case err = <-dbProxyErrorCh:
-			clientSession.logger.WithError(err).Debugln("error from db proxy")
+			clientSession.logger.Debugln("Stop to proxy Database -> AcraServer")
+			interruptSide = acraDbSide
```
vixentael: 👍

decryptor/postgresql/packet_handler.go
```diff
@@ -300,12 +302,23 @@ func (packet *PacketHandler) ReadClientPacket() error {
 	*/
 	switch packetBuf[0] {
 	// all known message types with flags (F) or (F/B) on https://www.postgresql.org/docs/current/static/protocol-message-formats.html
-	case 'X', 'S', 'p', 'F', 'H', 'E', 'D', 'f', 'c', 'd', 'C', 'B', 'Q':
+	case 'S', 'p', 'F', 'H', 'E', 'D', 'f', 'c', 'd', 'C', 'B', 'Q':
```
vixentael: Why did you remove `X`?

vixentael: got it. `X` is terminate packet, right?

Lagovas: yes

PR #276, opened 2018-10-29
add jaeger auth params

Lagovas:
* add usename\password for basic auth for jaeger exporters. refactor jaeger params. register and fetch jaeger params with same method for all parts that use jaeger
* split tests. test mysql and postgresql separately. it decrease time of tests (from ~30 min to 18-20)
* some improvements in tests (explicitly set timeout for unix sockets)
* re-generate configs


.circleci/config.yml
```diff
@@ -56,8 +38,93 @@ jobs:
...
+    - run: if [ -f  $FILEPATH_ERROR_FLAG ]; then cat "$FILEPATH_ERROR_FLAG"; rm "$FILEPATH_ERROR_FLAG"; exit 1; fi
 workflows:
   version: 2
   tests:
     jobs:
-      - x86_64
\ No newline at end of file
+      - postgresql
+      - mysql
+      # has errors with tls connections
```
vixentael: :(

Lagovas: yeah. tried mariadb docker container as is and took ssl error with an incompatible version (old ssl version on mariadb side, maybe it is default setting in image or something else...). next time will spend more time to understand what is wrong and add to tests

PR #278, opened 2018-11-06
Update DB versions in docker demos

shadinua:
* PostgreSQL 9.6 -> 11
* MySQL 5.7 -> MariaDB 10.3


vixentael:
yey, MariaDB in da house!

examples/python/example_with_zone.py
```diff
@@ -141,14 +143,14 @@ def write_data(data, connection):
     metadata.create_all(engine)
 
     if args.print:
-        print_data(args.zone_id, connection)
+        print_data(args.zone_id, connection, driver)
```
Lagovas: what about place `print('DB driver: {}'.format(driver))` before `if args.print else ` statement and don't pass it to 2 functions with extra parameter?

examples/python/example_without_zone.py
```diff
@@ -109,6 +111,6 @@ def write_data(data, connection):
     connection = engine.connect()
 
     if args.print:
-        print_data(connection)
+        print_data(connection, driver)
```
Lagovas: same here

examples/python/example_with_zone.py
shadinua: Sure, thanks :-)

PR #279, opened 2018-11-06
drop tracing_enable param

Lagovas:
@shadinua correctly noticed that `--tracing_enable` looks like redundant and enough to leave `trace_log_enable` and `trace_jaeger_enable`. actually, `tracing_enable` doesn't provide any useful setting or configuration so we can drop it
plus did some refactoring of handling cli params related with tracing. moved repeated logic to cmd module and use it from places that depend on it


vixentael:
👍

PR #280, opened 2018-11-07
process queries in prepared statements in binary protocol (pgsql)

Lagovas:
noticed that now we don't process queries from prepared statements in binary protocol to postgresql with acra-censor (decryption worked before)
but can't add good integration tests now, because python hasn't good libraries to work with postgresql with binary protocol. psycopg2 (most popular) use text protocol for prepared statements (that we test too). found asyncpg, but tests with it unstable in my local environment (tests fail on closing connection or unexpected empty result without any errors)

now tested it manually before push and check that branch of fetching query from Parse packet works as expected and leave tests with asyncpg which can be used by commenting `self.skipTest` in `checkSkip` function

### DIFF
--- a/decryptor/mysql/response_proxy.go
+++ b/decryptor/mysql/response_proxy.go
@@ -322,10 +322,15 @@ func (handler *MysqlHandler) ClientToDbConnector(errCh chan<- error) {
 				}
 				continue
 			}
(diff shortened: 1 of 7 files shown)


vixentael:
Neat!
Could you please check prepared statements for MySQL as well?

Lagovas:
checked and fixed too

PR #283, opened 2018-11-12
test prepared statements with acra-censor

Lagovas:
* add usage of prepared statements in acra-censor tests and refactored connections to re-use the same code (connections with different drivers and text/binary prepared statements per db type) in several places 
* add query as is if can't normalize to QueryIgnore handler to allow ignore queries that can't parse

### DIFF
--- a/acra-censor/handlers/queryignore_handler.go
+++ b/acra-censor/handlers/queryignore_handler.go
@@ -54,7 +54,7 @@ func (handler *QueryIgnoreHandler) AddQueries(queries []string) {
 	for _, query := range queries {
 		normalizedQuery, _, err := NormalizeAndRedactSQLQuery(query)
 		if err != nil {
-			logrus.WithError(err).Warningln("Can't add query to QuieryIgnoreHandler in normalized form, added as is")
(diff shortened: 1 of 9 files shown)


acra-censor/handlers/queryignore_handler.go
```diff
@@ -52,6 +54,9 @@ func (handler *QueryIgnoreHandler) AddQueries(queries []string) {
 	for _, query := range queries {
 		normalizedQuery, _, err := NormalizeAndRedactSQLQuery(query)
 		if err != nil {
+			logrus.WithError(err).Warningln("Can't add query to QuieryIgnoreHandler in normalized form, added as is")
```
vixentael: Quiery -> Query

vixentael:
```suggestion
			logrus.WithError(err).Warningln("Can't add query to QueryIgnoreHandler in normalized form, added as is")
```

vixentael:
So, at the end:

AcraCensor can filter prepare statements only using QueryCapture/QueryIgnore (by comparing incoming SQL char-by-char with SQL from config file). AcraCensor can't apply allow/deny rules for SQL requests with prepared statements.

PR #285, opened 2018-11-21
transparent encryption on server side

Lagovas:
**IMPORTANT** not completed PR. I will add integration tests (now only unit-tests exists and I manually tested on local machine). But because it will huge part too, I decided to create PR before it for review less part of code and review general logic. I will push integration tests later before merging and one more review

features:
* encrypt data in `insert`/`update` queries with encryption settings from config file (which key use to encrypt per table+column)
* config file (https://github.com/cossacklabs/acra/compare/master...Lagovas:lagovas/transparent-encryption?expand=1#diff-ae17db40477a87c4508d60b7894e2b3a) where user should tell name of table where we must encrypt data, which columns encrypt and optionally all columns in table to support `insert` statements without list of columns before values (`insert into table (col1, col2, col3)`
* don't re-encrypt AcraStructs. if client's app already send data encrypted with acrawriter then proxy will no encrypt one more time (not supported for injected AcraStruct)


Lagovas:
in `sqlparser` was touched next files:
* ast.go
* dependency/sqltypes/value.go
* github_test.go
* parse_test.go
* sql.y
* token.go
Other files was leaved unchanged (only updated import paths)

cmd/acra-server/client_session.go
```diff
@@ -149,10 +170,13 @@ func (clientSession *ClientSession) HandleClientConnection(clientID []byte, decr
 				if clientSession.config.UseMySQL() {
 					break
 				} else {
-					pgProxy.TLSCh <- true
-					// in postgresql mode timeout used to stop listening connection in background goroutine
-					// and it's normal behaviour
-					continue
+					if pgProxy.TLSCh != nil {
```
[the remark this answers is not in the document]
Lagovas: avoid the second usage of this error after established TLS

configs/acra-encryptor.example.yaml
```diff
@@ -0,0 +1,22 @@
...
+    zone_id: DDDDDDDDMatNOMYjqVOuhACC
+
+- table: test2
+  columns:
+  - id
+  - zone
+  - data
+  - raw_data
+  encrypted:
+  - name: data
```
vixentael: column_name

vixentael: column

encryptor/queryDataEncryptor.go
```diff
@@ -0,0 +1,263 @@
...
+						return changed, err
+					} else if changedValue {
+						changed = true
+					}
+				}
+			}
+		}
+	}
+
+	if len(insert.OnDup) > 0 {
```
vixentael: onDup? duplicate key?

encryptor/queryDataEncryptor.go
```diff
@@ -0,0 +1,263 @@
...
+	return false
+}
+
+// encryptUpdateExpressions try to encrypt all supported exprs. Use firstTable if column has not explicit table name because it's implicitly used in DBMSs
+func (encryptor *QueryDataEncryptor) encryptUpdateExpressions(exprs sqlparser.UpdateExprs, firstTable sqlparser.TableName, qualifierMap qualifierToTableMap) (bool, error) {
+	var schema *TableSchema
+	changed := false
+	for _, expr := range exprs {
+		// recognize table name of column
+		if expr.Name.Qualifier.IsEmpty() {
```
vixentael: first table is default table?

encryptor/queryDataEncryptor.go
Lagovas: what to do if database recognize duplication. it's like update on duplicate or upsert.

decryptor/postgresql/packet_handler.go
```diff
@@ -227,11 +232,43 @@ func (packet *PacketHandler) IsParse() bool {
...
+	packet.logger.Debugln("GetParseQuery")
+	parse, err := NewParsePacket(packet.descriptionBuf.Bytes())
 	if err != nil {
+		packet.logger.Debugln("GetParseQuery error")
 		return "", err
 	}
-	return string(query[:len(query)-1]), nil
+	packet.logger.Debugln("GetParseQuery success")
+	return parse.QueryString(), nil
+	//query, err := FetchQueryFromParse(packet.descriptionBuf.Bytes())
```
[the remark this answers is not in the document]
Lagovas: yes. will remove

encryptor/queryDataEncryptor.go
Lagovas: yes. if multi-table update then if column without any table as prefix then work as it column from first table in the list

vixentael:
Looks awesome to me!

PR #291, opened 2018-11-28
Censor fixes

storojs72:
1) Move common table matching logic from whitelist/blacklist handlers;
2) Fix parsing error logic
3) Some refactoring

### DIFF
--- a/acra-censor/common/common.go
+++ b/acra-censor/common/common.go
@@ -14,12 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package handlers
+package common
 
 import (
+	"errors"
 	"github.com/cossacklabs/acra/logging"
 	log "github.com/sirupsen/logrus"
 	"github.com/xwb1989/sqlparser"
+	"github.com/xwb1989/sqlparser/dependency/querypb"
 	"strings"
 )
 
@@ -27,6 +29,30 @@ type pattern struct {
 	placeholder string
 	replacer    string
 }
+
+// Errors returned during parsing SQL queries.
+var (
+	ErrQueryNotInWhitelist             = errors.New("query not in whitelist")
+	ErrQueryInBlacklist                = errors.New("query in blacklist")
+	ErrAccessToForbiddenTableBlacklist = errors.New("query tries to access forbidden table")
+	ErrAccessToForbiddenTableWhitelist = errors.New("query tries to access forbidden table")
+	ErrBlacklistPatternMatch           = errors.New("query's structure is forbidden")
+	ErrWhitelistPatternMismatch        = errors.New("query's structure is forbidden")
+	ErrPatternSyntaxError              = errors.New("fail to parse specified pattern")
+	ErrPatternCheckError               = errors.New("failed to check specified pattern match")
+	ErrQuerySyntaxError                = errors.New("fail to parse specified query")
+	ErrComplexSerializationError       = errors.New("can't perform complex serialization of queries")
+	ErrCantOpenFileError               = errors.New("can't open file to write queries")
+	ErrCantReadQueriesFromFileError    = errors.New("can't read queries from file")
+	ErrUnexpectedTypeError             = errors.New("should never appear")
+)
+
+const (
+	// LogQueryLength is maximum query length for logging to syslog.
+	LogQueryLength = 100
+	// ValueMask is used to mask real Values from SQL queries before logging to syslog.
+	ValueMask = "replaced"
+)
 
 const (
 	// UnionPlaceholder is used when matching %%UNION%% pattern
@@ -149,4 +175,36 @@ func ParsePatterns(rawPatterns []string) ([]sqlparser.Statement, error) {
 		outputPatterns = append(outputPatterns, statement)
 	}
 	return outputPatterns, nil
+}
+
+// TrimStringToN trims query to N chars.
+func TrimStringToN(query string, n int) string {
+	if len(query) <= n {
+		return query
+	}
+	return query[:n]
+}
+
+// NormalizeAndRedactSQLQuery returns a normalized (lowercases SQL commands) SQL string,
+// and redacted SQL string with the params stripped out for display.
+// Taken from sqlparser package
+func NormalizeAndRedactSQLQuery(sql string) (normalizedQuery string, redactedQuery string, error error) {
+	bv := map[string]*querypb.BindVariable{}
+	sqlStripped, _ := sqlparser.SplitMarginComments(sql)
+
+	// sometimes queries might have ; at the end, that should be stripped
+	sqlStripped = strings.TrimSuffix(sqlStripped, ";")
+
+	stmt, err := sqlparser.Parse(sqlStripped)
+	if err != nil {
+		return "", "", err
+	}
+
+	normalizedQ := sqlparser.String(stmt)
+
+	// redact and mask VALUES
+	sqlparser.Normalize(stmt, bv, ValueMask)
+	redactedQ := sqlparser.String(stmt)
+
+	return normalizedQ, redactedQ, nil
 }
--- a/acra-censor/common/common.go
+++ b/acra-censor/common/common.go
@@ -185,10 +185,10 @@ func TrimStringToN(query string, n int) string {
 	return query[:n]
 }
 
-// NormalizeAndRedactSQLQuery returns a normalized (lowercases SQL commands) SQL string,
+// HandleRawSQLQuery returns a normalized (lowercases SQL commands) SQL string,
 // and redacted SQL string with the params stripped out for display.
 // Taken from sqlparser package
-func NormalizeAndRedactSQLQuery(sql string) (normalizedQuery string, redactedQuery string, error error) {
+func HandleRawSQLQuery(sql string) (normalizedQuery, redactedQuery string, parsedQuery sqlparser.Statement, err error) {
 	bv := map[string]*querypb.BindVariable{}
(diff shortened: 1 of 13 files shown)


acra-censor/common/common.go
```diff
@@ -150,3 +176,142 @@ func ParsePatterns(rawPatterns []string) ([]sqlparser.Statement, error) {
...
+// Taken from sqlparser package
+func HandleRawSQLQuery(sql string) (normalizedQuery, redactedQuery string, parsedQuery sqlparser.Statement, err error) {
+	bv := map[string]*querypb.BindVariable{}
+	sqlStripped, _ := sqlparser.SplitMarginComments(sql)
+
+	// sometimes queries might have ; at the end, that should be stripped
+	sqlStripped = strings.TrimSuffix(sqlStripped, ";")
+
+	stmt, err := sqlparser.Parse(sqlStripped)
+	outputStmt, _ := sqlparser.Parse(sqlStripped)
```
Lagovas: for what second parsing the same query?

acra-censor/common/common.go
```diff
@@ -150,3 +176,142 @@ func ParsePatterns(rawPatterns []string) ([]sqlparser.Statement, error) {
...
+		return false
+	}
+	return true
+}
+
+func CheckTableNamesMatch(parsedQuery sqlparser.Statement, setOfTables map[string]bool) (bool, bool) {
+	atLeastOneTableNameMatch := false
+	allTableNamesMatch := false
+
+	switch parsedQuery.(type) {
```
Lagovas: To avoid casts inside `case` statements we can cast once using switch statement and reduce lines of code below
```
switch query := parsedQuery.(type) {
  case *sqlparser.Select:
      // query here casted to *sqlparser.Select
      //query.From
}
```

acra-censor/common/common.go
```diff
@@ -150,3 +176,142 @@ func ParsePatterns(rawPatterns []string) ([]sqlparser.Statement, error) {
...
+	case *sqlparser.Select:
+		selectQuery := parsedQuery.(*sqlparser.Select)
+		atLeastOneTableNameMatch, allTableNamesMatch = checkTableExprsMatch(selectQuery.From, setOfTables)
+		break
+	case *sqlparser.Insert:
+		insertQuery := parsedQuery.(*sqlparser.Insert)
+		if setOfTables[insertQuery.Table.Name.String()] {
```
Lagovas: is it works for case `.... FROM table1 as tbl` ? Is `Table.Name.String()` will not return `tbl` instead `table1`?

acra-censor/common/common.go
```diff
@@ -150,3 +176,145 @@ func ParsePatterns(rawPatterns []string) ([]sqlparser.Statement, error) {
...
+	oneTableMatch := false
+	allTablesMatch := false
+	counter := 0
+	for _, tableExpr := range tables {
+		oneTableMatchInternal, allTablesMatchInternal := checkTableExprMatch(tableExpr, setOfTables)
+		if oneTableMatchInternal {
+			oneTableMatch = true
+			if allTablesMatchInternal {
+				counter++
+			}
```
Lagovas:
```
else { 
  break
}
```
we don't need to continue if we have one false for `alltablesMatchInternal`

acra-censor/common/common.go
```diff
@@ -150,3 +176,145 @@ func ParsePatterns(rawPatterns []string) ([]sqlparser.Statement, error) {
...
+		allTablesMatch = true
+	}
+	return oneTableMatch, allTablesMatch
+}
+
+func checkTableExprMatch(table sqlparser.TableExpr, setOfTables map[string]bool) (bool, bool) {
+	oneTableMatch := false
+	allTablesMatch := false
+
+	switch table.(type) {
```
Lagovas: I think we can simplify this function and don't add specific functions for each type of table. Each case will be < 10 lines of code and here we work with specific types. On my opinion better to check all cases here instead in separate functions (checkAliasedTable, checkJoinedTable, checkParenTable)
`switch tbl := table.(type) {`
 then we can reduce code and use recursion
```
case *sqlparser.ParenTableExpr:
		oneTableMatch, allTablesMatch = checkParenTable(tbl.Exprs, setOfTables)
```

acra-censor/handlers/blacklist_handler.go
```diff
@@ -42,150 +43,33 @@ func NewBlacklistHandler() *BlacklistHandler {
...
-				default:
-					return false, ErrUnexpectedTypeError
-				}
-			}
-		case *sqlparser.Insert:
-			if handler.tables[parsedQuery.Table.Name.String()] {
-				handler.logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorQueryIsNotAllowed).WithError(ErrAccessToForbiddenTableBlacklist).Errorln("Query has been blocked by blacklist [tables]")
-				return false, ErrAccessToForbiddenTableBlacklist
-			}
+		atLeastOneTableInBlacklist, _ := common.CheckTableNamesMatch(parsedQuery, handler.tables)
```
Lagovas: nice

acra-censor/common/common.go
storojs72: Because I need initial query (without redacted values) in parsed form outside

acra-censor/handlers/blacklist_handler.go
```diff
@@ -204,22 +88,22 @@ func (handler *BlacklistHandler) Release() {
...
 	for _, query := range queries {
-		normalizedQuery, _, err := NormalizeAndRedactSQLQuery(query)
-		if err != nil {
-			continue
+		handler.queries[query] = true
+		normalizedQuery, _, _, err := common.HandleRawSQLQuery(query)
+		if err == nil {
+			handler.queries[normalizedQuery] = true
 		}
-		handler.queries[normalizedQuery] = true
```
Lagovas: what about store query as is without normalization? to support unparsable queries

Lagovas: oh, I missed `handler.queries[query] = true`
I think we should store one variant: normalized or as is. We don't need to store both IMHO

acra-censor/common/common.go
storojs72: This is INSERT statement. Only one table is possible

acra-censor/common/common.go
Lagovas: then better to parse second time after error check because in worst case we will parse two times with errors without need

acra-censor/common/common.go
storojs72: To use it both whitelist and blacklist we should have information if AT LEAST ONE table is in list (blacklist blocks table if it's true, allows - otherwise) and if ALL tables are in list (whitelist allows table if it's true and blocks otherwise)

acra-censor/handlers/blacklist_handler.go
storojs72: I think that we should save both versions (normalized and raw). If parsing error occurred, we add/remove only raw version from list in all cases

acra-censor/common/common.go
Lagovas: yes, but if we have `oneTableMatch == true` and meet first `allTablesMatchInternal == false` then we can stop iteration and return `true, false` because all ``allTablesMatchInternal` will not true anyway

acra-censor/handlers/blacklist_handler.go
Lagovas: for what we need store both versions? how we use it? we can add/remove only normalized or raw query. anyway, we try to parse in both cases and check with logical OR both versions (in queryignore handler). so we can reduce overhead of memory and store one version

acra-censor/common/common.go
storojs72: Agree

acra-censor/common/common.go
storojs72: Agree

acra-censor/handlers/blacklist_handler.go
storojs72: In previous version we tried to parse and store normilized. If parse error occurred, we save as is. I agree that we have to save only normalized or raw. It seems that raw type is better - no parse, faster work

acra-censor/common/common.go
storojs72: Agree

acra-censor/handlers/blacklist_handler.go
storojs72: Eventually I decided to use normalized queries CheckQuery and normalize them in Add/Remove Queries functions. So "SELECT * FROM STUDENTS;" will match "select * from students"

Lagovas: if we can parse then we operate with parsed queries. and when we will try to add/remove parsable query then it will correctly added/removed. and when we cant parse then we add as is in raw form and add/remove will fail with parsing too and correctly added/removed using raw form of query

storojs72: I think we shouldn't allow add to black/white list unparsed queries - we check normalized queries in CheckQuery - not raw, and normalized query means that it's parsed.

storojs72: Currently every input query is parsed twice (one statement used for hiding values and second is processed) independently of further handling logic. When we parse while add/remove, it's done only once when loading. So even if we use unparsed queries in checking queries, It will not increase performance, but may cause some annoying typo errors, like add to whitelist "select * from x" but it will not block "select * from x;" because strings are not actually equal.

PR #292, opened 2018-12-11
encode data to bytes in tests/generate_acrastruct.py

Lagovas:
prevent decryption unexpected result with \0\0\0 values without correct encoding


vixentael:
👍 

makes sense to build & upload new package acrawriter python?

Lagovas:
no because here wasn't touched acrawriter, only helper for tests

vixentael:
ok, got it!

thank you!

PR #295, opened 2018-12-21
T864 - Separate log file for unparsed queries

storojs72:


acra-censor/acra-censor_implementation.go
```diff
@@ -20,23 +20,26 @@ import (
...
+	parseErrorsLogPath string
+	logger             *log.Entry
 }
 
 // NewAcraCensor creates new censor object.
 func NewAcraCensor() *AcraCensor {
 	acraCensor := &AcraCensor{}
 	acraCensor.logger = log.WithField("service", ServiceName)
 	acraCensor.ignoreParseError = false
+	acraCensor.parseErrorsLogPath = ""
```
Lagovas: we don't need such initialization because it's already initialized with [empty string](https://tour.golang.org/basics/12). Above assigning is redundant too

acra-censor/acra-censor_configuration_provider.go
```diff
@@ -50,6 +52,15 @@ func (acraCensor *AcraCensor) LoadConfiguration(configuration []byte) error {
 		return err
 	}
 	acraCensor.ignoreParseError = censorConfiguration.IgnoreParseError
+	acraCensor.parseErrorsLogPath = censorConfiguration.ParseErrorsLog
+	if acraCensor.parseErrorsLogPath != "" {
+		openedFile, err := os.OpenFile(acraCensor.parseErrorsLogPath, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0600)
+		if err != nil {
+			return err
+		}
+		defer openedFile.Close()
```
Lagovas: if we doesn't use this file after creation we can close it here without defer. but better to store file descriptor as censor field to avoid opening every write operation and reduce count of  syscalls (open/close)

acra-censor/acra-censor_implementation.go
```diff
@@ -72,6 +75,13 @@ func (acraCensor *AcraCensor) HandleQuery(rawQuery string) error {
 	normalizedQuery, queryWithHiddenValues, parsedQuery, err := common.HandleRawSQLQuery(rawQuery)
 	if err == common.ErrQuerySyntaxError {
 		acraCensor.logger.WithError(err).Warning("Failed to parse input query")
+		if acraCensor.parseErrorsLogPath != "" {
+			err := acraCensor.saveQuery(rawQuery)
+			if err != nil {
+				acraCensor.logger.WithError(err).Errorf("An error occurred while saving unparsable query")
```
Lagovas: use `Errorln` to append '\n' at end of log msg. `Errorf` useful to create msg with some format of string

acra-censor/acra-censor_implementation.go
```diff
@@ -72,6 +75,13 @@ func (acraCensor *AcraCensor) HandleQuery(rawQuery string) error {
...
+				return err
```
Lagovas: I think if censor configured with `acracensor.ignoreParseError` then we should not stop processing on error related with unparsed query and enough log error and try to save without stopping processing.

acra-censor/acra-censor_implementation.go
```diff
@@ -110,3 +120,19 @@ func (acraCensor *AcraCensor) HandleQuery(rawQuery string) error {
...
+func (acraCensor *AcraCensor) saveQuery(rawQuery string) error {
+	openedFile, err := os.OpenFile(acraCensor.parseErrorsLogPath, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0600)
+	if err != nil {
+		return err
+	}
+
+	defer openedFile.Close()
+
+	_, err = openedFile.WriteString(rawQuery + "\n")
+	if err != nil {
```
Lagovas: `return err` enough imho :)

plus better to use opened file on loading configuration and use it to avoid open/close on every save

configs/acra-censor.example.yaml
```diff
@@ -1,4 +1,5 @@
 ignore_parse_error: false
+parse_errors_log: unparsed_queries_log
```
Lagovas: what about `uparsed_queries.log` name? to add some common extension to avoid meaning some binary file without extension on linux

acra-censor/acra-censor_implementation.go
```diff
@@ -110,3 +120,19 @@ func (acraCensor *AcraCensor) HandleQuery(rawQuery string) error {
...
+
+func (acraCensor *AcraCensor) saveQuery(rawQuery string) error {
+	openedFile, err := os.OpenFile(acraCensor.parseErrorsLogPath, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0600)
+	if err != nil {
+		return err
+	}
+
+	defer openedFile.Close()
+
+	_, err = openedFile.WriteString(rawQuery + "\n")
```
Lagovas: here we will have same problem which we had with querycapture handler where we try to buffer data before dumping to file to avoid slow write on each query. here we will have same problem with situation when client will use not supported queries. plus we don't need repeated queries, only unique. maybe would be good to write some `QueryWriter` (and use in querycapture if we can to avoid duplicated code) which will do this logic with buffered write and timeouts and store unique queries in memory to avoid repeated queries

acra-censor/acra-censor_implementation.go
Lagovas: @vixentael what you think about this?

acra-censor/acra-censor_configuration_provider.go
storojs72: When and where should I close it is such case when acraserver stops?

Lagovas: just here) `openedFile.Close()` instead `defer openedFile.Close()` because this file doesn't used anywhere after creation. As I understand here we just create file and check that we can do this and close it before next write.
But if we will store opened file in censor then we don't need `.Close()` here even with `defer`

acra-censor/acra-censor_implementation.go
storojs72: Ok, I will write additional handler for this (with functionality like QueryCapture)

acra-censor/acra-censor_implementation.go
vixentael: agree, if `acracensor.ignoreParseError` is true, then log error and move forward

acra-censor/acra-censor_implementation.go
vixentael: agree, it makes sense to allocate same functionality into some object lke `QueryWriter` and use it for both `QueryCapture` and here

acra-censor/acra-censor_implementation.go
```diff
@@ -18,25 +18,30 @@ package acracensor
...
+	logger                *log.Entry
 }
 
 // NewAcraCensor creates new censor object.
 func NewAcraCensor() *AcraCensor {
 	acraCensor := &AcraCensor{}
 	acraCensor.logger = log.WithField("service", ServiceName)
 	acraCensor.ignoreParseError = false
+	acraCensor.parsedQueriesWriter = nil
+	acraCensor.unparsedQueriesWriter = nil
```
Lagovas: we can skip it because its already nil

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,285 @@
...
+	handler.signalBackgroundExit = signalBackgroundExit
+
+	handler.serializationTimeout = DefaultSerializationTimeout
+	handler.serializationTicker = time.NewTicker(DefaultSerializationTimeout)
+	handler.logger = log.WithField("internal_object", "querywriter")
+
+	// read existing queries from file
+	err = handler.readAllQueriesFromFile()
+	if err != nil {
+		handler.logger.WithError(ErrCantReadQueriesFromFileError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError).Errorln("Can't create QueryWriter instance")
```
Lagovas: I think we should log original error `err` using `WithError`

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,285 @@
...
+		openedFile.Close()
```
Lagovas: we should process error to by logging it

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	//handling goroutine
+	go func() {
+		for {
+			select {
+			case <-handler.serializationTicker.C:
+				err := handler.dumpBufferedQueriesToFile(openedFile)
+				if err != nil {
+					handler.logger.WithError(ErrComplexSerializationError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError).Errorln("Can't create QueryWriter instance")
```
Lagovas: log original error

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,285 @@
...
+				}
+				handler.serializationTicker.Stop()
+				handler.serializationTicker = time.NewTicker(handler.serializationTimeout)
+
+			case <-signalBackgroundExit:
+				err := handler.finishAndCloseFile(openedFile)
+				if err != nil {
+					handler.logger.WithError(ErrComplexSerializationError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError).Errorln("Can't create QueryWriter instance")
```
Lagovas: same

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,285 @@
...
+
+// GetSerializationTimeout gets timeout of dumping captured queries to the file.
+func (queryWriter *QueryWriter) GetSerializationTimeout() time.Duration {
+	return queryWriter.serializationTimeout
+}
+
+// DumpAllQueriesToFile writes stored queries into file.
+func (queryWriter *QueryWriter) DumpAllQueriesToFile() error {
+	// open or create file, NO APPEND
+	f, err := os.OpenFile(queryWriter.filePath, os.O_TRUNC|os.O_RDWR|os.O_CREATE, 0600)
```
Lagovas: better to close file in function which open it with `defer`

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,285 @@
...
+	err := queryWriter.appendQueries(queryWriter.BufferedQueries, openedFile)
+	if err != nil {
+		return err
+	}
+	// clean buffered queries only after successful write
+	queryWriter.BufferedQueries = nil
+	return nil
+}
+
+func (queryWriter *QueryWriter) appendQueries(queries []*QueryInfo, openedFile *os.File) error {
```
Lagovas: because we don't use any os.File specific API then we can take io.Writer as param.

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+		if len(jsonQueryInfo) > 0 {
+			jsonQueryInfo = append(jsonQueryInfo, '\n')
+			linesToAppend = append(linesToAppend, jsonQueryInfo...)
+		}
+	}
+	return linesToAppend
+}
+
+func (queryWriter *QueryWriter) finishAndCloseFile(openedFile *os.File) error {
```
Lagovas: https://golang.org/pkg/io/#Closer enough here. interfaces better to use instead specific types

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,285 @@
...
+		return nil, err
+	}
+
+	// signals
+	signalShutdown := make(chan os.Signal, 2)
+	signal.Notify(signalShutdown, os.Interrupt, syscall.SIGTERM)
+	signalBackgroundExit := make(chan bool)
+
+	// create handler
+	handler := &QueryWriter{}
```
Lagovas: I think we should assign `openedFile` to the related writer which use it to not forget to close it on Release/Close. resource owner responsible for releasing resources. plus it allow to avoid extra OpenFile calls because file already opened

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,285 @@
...
+			case <-handler.serializationTicker.C:
+				err := handler.dumpBufferedQueriesToFile(openedFile)
+				if err != nil {
+					handler.logger.WithError(ErrComplexSerializationError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError).Errorln("Can't create QueryWriter instance")
+				}
+				handler.serializationTicker.Stop()
+				handler.serializationTicker = time.NewTicker(handler.serializationTimeout)
+
+			case <-signalBackgroundExit:
+				err := handler.finishAndCloseFile(openedFile)
```
Lagovas: imho better to call `handler.Release` here and rename `finishAndCloseFile` to `Release`, which should dump data and close file

acra-censor/common/logging_logic.go
storojs72: We use `openedFile.Write(lines)` in this function

acra-censor/common/logging_logic.go
storojs72: Can you please, explain in more details

acra-censor/common/file_log_storage.go
```diff
@@ -0,0 +1,70 @@
...
+
+// ReadAll returns stored queries in raw form from internal file
+func (storage *FileLogStorage) ReadAll() ([]byte, error) {
+	return ioutil.ReadFile(storage.filePath)
+}
+
+// WriteAll writes raw queries that need to be stored to internal file
+func (storage *FileLogStorage) WriteAll(p []byte) error {
+	if storage.openedToAppend {
+		openedFile, err := os.OpenFile(storage.filePath, os.O_TRUNC|os.O_RDWR|os.O_CREATE, 0600)
```
Lagovas: we don't use already opened file in our constructor. we didn't close it and didn't reuse. I prefer to reuse:
```
if err := storage.file.Truncate(0); err != nil {return err} // drop all data but leave file already opened and ready to write
```

acra-censor/common/file_log_storage.go
```diff
@@ -0,0 +1,70 @@
...
+		if err != nil {
+			return err
+		}
+		storage.file = openedFile
+		storage.openedToAppend = false
```
Lagovas: for what we mark it? in function `WriteAll` we can truncate file and write to empty file. In `Append` function we use already opened file and just write to the end of file. So we can drop this field

acra-censor/common/file_log_storage.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Append appends raw queries to the end of internal file
+func (storage *FileLogStorage) Append(p []byte) error {
+	if !storage.openedToAppend {
+		openedFile, err := os.OpenFile(storage.filePath, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0600)
```
Lagovas: we can just into the file that opened before

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,268 @@
...
+func (queryWriter *QueryWriter) readStoredQueries() error {
+	queryWriter.mutex.Lock()
+	defer queryWriter.mutex.Unlock()
+
+	q, err := queryWriter.deserializeQueries()
+	if err != nil {
+		queryWriter.logger.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError).Errorln("Can't read stored queries")
+		return err
+	}
+	queryWriter.Queries = q
```
Lagovas: we need lock only here. deserialization doesn't use any shared resources but can take some time to read from slow FS and better to not include it into lock time.

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// GetSerializationTimeout gets timeout of dumping captured queries to the file.
+func (queryWriter *QueryWriter) GetSerializationTimeout() time.Duration {
+	return queryWriter.serializationTimeout
+}
+
+// DumpQueries writes all queries into file.
+func (queryWriter *QueryWriter) DumpQueries() error {
+	rawData := queryWriter.serializeQueries(queryWriter.Queries)
```
Lagovas: here we need lock to avoid background append to Queries

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,268 @@
...
+			linesToAppend = append(linesToAppend, jsonQueryInfo...)
+		}
+	}
+	return linesToAppend
+}
+
+// GetAllInputQueries returns a list of non-masked RawQueries.
+func (queryWriter *QueryWriter) GetAllInputQueries() []string {
+	var queries []string
+	for _, queryInfo := range queryWriter.Queries {
```
Lagovas: here we need Rlock. we read shareable resource that may be changed during read. and go runtime will panic on loop through map that was changed at this time.

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,268 @@
...
+		return
+	}
+	queryWriter.MarkQueryAsForbidden(queryWithHiddenValues)
+}
+
+// MarkQueryAsForbidden marks particular query as forbidden.
+// It will be written to file on Stop, Reset or Release.
+// Expects redacted query
+func (queryWriter *QueryWriter) MarkQueryAsForbidden(queryWithHiddenValues string) {
+	for index, queryInfo := range queryWriter.Queries {
```
Lagovas: here too RLock

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,268 @@
...
+	//skip already captured queries
+	for _, queryInfo := range queryWriter.Queries {
+		if strings.EqualFold(queryInfo.RawQuery, queryWithHiddenValues) {
+			return true, nil
+		}
+	}
+	queryInfo := &QueryInfo{}
+	queryInfo.RawQuery = queryWithHiddenValues
+	queryInfo.IsForbidden = false
+	queryWriter.Queries = append(queryWriter.Queries, queryInfo)
```
Lagovas: Lock on write

acra-censor/acra-censor_implementation.go
```diff
@@ -110,3 +103,39 @@ func (acraCensor *AcraCensor) HandleQuery(rawQuery string) error {
...
+
+// SetLoggingTimeout sets timeout of censor's logging process
+func (acraCensor *AcraCensor) SetLoggingTimeout(duration time.Duration) {
+	acraCensor.parsedQueriesWriter.SetSerializationTimeout(duration)
+	acraCensor.unparsedQueriesWriter.SetSerializationTimeout(duration)
+}
+
+func (acraCensor *AcraCensor) saveUnparsedQuery(query string) {
+	if acraCensor.unparsedQueriesWriter != nil {
+		saveQuery(acraCensor.unparsedQueriesWriter, query)
```
Lagovas: imho better to use `unparsedQueriesWriter.WriteQuery(query)` or move `saveQuery` to file with this handler

acra-censor/common/logging_logic.go
storojs72: Can you please explain what is Rlock? I googled but only usual Mutex.lock found

acra-censor/common/logging_logic.go
```diff
@@ -69,7 +69,7 @@ func NewFileQueryWriter(filePath string) (*QueryWriter, error) {
 	// read existing queries
 	err = writer.readStoredQueries()
 	if err != nil {
-		writer.logger.WithError(ErrCantReadQueriesFromFileError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError).Errorln("Can't create QueryWriter instance")
+		writer.logger.WithError(ErrCantReadQueriesFromStorageError).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError).Errorln("Can't create QueryWriter instance")
```
Lagovas: we should log origin error too

acra-censor/common/logging_logic.go
Lagovas: after serialization, we can unlock and avoid lock on writing to file because we have serialized copy of data

acra-censor/common/logging_logic.go
```diff
@@ -134,13 +127,13 @@ func (queryWriter *QueryWriter) Release() {
 }
 
 func (queryWriter *QueryWriter) reset() {
+	queryWriter.mutex.Lock()
+	defer queryWriter.mutex.Unlock()
```
Lagovas: `defer` good to use if we have a lot of points of exit to avoid multiple calls of `Unlock`. For example
```
func t(){
   l.Lock()
   _, err := func1()
   if err != nil{
      l.Unlock()
      return
   }
   _, err := func2()
  if err != nil{
      l.Unlock()
      return
   }
   _, err := func3()
  if err != nil{
      l.Unlock()
      return
   }
}
```
In such case to not forget unlock and call it once we should use `defer l.Unlock` and it simple
But `defer` has a small overhead. It's really small but it have. And when we have simple cases such this we can avoid it without any care that we can forget

acra-censor/common/file_log_storage.go
```diff
@@ -20,25 +19,18 @@ func NewFileLogStorage(filePath string) (*FileLogStorage, error) {
...
 // WriteAll writes raw queries that need to be stored to internal file
 func (storage *FileLogStorage) WriteAll(p []byte) error {
-	if storage.openedToAppend {
-		openedFile, err := os.OpenFile(storage.filePath, os.O_TRUNC|os.O_RDWR|os.O_CREATE, 0600)
-		if err != nil {
-			return err
-		}
-		storage.file = openedFile
-		storage.openedToAppend = false
+	if err := storage.file.Truncate(0); err != nil {
```
Lagovas: looks better)

acra-censor/common/file_log_storage.go
```diff
@@ -7,7 +7,6 @@ import (
 
 // FileLogStorage is a file-based implementation of LogStorage interface
 type FileLogStorage struct {
-	filePath       string
 	file           *os.File
 	openedToAppend bool
```
Lagovas: do we need it more?

acra-censor/common/logging_logic.go
storojs72: It will be logged from readStoredQueries function

acra-censor/common/file_log_storage.go
storojs72: Sure. I just didn't know that Truncate method exists

acra-censor/common/file_log_storage.go
storojs72: no it seems we didn't. I used it when reopening file with other modes set, but now Truncate allows to avoid it

acra-censor/common/logging_logic.go
```diff
@@ -0,0 +1,271 @@
...
+	if err := queryWriter.logStorage.WriteAll(rawData); err != nil {
+		queryWriter.logger.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError).Errorln("Can't dump queries to storage")
+		return err
+	}
+	return nil
+}
+
+// Release dumps all Captured queries to file and resets Captured queries list.
+func (queryWriter *QueryWriter) Release() {
+	queryWriter.DumpQueries()
```
Lagovas: we should handle error and log about it

acra-censor/common/logging_logic_test.go
```diff
@@ -0,0 +1,251 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	for _, query := range testQueries {
+		_, err = writer.RedactAndCheckQuery(query)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+	time.Sleep(DefaultSerializationTimeout + 100*time.Millisecond)
```
Lagovas: in previous PR's we decreased time of tests twice just by decreasing granulation of timeouts. and there we can't it handle more deterministic because we can't control environment fully
but here in unit-tests we can. what about to move background goroutine in our constructor `NewFileQueryWriter` to some function like `start()` and rewrite like something:
```
func (writer *QueryWriter) start(){
  for {
			select {
			case <-writer.serializationTicker.C:
				err := writer.dumpBufferedQueries()
				if err != nil {
					writer.logger.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError).Errorln("Can't dump buffered queries")
				}
				writer.serializationTicker.Stop()
				writer.serializationTicker = time.NewTicker(writer.serializationTimeout)
			case <-signalBackgroundExit:
				writer.serializationTicker.Stop()
				err := writer.logStorage.Close()
				if err != nil {
					writer.logger.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError).Errorln("Error occurred on exit QueryWriter instance")
				}
			case <-signalShutdown:
				writer.serializationTicker.Stop()
				err := writer.logStorage.Close()
				if err != nil {
					writer.logger.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorIOError).Errorln("Error occurred on shutdown QueryWriter instance")
				}
				return
			default:
				//do nothing. This means that channel has no data to read yet
			}
		}
}

// NewFileQueryWriter creates QueryWriter instance
func NewFileQueryWriter(filePath string) (*QueryWriter, error) {
  writer := &QueryWriter{}
// initialization logic
  go writer.Start()
  return writer, nil
}
```

after that we can in our tests do:
```
writer, err := NewFileQueryWriter(tmpFile.Name())
for _, query := range testQueries {
		_, err = writer.RedactAndCheckQuery(query)
		if err != nil {
			t.Fatal(err)
		}
	}
writer.Release() // or writer.DumpQueries() or any other private method
go writer.start()  // here we restart our background goroutine 
```
plus we need this background goroutine only to test logic with background serialization and stop. in other tests (dumping/serialization/deserialization) we can test without it, just call direct methods with `writer.dumpBlaBla`, `writer.deserializeBlaBla`

acra-censor/common/logging_logic.go
storojs72: DumpQueries logs errors inside

acra-censor/common/logging_logic.go
```diff
@@ -52,26 +52,24 @@ type QueryWriter struct {
 // NewFileQueryWriter creates QueryWriter instance
 func NewFileQueryWriter(filePath string) (*QueryWriter, error) {
 	// create writer
-	writer := &QueryWriter{}
+	writer := &QueryWriter{
+		queryIndex:0,
```
Lagovas: i'm not sure, but 90% that this file not formatted with gofmt )

acra-censor/handlers/querycapture_handler.go
```diff
@@ -59,14 +59,14 @@ func (handler *QueryCaptureHandler) MarkQueryAsForbidden(query string) error {
...
-			newQueryInfo.RawQuery = queryWithHiddenValues
-			newQueryInfo.IsForbidden = true
-			handler.writer.SetQuery(newQueryInfo, index)
+	err = handler.writer.WalkQueries(func(queryInfo *common.QueryInfo) error {
+		if strings.EqualFold(queryInfo.RawQuery, queryWithHiddenValues) {
+			queryInfo.IsForbidden = true
 		}
+		return nil
+	})
+	if err != nil {
```
Lagovas: `return err`  : )

storojs72: It's output value of internal function that passed as argument to WalkQueries

PR #296, opened 2018-12-27
support pgsql type cast [T883]

Lagovas:
* support `RETURNING` clause for `insert`'s which used in postgresql and our django examples
* add support of quoted identifiers in sql to allow use `"tablename"` and `where "column"=1` in postgresql/mysql which uses by orm (ex. django orm)
p.s. this pr fix usage transparent encryption mode for django example (@shadinua )


storojs72:
Can you please explain why and how do you need to use sqlparser somewhere except acra-censor?

Lagovas:
sqlparser we need to parse insert/update queries and replace values with encrypted. 1. match table which touched with query and match columns which marked in config to encrypt

PR #297, opened 2018-12-29
T863 acracensor

storojs72:
1) Rename Whitelist/Blacklist into Allow/Deny
2) Add AllowAll/DenyAll handlers
3) Allow handler now doesn't block any query. If query is whitelisted, it passes and further verifications do not apply, else - othervwise
4) Handlers can be added in arbitrary order in configuration file

@shadinua look please on new configuration file organization


acra-censor/acra-censor_implementation.go
```diff
@@ -75,21 +75,34 @@ func (acraCensor *AcraCensor) HandleQuery(rawQuery string) error {
...
+				return nil
+			} else {
+				continue
+			}
 		}
+		// Security checks (allow/deny handlers)
 		continueHandling, err := handler.CheckQuery(normalizedQuery, parsedQuery)
 		if err != nil {
-			acraCensor.logger.Errorf("Forbidden query: '%s'", queryWithHiddenValues)
+			acraCensor.logger.Errorf("Denied query: '%s'", queryWithHiddenValues)
```
Lagovas: would be good to add debug log with handlers type like `acracensor.logger.Debugf("Denied query by handler %T", handler)` which will log handler's type

acra-censor/handlers/allowall_handler.go
```diff
@@ -0,0 +1,29 @@
+package handlers
+
+import (
+	"github.com/cossacklabs/acra/sqlparser"
+	log "github.com/sirupsen/logrus"
+)
+
+// AllowallHandler allows any input query
+type AllowallHandler struct {
```
Lagovas: what about to rename to `AllowAllHandler`? upper 'all'

acra-censor/acra-censor_configuration_provider.go
```diff
@@ -112,7 +107,7 @@ func (acraCensor *AcraCensor) LoadConfiguration(configuration []byte) error {
 			go queryCaptureHandler.Start()
 			acraCensor.AddHandler(queryCaptureHandler)
 		default:
-			break
+			acraCensor.logger.Errorln("Unexpected handler in configuration")
```
Lagovas: would be good to return some error, not just log. with error, we disallow to run server with misconfigured censor

PR #299, opened 2019-01-09
optimizations/refactoring

Lagovas:
* added interface for Proxy that process mysql/postgresql connections and unified signatures for mysql/postgresql proxy methods to be compatible with one interface
* added interface for decryptors and unified signatures for db specific decryptors
* added decryptor/proxy factories that initialized at start of program and then used after accepting new connections. old code did initialization of components in different steps of processing connections (it was from our first version of server and need refactoring)
* initialize all db specific settings and components in main function once
* dropped redundant and unused fields/methods/constants in project (found them when tested golangci-lint : )
* refactored and encapsulated switching to tls in postgresql proxy (it was need to unifying usage of abstract proxy component without postgresql specific logic)
* optimized OnQueryObserver logic to re-use parsed queries between several observers (to avoid several parsing query, reuse ast tree and serialize to string only at end)
* add using absolute paths in integration tests to allow re-use in another projects/repos
* make some methods in our default keystore public to use in enterprise version

As I remember, I didn't change any public CLI or API interfaces that should be changed in documentation. And didn't change any behavior of server. Touched only some log messages (uppercase or change level from DEBUG to INFO)


decryptor/mysql/packet.go
```diff
@@ -128,7 +128,7 @@ func (packet *MysqlPacket) replaceQuery(newQuery string) {
 
 // readPacket read header to struct and return payload as return result or error
 func (packet *MysqlPacket) readPacket(connection net.Conn) ([]byte, error) {
-	if _, err := connection.Read(packet.header); err != nil {
+	if _, err := io.ReadFull(connection, packet.header); err != nil {
```
vixentael: what is the difference?

decryptor/postgresql/proxy.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &proxyFactory{
+		dataEncryptor: dataEncryptor,
+		setting:       proxySetting,
+	}, nil
+}
+
+// New return mysql proxy implementation
```
vixentael:
```suggestion
// New return postgresql proxy implementation
```

keystore/filesystem/filenames.go
```diff
@@ -18,11 +18,8 @@ package filesystem
 
 import (
 	"fmt"
-	"sync"
 )
 
-var lock = sync.RWMutex{}
```
vixentael: bye-bye mutex 😢

decryptor/mysql/packet.go
Lagovas: helper read more than once (if first read call was not full and return only part of data cached in os buffer). until some error or until no data was returned. ReadFull read in loop until read expected num of bytes. I meet one failed test when here read return part of packet and without error

keystore/filesystem/filenames.go
Lagovas: it wasn't used)

PR #300, opened 2019-01-09
use newer version of pyyaml due to CVE-2017-18342

Lagovas:
pyyaml uploaded to PYPI only 3.13 version so changed dependency to use directly sources from github instead PYPI until they update their package in PYPI
anyway we use this library only for tests and parse our static file from our repo and this fix just to hide alert from github : )


vixentael:
That's cool that github sends these vulnerability alerts and we fix them, even if we use this lib for tests only.

Great job!

PR #302, opened 2019-01-15
public methods to extract table aliases from query

Lagovas:
* make public `GetTablesWithAliases` function (was method) and `AliasedTableName` to reuse it in enterprise code 
* fix govet, use pointer to struct with sync.Once - https://goreportcard.com/report/github.com/cossacklabs/acra
* return decrypted string as is without encoding to hex if it simple string. it reduces length of packet

### DIFF
--- a/utils/dbByteArrayEncoders.go
+++ b/utils/dbByteArrayEncoders.go
@@ -89,8 +89,8 @@ func DecodeOctal(data []byte) ([]byte, error) {
 
 // DecodeEscaped with hex or octal encodings
 func DecodeEscaped(data []byte) ([]byte, error) {
-	hexdata := data[2:]
-	if bytes.Equal(data[:2], []byte{'\\', 'x'}) {
+	if len(data) > 2 && bytes.Equal(data[:2], []byte{'\\', 'x'}) {
+		hexdata := data[2:]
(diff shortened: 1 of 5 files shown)


utils/dbByteArrayEncoders.go
```diff
@@ -89,8 +89,8 @@ func DecodeOctal(data []byte) ([]byte, error) {
 
 // DecodeEscaped with hex or octal encodings
 func DecodeEscaped(data []byte) ([]byte, error) {
-	hexdata := data[2:]
-	if bytes.Equal(data[:2], []byte{'\\', 'x'}) {
+	if len(data) > 2 && bytes.Equal(data[:2], []byte{'\\', 'x'}) {
```
vixentael: what is data len < 2?

Lagovas: there check of escape type. hex encoding or not. second case printable chars used as is and string may contain <2 characters that are printable and will be correct

Lagovas: hex encoded must have prefix \x before encoded string so here we check it

vixentael: ok, got it, thankx

PR #304, opened 2019-01-18
fix occasionally commented censor's query capture test

storojs72:
While debugging censor's logging logic I occasionally commented this test. Now its time to restore it. Sorry, my fault

### DIFF
--- a/acra-censor/common/logging_logic_test.go
+++ b/acra-censor/common/logging_logic_test.go
@@ -195,79 +195,58 @@ func TestQueryCapture(t *testing.T) {
 		"SELECT * FROM X;",
 		"SELECT * FROM Y;",
 	}
-	_ = testQueries
-	/*
-		for _, query := range testQueries {
-			_, err = writer.RedactAndCheckQuery(query)
-			if err != nil {
-				t.Fatal(err)
-			}
-		}
-		expected := "{\"raw_query\":\"SELECT Student_ID FROM STUDENT\",\"_blacklisted_by_web_config\":false}\n" +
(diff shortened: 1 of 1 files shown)


storojs72:
I've updated it according to last changes in logging logic. Now it's user responsibility to redact query before writing to log, so we should only check if all queries have been written successfully

acra-censor/common/logging_logic_test.go
```diff
@@ -195,79 +195,58 @@ func TestQueryCapture(t *testing.T) {
...
-			"{\"raw_query\":\"SELECT * FROM Y\",\"_blacklisted_by_web_config\":false}\n" +
-			"{\"raw_query\":\"SELECT * FROM Z\",\"_blacklisted_by_web_config\":false}\n" +
-			"{\"raw_query\":\"select songName from t where personName in"
+	for _, query := range testQueries {
+		writer.WriteQuery(query)
+	}
+	expected := "{\"raw_query\":\"SELECT Student_ID FROM STUDENT;\",\"_blacklisted_by_web_config\":false}\n" +
+		"{\"raw_query\":\"SELECT * FROM STUDENT;\",\"_blacklisted_by_web_config\":false}\n" +
+		"{\"raw_query\":\"SELECT * FROM X;\",\"_blacklisted_by_web_config\":false}\n" +
+		"{\"raw_query\":\"SELECT * FROM Y;\",\"_blacklisted_by_web_config\":false}\n"
```
Lagovas: we can drop it because it repeat previous or change "Y" -> "X" to test duplicated

PR #305, opened 2019-01-24
return data as is on incorrect AcraStruct begin tag

Lagovas:


vixentael:
This is a hotfix for AcraServer communication error with some ORMs, like xorm.

It was found that in some cases, when AcraServer got plaintext data in response, it can't decrypt it (obviously) and can't find poison record (obviously), and throws decryption error to the top (instead of skipping it).

This fix will be released in `0.84.1` for current users.
Together with proper test cases, it will be part of future release `0.85.0` as well.

PR #309, opened 2019-02-01
correctly turn off encryptor

Lagovas:
don't initialize encryptor if config is empty and do not add it to OnQuery observers. it helps to avoid extra parsing of incoming queries

### DIFF
--- a/decryptor/mysql/proxy_test.go
+++ b/decryptor/mysql/proxy_test.go
@@ -13,7 +13,7 @@ func (*decryptorFactory) New(clientID []byte) (base.Decryptor, error) {
 	return nil, nil
 }
 
-type tableSchemaStore struct{empty bool}
+type tableSchemaStore struct{ empty bool }
 
 func (*tableSchemaStore) GetTableSchema(tableName string) *config.TableSchema {
 	panic("implement me")
@@ -52,4 +52,3 @@ func TestEncryptorTurnOnOff(t *testing.T) {
 		t.Fatal("Unexpected observers count")
 	}
 }
-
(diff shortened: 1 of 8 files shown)


vixentael:
Can we test it somehow?

Lagovas:
I will add tests soon to this pr

decryptor/mysql/proxy_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+func TestEncryptorTurnOnOff(t *testing.T) {
+	emptyStore := &tableSchemaStore{true}
+	nonEmptyStore := &tableSchemaStore{false}
+	setting := base.NewProxySetting(&decryptorFactory{}, emptyStore, nil, nil, nil)
+	proxyFactory, err := NewProxyFactory(setting)
+	if err != nil {
+		t.Fatal(setting)
+	}
+	proxy, err := proxyFactory.New(context.TODO(), nil, nil, nil)
```
storojs72: Why did you create new proxy factory with `context.TODO()` in mysql and with nil in postgres?

storojs72: It turns out that you need sometimes to pass some context when you create proxy in mysql case, but not in postgres right?

Lagovas: yes. mysql proxy implementation initialize logger in constructor - https://github.com/cossacklabs/acra/blob/master/decryptor/mysql/response_proxy.go#L175
but postgresql just save context (https://github.com/cossacklabs/acra/blob/master/decryptor/postgresql/pg_decryptor.go#L109) and create logger after (https://github.com/cossacklabs/acra/blob/master/decryptor/postgresql/pg_decryptor.go#L127, https://github.com/cossacklabs/acra/blob/master/decryptor/postgresql/pg_decryptor.go#L481)

PR #312, opened 2019-02-12
refactoring of log messages

Lagovas:
* added error codes for all warning/error messages
* synchronized some messages for mysql/postgresql
* removed old/redundant messages
* added killing processes in integrations tests via pidof to catch situations with sighup/restart on acra-server
* change blocking method in main function for acra-server, swapped SIGHUP with SIGTERM handler. actually it doesn't change any logic but tried it to fix unknown sighup messages on circleci. thought that maybe it is some golang runtime special behavior with closing channel at end or something like that...
* pass logger's with context to inner functions to log debug messages with related client/zone id
* avoid error logs with decryption failures when it expected and are allowed (binary data but not acrastruct) and use warnings instead errors when it failure on acrastruct (to notify clients because may be situation when clients think that all data correctly encrypted but it's not)


acra-censor/acra-censor_configuration_provider.go
```diff
@@ -107,7 +108,7 @@ func (acraCensor *AcraCensor) LoadConfiguration(configuration []byte) error {
 			go queryCaptureHandler.Start()
 			acraCensor.AddHandler(queryCaptureHandler)
 		default:
-			acraCensor.logger.Errorln("Unexpected handler in configuration")
+			acraCensor.logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSetupError).Errorln("Unexpected handler in configuration")
```
vixentael:
```suggestion
			acraCensor.logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSetupError).Errorln("Unexpected handler in configuration: probably AcraCensor configuration (acra-censor.yaml) is outdated.")
```

acra-censor/handlers/denyall_handler.go
```diff
@@ -20,7 +21,7 @@ func NewDenyallHandler() *DenyAllHandler {
 // CheckQuery blocks any input query
 func (handler *DenyAllHandler) CheckQuery(sqlQuery string, parsedQuery sqlparser.Statement) (bool, error) {
 	// deny any query and stop further checks
-	handler.logger.Errorf("Query has been block by Denyall handler")
+	handler.logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorQueryIsNotAllowed).Errorf("Query has been block by Denyall handler")
```
vixentael: Sounds a bit weird, but we started using "deny" everywhere, so..

```suggestion
	handler.logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorQueryIsNotAllowed).Errorf("Query has been denied by Denyall handler")
```

acra-censor/acra-censor_configuration_provider.go
ilammy: I find long lines like this a bit hard to read. Something like this would be clearer, I think.
```suggestion
			acraCensor.logger.
				WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSetupError).
				Errorln("Unexpected handler in configuration")
```

Though I don't know if one can convince **gofmt** that this code style is okay.

cmd/acra-server/acra-server.go
```diff
@@ -452,5 +452,5 @@ func main() {
 
 	// on sighup we run callback that stop all listeners (that stop background goroutine of server.Start())
```
ilammy: It seems that this comment line now belongs to line 426 where we install SIGHUP handler.

decryptor/postgresql/pg_decryptor.go
```diff
@@ -167,6 +171,7 @@ func (proxy *PgProxy) ProxyClientConnection(errCh chan<- error) {
 				proxy.TLSCh <- true
 				return
 			}
+			// log message with debug level because onlt here we expect and can meet errors with closed connections .ioEOF
```
ilammy: Typo
```suggestion
			// log message with debug level because only here we expect and can meet errors with closed connections .ioEOF
```

keystore/filesystem/server_keystore.go
```diff
@@ -75,8 +75,9 @@ func newFilesystemKeyStore(privateKeyFolder, publicKeyFolder string, encryptor k
 		return nil, err
 	}
 	fi, err := os.Stat(directory)
-	if nil == err && runtime.GOOS == "linux" && fi.Mode().Perm().String() != "-rwx------" {
-		log.Errorln("Key store folder has an incorrect permissions")
+	const expectedPermission = "-rwx------"
+	if nil == err && runtime.GOOS == "linux" && fi.Mode().Perm().String() != expectedPermission {
+		log.Errorf("Key store folder has an incorrect permissions, expected: %s", expectedPermission)
```
ilammy: It may be useful to log actual permissions as well. Like we do in utils/utils.go for files.

cmd/acra-server/listener.go
```diff
@@ -155,19 +155,21 @@ func (server *SServer) processConnection(connection net.Conn, callback *callback
...
 	}
+	logger = logger.WithField("client_id", string(clientID))
 	wrapSpan.End()
 	var span *trace.Span
 	if server.config.WithConnector() {
+		logger.Debugln("Read trace")
 		spanContext, err := network.ReadTrace(wrappedConnection)
 		if err != nil {
-			log.WithError(err).Errorln("Can't read trace from Acra-Proxy")
+			log.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorTracingCantReadTrace).WithError(err).Errorln("Can't read trace from Acra-Proxy")
```
vixentael:
```suggestion
			log.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorTracingCantReadTrace).WithError(err).Errorln("Can't read trace from AcraConnector")
```

logging/logging.go
```diff
@@ -43,6 +43,11 @@ type LoggerSetter interface {
 
 type loggerKey struct{}
 
+// IsDebugLevel return true if logger configured to log debug messages
+func IsDebugLevel(logger *log.Entry) bool {
```
ilammy: An idea of (possibly) more convenient interface for debug logging:

```go
some.logger.OnlyForLevel(log.DebugLevel).
	WithField("machine_readable", data).
	Debugf("human-readable message")
```

No idea how to implement this tho (at the moment), but with this we should avoid one unnecessary nesting level of the `if` check before logging.

Though, I believe the current approach has its merits as well. We may use `IsDebugLevel` to avoid doing some heavy computations for debug logging if it is not actually enabled.

cmd/prometheus.go
```diff
@@ -37,7 +38,7 @@ func RunPrometheusHTTPHandler(connectionString string) (net.Listener, *http.Serv
 		logrus.WithField("connection_string", connectionString).Infoln("Start prometheus http handler")
 		err := server.Serve(listener)
 		if err != nil {
-			logrus.WithError(err).Errorln("Error from http server that process prometheus metrics")
+			logrus.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorPrometheusHTTPHandler).WithError(err).Errorln("Error from http server that process prometheus metrics")
```
vixentael: I saw that you prefer HTTP now :)

```suggestion
			logrus.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorPrometheusHTTPHandler).WithError(err).Errorln("Error from HTTP server that process prometheus metrics")
```

tests/utils.py
```diff
@@ -15,6 +15,18 @@
...
 
+def send_signal_by_process_name(name, signal, timeout=1):
+    separator=' '
+    try:
+        output = subprocess.check_output(['pidof', '-S', separator, name], timeout=timeout)
+    except subprocess.CalledProcessError:
+        return
+    output = output.strip().decode('utf-8').split(separator)
+    for pid in output:
+        os.kill(int(pid), signal)
```
ilammy: If we're using subprocess anyway then it may be more simple to call **killall** utility which sends a signal to processes by their name pattern. It's a pretty old BSD utility which should be generally available everywhere.

decryptor/postgresql/pg_decryptor.go
```diff
@@ -436,14 +443,12 @@ func (proxy *PgProxy) processInlineBlockDecryption(ctx context.Context, packet *
 		if err != nil {
 			span.AddAttributes(trace.BoolAttribute("failed_decryption", true))
 			base.AcrastructDecryptionCounter.WithLabelValues(base.DecryptionTypeFail).Inc()
-			logger.WithError(err).Warningln("Can't unwrap symmetric key")
+			logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorDecryptorCantDecryptBinary).WithError(err).Warningln("Can't decrypt AcraStruct")
```
vixentael: Imagine we get these logs from customer. How could we understand where exactly decryption failed.

```suggestion
			logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorDecryptorCantDecryptBinary).WithError(err).Warningln("Can't decrypt AcraStruct: can't unwrap symmetric key")
```

decryptor/postgresql/pg_decryptor.go
```diff
@@ -457,12 +462,13 @@ func (proxy *PgProxy) processInlineBlockDecryption(ctx context.Context, packet *
 		if err != nil {
 			span.AddAttributes(trace.BoolAttribute("failed_decryption", true))
 			base.AcrastructDecryptionCounter.WithLabelValues(base.DecryptionTypeFail).Inc()
-			logger.WithError(err).Warningln("Can't decrypt data with unwrapped symmetric key")
+			logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorDecryptorCantDecryptBinary).WithError(err).Warningln("Can't decrypt data with unwrapped symmetric key")
```
vixentael: Do you think we could improve this error message too?

```suggestion
			logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorDecryptorCantDecryptBinary).WithError(err).Warningln("Can't decrypt AcraStruct: can't decrypt data with unwrapped symmetric key")
```

utils/logging.go
```diff
@@ -1,26 +0,0 @@
...
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
-
-package utils
-
-import "fmt"
-
-// TODO: move to /logging and refactor everything
```
vixentael: 👍

decryptor/postgresql/pg_decryptor.go
Lagovas: im not agree. this info will be available in debug mode. but I removed logging as warnings such deep information where it was failed. now it's general warning that can't decrypt probably matched AcraStruct. and in common cases it should not be often logged

vixentael: ok, makes sense

tests/utils.py
Lagovas: yes but killall is not installed on all systems by default. but pidof is posix utility that available in more OS's. it more useful for local running tests without requirement to install killall on local/docker machine. 
p.s. debian which used in our tests hasn't installed killall by default

logging/logging.go
Lagovas: now this method used to check that logging configured with DebugLevel and we can add to logger's context more data (now used only for mode of decryption whole|inline) which will be used in inner functions but not unnecessary for default InfoLevel

Lagovas:
return usage of sighup handler as last block function because it didn't fix sighup problem
applied your suggestions as one commit to avoid many commits for each suggestion

decryptor/postgresql/pg_escape_decryptor.go
```diff
@@ -99,7 +106,7 @@ func (decryptor *PgEscapeDecryptor) readOctalData(data, octData []byte, reader i
 			return dataIndex, octDataIndex, err
 		}
 		if n != 1 {
-			log.Debugln("readOctalData read 0 bytes")
+			decryptor.logger.Debugln("readOctalData read 0 bytes")
```
storojs72: As I understand, now we write logs from capital letter

Lagovas: yes but I leaved `readOctalData` as is because it's name of method

PR #314, opened 2019-02-14
fix encoding empty string literals

Lagovas:
previously when application send empty string acra-server took it as `\x` value. It detected that it's postgresql string literal with hex value -> decode -> encrypt (transparent mode) -> encode -> send
Error was in `encode` part when was incorrect check `len(val) > len(hexPrefix) && bytes.Equal(hexPrefix, val[:2])` **greater** instead **greater equal** and it skipped special logic for pg hex literals to avoid escaping first slash. so acra-server changed '\x' -> '\\x' with escaped slash.

plus was incorrect processing `WHERE` constructions with replacing encrypted data. If encrypting of some expression return error then it discards all changes and returns query as is. Even if was used unsupported literal (it's okay, we should just log and skip it). Or when something was wrong with decoding/encoding, databases added new coding format and we don't know how to work with it now. Better if this case will be left as is but other encryptions/decryptions work like before.


vixentael:
So here are two fixes.

1. Correct handling of empty strings from application to PostgreSQL database.
2. Correct handling Transparent encryption with unsupported literals.

Right?

Lagovas:
yes. but both of them related to transparent encryption. other parts didn't touch with these issues.

encryptor/queryDataEncryptor.go
```diff
@@ -107,8 +108,11 @@ func UpdateExpressionValue(expr sqlparser.Expr, coder DBDataCoder, updateFunc fu
 		case sqlparser.StrVal, sqlparser.HexVal, sqlparser.PgEscapeString:
 			rawData, err := coder.Decode(val)
 			if err != nil {
-				if err != errUnsupportedExpression {
-					logrus.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCodingCantDecodeSQLValue).WithError(err).Errorln("Can't decode data")
+				if err == utils.ErrDecodeEscapedString || err == errUnsupportedExpression {
+					logrus.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCodingCantDecodeSQLValue).
+						WithError(err).
+						Warningln("Can't decode data with unsupported coding format or unsupported expression")
```
vixentael: Shall we write anywhere that this is an error while encrypting value in Transparent mode? Because users will get plaintext data due to error, but will see "decoding" error, which is misleading.

Maybe add extra log on line 146/150 "Encryption error"

https://github.com/cossacklabs/acra/pull/314/files#diff-4aa4720957d4bb4bcaf1c7b968531a31R150

Lagovas: it will be logged on a higher level
https://github.com/cossacklabs/acra/blob/47c33ac882700ee750bd3033827936dc40921f7e/encryptor/queryDataEncryptor.go#L213

vixentael: got it, thank you! great.

tests/test.py
```diff
@@ -564,7 +568,7 @@ def execute_prepared_statement(self, query, args=None):
                 cursorclass=pymysql.cursors.DictCursor,
                 **pymysql_tls_args)) as connection:
             with connection.cursor() as cursor:
-                cursor.execute("PREPARE test_statement FROM '{}'".format(query))
+                cursor.execute("PREPARE test_statement FROM {}".format(str(sa.literal(query).compile(dialect=db_dialect, compile_kwargs={"literal_binds": True}))))
```
storojs72: Can you please explain what does it mean? How query will be changed?

tests/test.py
```diff
@@ -1251,7 +1258,7 @@ def testWhitelist(self):
 
         # all those queries should be denied because no matching allow rules specified
         testQueries = ["select * from acrarollback_output",
-                       "insert into test(id, data) values(1, DEFAULT)"]
+                       "insert into test(id, data, empty) values(1, DEFAULT, '')"]
```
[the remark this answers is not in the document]
Lagovas: if table has string/bytearray type and application send empty string/array. Postgresql packet has row description packet that tell that response will have 5 packets and then send 5 packets as pair data_length + data[data_length] and for empty values it send data_length=0 and nothing in data.

if column has constraint `not null` then you can't skip value for this column in `insert` if you didn't declare some default (like NOW() or next(sequence_id)). Detected this issues when database return empty string in last column that was not null and had bytea type )

tests/test.py
Lagovas: in mysql sql query should be string literal or some variable that contains query to execute. Literal should be passed in the same manner as other literals in quotes and may have other quotes and inner literals (for exampe `select * from table where a='123'). Here I use sqlalchemy python library that can create sql queries for specified database (if they have some differences then it will create different sql queries for each dialect) with correct escaping quotes inside string literal. 
As default it return placeholder for value that should be replaced with value before sending to database (sqlalchemy uses ":param1", ":param2" placeholders and then replace with values) but here we need binded values instead placeholder because then this query will be sent without sqlalchemy usage

PR #316, opened 2019-02-15
handle last empty column (fix to master)

Lagovas:
fix the case when last data in data row column from PostgreSQL come with empty data and server try to read 0 bytes, take io.EOF and stop processing connection due to an unexpected error in parsing packets

### DIFF
--- a/decryptor/postgresql/pg_escape_decryptor.go
+++ b/decryptor/postgresql/pg_escape_decryptor.go
@@ -197,7 +197,7 @@ func (decryptor *PgEscapeDecryptor) readDataLength(reader io.Reader) (uint64, []
 		return 0, decryptor.octLengthBuf[:octLenCount], base.ErrFakeAcraStruct
 	}
 	decryptor.outputSize += octLenCount
-	binary.Read(bytes.NewBuffer(decryptor.lengthBuf[:]), binary.LittleEndian, &length)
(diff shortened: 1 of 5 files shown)


decryptor/postgresql/pg_escape_decryptor.go
```diff
@@ -197,7 +197,7 @@ func (decryptor *PgEscapeDecryptor) readDataLength(reader io.Reader) (uint64, []
 		return 0, decryptor.octLengthBuf[:octLenCount], base.ErrFakeAcraStruct
 	}
 	decryptor.outputSize += octLenCount
-	binary.Read(bytes.NewBuffer(decryptor.lengthBuf[:]), binary.LittleEndian, &length)
+	binary.Read(bytes.NewReader(decryptor.lengthBuf[:]), binary.LittleEndian, &length)
```
vixentael: what's the difference?

vixentael:
Same as #315 but for master branch

decryptor/postgresql/pg_escape_decryptor.go
Lagovas: bytes.Reader wrap []byte with method Read to implement io.Reader and just return data from bytes
bytes.NewBuffer has overhead with internal buffer for short slices and store []byte too. NewReader doesn't have extra internal buffer and uses less memory

vixentael: > uses less memory

👍

PR #319, opened 2019-02-21
print version on start

Lagovas:
added logging version at start. it looks now like that:
```
INFO[0000] Starting service acra-server [pid=591]        version=0.84.2
INFO[2019-02-21T14:46:00+02:00] Validating service configuration...
...
```

### DIFF
--- a/cmd/acra-connector/acra-connector.go
+++ b/cmd/acra-connector/acra-connector.go
@@ -209,7 +209,7 @@ type Config struct {
 }
 
 func main() {
-	log.Infof("Starting service %v [pid=%v]", ServiceName, os.Getpid())
+	log.WithField("version", utils.VERSION).Infof("Starting service %v [pid=%v]", ServiceName, os.Getpid())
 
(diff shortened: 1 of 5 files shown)


cmd/acra-connector/acra-connector.go
```diff
@@ -209,7 +209,7 @@ type Config struct {
 }
 
 func main() {
-	log.Infof("Starting service %v [pid=%v]", ServiceName, os.Getpid())
+	log.WithField("version", utils.VERSION).Infof("Starting service %v [pid=%v]", ServiceName, os.Getpid())
```
ilammy: Hm... Why add version as a field but keep service name and PID just a part of the log string?

Lagovas: hm... I don't know what to say) legacy version vs new version)) All our components log at startup like `Starting service BLABLA [pid=123]` and only 3 (server/connector/translator) works as a daemon and good to log version for them. So used field to log instead raw string format

PR #320, opened 2019-02-22
export version as metric

Lagovas:
* export `<service_name>_version_major`, `<service_name>_version_minor`, `<service_name>_version_patch` metrics where <service_name> replaced with _acraserver_ | _acraconnector_ | _acratranslator_ (all services that works as daemon)
* renamed _translator_request_processing_seconds_ -> _**acra**translator_request_processing_seconds_ metric name according to other metrics name
* updated tests that should check that all expected metrics exported and check that they have any value where it possible

### DIFF
--- a/cmd/prometheus.go
+++ b/cmd/prometheus.go
@@ -17,12 +17,17 @@ limitations under the License.
 package cmd
 
 import (
+	"fmt"
 	"github.com/cossacklabs/acra/logging"
 	"github.com/cossacklabs/acra/network"
+	"github.com/cossacklabs/acra/utils"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
 	"net"
 	"net/http"
+	"strings"
+	"sync"
 )
 
 // RunPrometheusHTTPHandler run in goroutine http server that process with connectionString address and export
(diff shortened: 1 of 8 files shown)


cmd/prometheus.go
```diff
@@ -43,3 +48,67 @@ func RunPrometheusHTTPHandler(connectionString string) (net.Listener, *http.Serv
 	}()
 	return listener, server, nil
 }
+
+var registerLock = sync.Once{}
+
+// serviceNameToLabelFormat convert service name to lower case and remove all '_'
```
ilammy: IMO, translating `Acra-Server` into `acra_server` is more readable. But I guess the idea behind _removing_ the dashes is to keep the service name as one word so that it's more easily distinguishable in labels?
There's also typo in the comment (`_` → `-`), we replace dashes, not underscores.

```suggestion
// serviceNameToLabelFormat convert service name to lower case and remove all '-'
```

tests/test.py
```diff
@@ -3377,41 +3380,99 @@ def checkMetrics(self, url, labels=None):
...
     def testAcraConnector(self):
         # connector should has some values in counter after connections checks
         # on setUp
-        labels = [
-            'acraconnector_connections_total',
-            'acraconnector_connections_processing_seconds_bucket',
-        ]
+        labels = {
+            'acraconnector_connections_total': {'min_value': 2},
+            #'acraconnector_connections_processing_seconds': {'min_value': TestPrometheusMetrics.MIN_EXECUTION_TIME},
```
ilammy: I guess this commented-out line can be removed.

```suggestion
```

utils/version.go
```diff
@@ -16,5 +16,65 @@ limitations under the License.
...
+	Patch string
+}
+
+// String format version as string
+func (v *Version) String() string {
+	return fmt.Sprintf("%s.%s.%s", v.Major, v.Minor, v.Patch)
+}
+
+// MajorAsFloat64 return major number as float64
+func (v *Version) MajorAsFloat64() (float64, error) {
```
ilammy: Version components represented as _real_ numbers look very weird to me. Using strings for components storage is weird as well.

I think it would be more natural to store `ints` in `Version`. We validate them as integers in `GetParsedVersion` after all. Then we can cast them as `float64` only in `ExportVersionMetric` as using real numbers looks like an implementation detail of gauges to me.

cmd/prometheus.go
vixentael: @Lagovas is this renaming crucial? 

I would rather not introduce the _third_ type of naming (currently we have `acra-server` as service name, and binary name etc, and `AcraServer` for human-friendly logs and texts).

Lagovas: all our metrics have next naming: `acraconnector_*`/`acraserver_*`/`acratranslator_*` according to our discussion in T662
this function was added to automatically convert serviceName to correct metric's format

utils/version.go
Lagovas: version used only in two places: as metrics where they exported as float64 according to prometheus client interface and as string in log message at the start
as is from `VERSION` variable. Storing as string in `VERSION` var allow override it at compilation time with command `go build -ldflags '-X github.com/cossacklabs/acra/utils.VERSION=<my_new_version>'` via deploying tools with hash of commit or other build's metadata. Golang allow override only string values, so I decided to store it as a string and parse into the structure for metrics or other usages.

in a future version may be extended with pre-release part or with build metadata as it allowed in [semver specification](https://semver.org/). and I thought that it is okay to store as strings and provide getters for specific types if necessary.

PR #321, opened 2019-02-27
censor versioning [T971]

Lagovas:
* added parsing version from censor's config and compare it with hardcoded version `var MinimalCensorConfigVersion = "0.84.2"`. If version in config is older than minimal then log error message and stop acra-server. But don't know how to check it better. Compare with version in config or with version of acra-server (also hardcoded). And which version will be supported by censor, less/greater/equal. When major/patch equals and ignore patch or full version value or maybe set minimal version (as it done in PR) and update hardcoded version after censor's updates?
* when I did this PR golang 1.12 was released so I added it to tests (and removed 1.9.7 as unsupported for now). Plus due to added support of tls1.3 in golang1.12, added tests with new tls version too as separate `go test` run
* because ci froze more frequently when I pushed commits and we decided that it's okay to add loop for tests with re-tries to avoid annoying failures so added it too 
* added exporting test's result in junit xml format to allow circleci parse them and show on his ui - https://circleci.com/docs/2.0/collect-test-data/. After this circleci show statistics with failed/successful tests, time of execution, etc.

### DIFF
--- a/.circleci/check_gotest.sh
+++ b/.circleci/check_gotest.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+# run test in each go environment and create $FILEPATH_ERROR_FLAG file if was any error. But all tests should
+cd $HOME
+for version in $VERSIONS; do
+    export GOROOT=$HOME/go_root_$version/go;
+    export PATH=$GOROOT/bin/:$PATH;
+    export GOPATH=$HOME/$GOPATH_FOLDER;
+    rm -rf $HOME/$GOPATH_FOLDER/bin;
+    rm -rf $HOME/$GOPATH_FOLDER/pkg;
+
+    go test -v github.com/cossacklabs/acra/...;
+    if [ "$?" != "0" ]; then
+        echo "$version-tls12" >> "$FILEPATH_ERROR_FLAG";
+    fi
+
+    # test with supported tls1.3
+    GODEBUG="tls13=1" go test -v github.com/cossacklabs/acra/...;
+    if [ "$?" != "0" ]; then
+        echo "$version-tls13" >> "$FILEPATH_ERROR_FLAG";
+    fi
+done
+
+# if file exists (exit code of stat == 0 ) then something was wrong. cat file with versions of environments where was error and return exit 1
+if [ -f  $FILEPATH_ERROR_FLAG ]; then
+    cat "$FILEPATH_ERROR_FLAG";
+    rm "$FILEPATH_ERROR_FLAG";
+    exit 1;
+fi
--- a/.circleci/check_gotest.sh
+++ b/.circleci/check_gotest.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
-# run test in each go environment and create $FILEPATH_ERROR_FLAG file if was any error. But all tests should
+# Run test in each go environment and log errors to $FILEPATH_ERROR_FLAG.
+# If all tests pass successfully then the file will not be created at all.
 cd $HOME
 for version in $VERSIONS; do
     export GOROOT=$HOME/go_root_$version/go;
@@ -10,19 +11,21 @@ for version in $VERSIONS; do
     rm -rf $HOME/$GOPATH_FOLDER/pkg;
 
     go test -v github.com/cossacklabs/acra/...;
-    if [ "$?" != "0" ]; then
+    status="$?"
+    if [[ "${status}" != "0" ]]; then
         echo "$version-tls12" >> "$FILEPATH_ERROR_FLAG";
     fi
 
     # test with supported tls1.3
     GODEBUG="tls13=1" go test -v github.com/cossacklabs/acra/...;
-    if [ "$?" != "0" ]; then
+    status="$?"
+    if [[ "${status}" != "0" ]]; then
         echo "$version-tls13" >> "$FILEPATH_ERROR_FLAG";
     fi
 done
 
 # if file exists (exit code of stat == 0 ) then something was wrong. cat file with versions of environments where was error and return exit 1
-if [ -f  $FILEPATH_ERROR_FLAG ]; then
+if [[ -f  $FILEPATH_ERROR_FLAG ]]; then
     cat "$FILEPATH_ERROR_FLAG";
     rm "$FILEPATH_ERROR_FLAG";
     exit 1;
--- a/.circleci/integration.sh
+++ b/.circleci/integration.sh
@@ -7,6 +7,12 @@ export TEST_DB_USER=test
 export TEST_DB_USER_PASSWORD=test
 export TEST_DB_NAME=test
 export GOPATH=$HOME/$GOPATH_FOLDER;
+# cirecle ci has timeout 10 minutes without output after that it stop execution
+# set timeout 8 minutes to give a time to re-start tests execution
+export TEST_RUN_TIMEOUT=480 # 8 minutes (8 * 60)
+
+export TEST_OUTPUT_FOLDER="${HOME}/tests_output"
+mkdir -p ${TEST_OUTPUT_FOLDER}
 
 cd $HOME/project
 # set correct permissions for ssl keys here because git by default recognize changing only executable bit
@@ -15,6 +21,7 @@ cd $HOME/project
 find tests/ssl -name "*.key" -type f -exec chmod 0600 {} \;
 for version in $VERSIONS; do
     echo "-------------------- Testing Go version $version"
+    context="golang-$version-tls-${TEST_TLS}"
 
     export TEST_ACRASERVER_PORT=$(expr ${TEST_ACRASERVER_PORT} + 1);
     export TEST_CONNECTOR_PORT=$(expr ${TEST_CONNECTOR_PORT} + 1);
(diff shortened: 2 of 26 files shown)


acra-censor/acra-censor_configuration_provider.go
```diff
@@ -36,24 +41,40 @@ const (
...
+var ErrUnsupportedConfigVersion = errors.New("acra-censor's config is outdated")
+
 // LoadConfiguration loads configuration of AcraCensor
 func (acraCensor *AcraCensor) LoadConfiguration(configuration []byte) error {
 	var censorConfiguration Config
 	err := yaml.Unmarshal(configuration, &censorConfiguration)
 	if err != nil {
 		return err
 	}
+	configVersion, err := utils.ParseVersion(censorConfiguration.Version)
```
vixentael: current censor configs don't have version, so it can be parsed. shouldn't we `return `ErrUnsupportedConfigVersion`?

acra-censor/acra-censor_configuration_provider.go
```diff
@@ -36,24 +41,40 @@ const (
...
+	if err != nil {
+		return err
+	}
+	censorVersion, err := utils.ParseVersion(MinimalCensorConfigVersion)
```
vixentael:
```suggestion
	currentlySupportedVersion, err := utils.ParseVersion(MinimalCensorConfigVersion)
```

acra-censor/acra-censor_configuration_provider.go
```diff
@@ -36,24 +41,40 @@ const (
...
+	if err != nil {
+		return err
+	}
+	if censorVersion.Compare(configVersion) == utils.Greater {
```
vixentael:
```suggestion
	if currentlySupportedVersion.Compare(configVersion) == utils.Greater {
```

acra-censor/acra-censor_test.go
```diff
@@ -1508,6 +1509,9 @@ func TestConfigurationProvider(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	re := regexp.MustCompile(`(version:\s+?)((\d+\.?){3})`)
+	// replace version from exapmle to current
```
vixentael: 😅
```suggestion
	// replace version from example to current
```

acra-censor/acra-censor_configuration_provider.go
```diff
@@ -36,24 +41,40 @@ const (
...
+		// censor has version newer than config
```
vixentael: I think we should log error in human readable form, smth like

```
acraCensor.logger.
  WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCensorSetupError).
  Errorln("AcraCensor config file version is not supported: probably AcraCensor configuration (acra-censor.yaml) is outdated, check docs for deprecation warnings https://docs.cossacklabs.com/pages/documentation-acra/#acracensor-acra-s-firewall")
```

utils/version_test.go
```diff
@@ -0,0 +1,71 @@
...
+func TestVersion_Compare(t *testing.T) {
+	testData := []struct {
+		v1, v2 string
+		result ComparisonStatus
+	}{
+		{"0.0.0", "0.0.0", Equal},
+		{"1.0.0", "0.0.0", Greater},
+		{"0.0.0", "1.0.0", Less},
+
+		{"0.1.0", "0.0.0", Greater},
```
vixentael:
```suggestion
		{"0.1.0", "0.0.0", Greater},
		{"0.84.2", "0.1.0", Greater},
		{"0.84.2", "1.0.0", Less},
```

acra-censor/acra-censor_configuration_provider.go
Lagovas: it will return [this](https://github.com/Lagovas/acra/blob/lagovas/T971-censor-versioning/acra-censor/acra-censor_configuration_provider.go#L57) error and log will looks like:
```
[...] "Can't setup censor" error="VERSION value has incorrect format (semver 2.0.0 format expected, https://semver.org/)"
```
There return another error value to distinguish cases when version is missed/has incorrect value and version is outdated. Maybe better to return 3 different error values: missed version/incorrect format/outdated?

Plus I think to change "VERSION value has incorrect format (semver 2.0.0 format expected, https://semver.org/)" -> "version value has incorrect format (semver 2.0.0 format expected, https://semver.org/)". Because at start it was a reference to variable VERSION

.circleci/check_gotest.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+# run test in each go environment and create $FILEPATH_ERROR_FLAG file if was any error. But all tests should
```
ilammy: I guess we should finish the comme— wait, someone's knocking at the door, I'll go take a look.

```suggestion
# Run test in each go environment and log errors to $FILEPATH_ERROR_FLAG.
# If all tests pass successfully then the file will not be created at all.
```

.circleci/integration.sh
```diff
@@ -25,10 +31,23 @@ for version in $VERSIONS; do
...
-    else
-        echo "no errors";
-    fi
+    for iteration in {1..3}; do
+        context="${iteration}-golang-${version}-tls-${TEST_TLS}"
+        export TEST_XMLOUTPUT="${TEST_OUTPUT_FOLDER}/${context}.xml"
+        LOG_OUTPUT="${TEST_OUTPUT_FOLDER}/${context}.log"
+        timeout ${TEST_RUN_TIMEOUT} python3 tests/test.py -v | tee "${LOG_OUTPUT}";
+        if [[ "$?" != "0" ]]; then
+            status="$?"
```
ilammy: The status should be saved immediately after the failed command. The `$?` here contains the result of the test performed by `if`: i.e, always zero as we're in the success branch.

```diff
         timeout ${TEST_RUN_TIMEOUT} python3 tests/test.py -v | tee "${LOG_OUTPUT}";
-        if [[ "$?" != "0" ]]; then
-                status="$?"
+        status="$?"
+        if [[ "$status" != "0" ]]; then
```

ilammy: And we should also add

```bash
set -o pipefail
```

somewhere at the start of the script. Otherwise the pipes return the exit code of the last command: `tee` in this case. Even if `python` terminates unsuccessfully, `tee` will just see the closed input and will exit cleanly.

tests/test.py
```diff
@@ -1437,8 +1475,8 @@ def tearDown(self):
         if not self.EXTERNAL_ACRA and hasattr(self, 'acra'):
             procs.append(self.acra)
         stop_process(procs)
-        send_signal_by_process_name('acra-server', signal.SIGTERM)
-        send_signal_by_process_name('acra-connector', signal.SIGTERM)
+        send_signal_by_process_name('acra-server', signal.SIGKILL)
+        send_signal_by_process_name('acra-connector', signal.SIGKILL)
```
shadinua: Why do we need to use `SIGKILL` instead of `SIGTERM`? Is there any problem with process termination dusring test?

Lagovas: siterm + sigkill after timeout sent via `stop_process` function above that uses pid's that was returned after expected fork from our code. `send_signal_by_process_name` use pidof to finally kill any zombie processes that may be forked after unexpected sighup signal.  because after sighup the processes that we forked from tests create own fork with new pid that we don't know from test environment and terminate itself.

in general cases `send_signal_by_process_name` will do nothing because all processes was terminated correctly before.

shadinua: Thanks for the explanation.

acra-censor/acra-censor_configuration_provider.go
Lagovas: added a check on empty version and return `
ErrUnsupportedConfigVersion`

acra-censor/acra-censor_configuration_provider.go
Lagovas: changed

acra-censor/acra-censor_test.go
Lagovas: fixed, thanks : )

acra-censor/acra-censor_configuration_provider.go
Lagovas: okay, added

utils/version_test.go
Lagovas: added (but imho it test the same cases : )

.circleci/integration.sh
Lagovas: thanks. nice catch

shadinua: If we're going to make this script really safe, I'd suggest to add `set -uo pipefail` at the start of the script.
It would add checks for unset variables (if that variables are there, they should have default values like a `${MYVAR:-defaultval}`). Often it is useful also to add `-e`, which terminate script on any command fail. In such case we have to handle expected errors inside `if` clauses.

Lagovas: I don't want to force use default values in bash scripts because most of them use environment variables from circleci's config `.circleci/config.yaml`. with such checks, we should to duplicate the same values from `config.yaml` to each bash script

acra-censor/acra-censor_configuration_provider.go
vixentael: thank you, makes sense!

PR #322, opened 2019-03-05
service config version [T1031]

Lagovas:
* extend config generation and handle one parameter: folder where it should be generated
* add check on CI that config files in `configs/` are up to date. it check by generating fresh configs to temporary folder and comparing it to `configs/*` files
* allow comparing versions by parts to provide a way to ignore patch part of version
* moved tests db independent to another workflow `general_validation` where run `.circleci/check_[gofmt|golint|gotest]`
* services check version from config that it equal with major + minor parts and ignore patch part of version
* synchronized error messages for all cmd/* binaries for incorrect config version case

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -251,6 +251,9 @@ func DumpConfig(configPath, serviceName string, useDefault bool) error {
 }
 
 func checkVersion(config map[string]interface{}) error {
+	if config == nil {
+		return nil
+	}
 	configVersion, ok := config["version"]
 	if !ok {
 		return errors.New("config hasn't version key")
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -18,6 +18,7 @@ limitations under the License.
 package cmd
 
 import (
+	"errors"
 	flag_ "flag"
 	"fmt"
 	"github.com/cossacklabs/acra/logging"
@@ -175,6 +176,10 @@ func PrintDefaults() {
 
 // GenerateYaml generates YAML file from CLI params
 func GenerateYaml(output io.Writer, useDefault bool) {
+	// write version as first line in yaml format
+	if _, err := fmt.Fprintf(output, "version: %s\n", utils.VERSION); err != nil {
+		panic(err)
+	}
 	flag_.CommandLine.VisitAll(func(flag *flag_.Flag) {
 		var s string
 		if useDefault {
@@ -245,6 +250,32 @@ func DumpConfig(configPath, serviceName string, useDefault bool) error {
 	return nil
 }
 
+func checkVersion(config map[string]interface{}) error {
+	configVersion, ok := config["version"]
+	if !ok {
+		return errors.New("config hasn't version key")
+	}
+	versionValue, ok := configVersion.(string)
+	if !ok {
+		return errors.New("value of version is not string")
+	}
+
+	version, err := utils.ParseVersion(versionValue)
+	if err != nil {
+		return err
+	}
+
+	serverVersion, err := utils.GetParsedVersion()
+	if err != nil {
+		return err
+	}
+
+	if serverVersion.CompareOnly(utils.MajorFlag|utils.MinorFlag, version) != utils.Equal {
+		return errors.New("config version is outdated")
+	}
+	return nil
+}
+
 // Parse loads CLI params from yaml config and cli
 func Parse(configPath, serviceName string) error {
 	/*load from yaml config and cli. if dumpconfig option pass than generate config and exit*/
@@ -258,6 +289,7 @@ func Parse(configPath, serviceName string) error {
 	if *config != "" {
 		configPath = *config
 	}
+	var yamlConfig map[string]interface{}
 	var args []string
 	// parse yaml and add params that wasn't passed from cli
 	if configPath != "" {
@@ -275,7 +307,6 @@ func Parse(configPath, serviceName string) error {
 			if err != nil {
 				return err
 			}
-			yamlConfig := map[string]interface{}{}
 			err = yaml.Unmarshal([]byte(data), &yamlConfig)
 			if err != nil {
 				return err
@@ -306,6 +337,9 @@ func Parse(configPath, serviceName string) error {
 	if *dumpconfig {
 		DumpConfig(configPath, serviceName, true)
 		os.Exit(0)
+	}
+	if err = checkVersion(yamlConfig); err != nil {
+		return err
 	}
 	return nil
 }
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -274,7 +274,7 @@ func checkVersion(config map[string]interface{}) error {
 	}
 
 	if serverVersion.CompareOnly(utils.MajorFlag|utils.MinorFlag, version) != utils.Equal {
-		return errors.New("config version is outdated")
+		return fmt.Errorf("config version \"%s\" is not supported, expects \"%s\" version", version.String(), serverVersion.String())
 	}
 	return nil
 }
--- a/configs/acra-encryptor.example.yaml
+++ b/configs/acra-encryptor.example.yaml
@@ -1,3 +1,4 @@
+schemas:
 - table: test
   columns:
   - id
--- a/configs/acra-encryptor.example.yaml
+++ b/configs/acra-encryptor.example.yaml
@@ -1,4 +1,3 @@
-schemas:
 - table: test
   columns:
   - id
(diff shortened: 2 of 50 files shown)


cmd/utils.go
```diff
@@ -245,6 +250,35 @@ func DumpConfig(configPath, serviceName string, useDefault bool) error {
...
+		return err
+	}
+
+	serverVersion, err := utils.GetParsedVersion()
+	if err != nil {
+		return err
+	}
+
+	if serverVersion.CompareOnly(utils.MajorFlag|utils.MinorFlag, version) != utils.Equal {
+		return errors.New("config version is outdated")
```
ilammy: Another possibility is that we're trying to use a new config with an old Acra version. I'd rephrase this message into something like

```suggestion
		return errors.New("config version is not supported")
```

It may be also useful to include the versions in the error so that it's clear what config version we have in the file and what is the version of Acra.

configs/regenerate.sh
```diff
@@ -1,11 +1,13 @@
...
+binaries=(server connector translator addzone webconfig rollback keymaker poisonrecordmaker authmanager rotate)
+
+args="--dump_config"
+
+
+for cmd in "${binaries[@]}"; do
+ if [[ "$#" == "1" ]]; then
+    args="--dump_config --config_file=$1/acra-${cmd}.yaml"
+ fi
+ go run ./cmd/acra-${cmd}/*.go `echo ${args}`
```
ilammy: Why use command substitution here instead of simply splicing the arguments?
```suggestion
 go run ./cmd/acra-${cmd}/*.go ${args}
```

tests/test.py
```diff
@@ -3823,6 +3842,155 @@ def testEmptyValues(self):
...
+        services = [i for i in files if os.path.isdir(os.path.join('cmd', i))]
+        self.assertTrue(services)
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            # generate configs for tests
+            subprocess.check_output(['configs/regenerate.sh', tmp_dir])
+
+            for service in services:
+                # remove first line that contains version
+                subprocess.check_output(['sed', '-i', '1d', os.path.join(tmp_dir, service + '.yaml')])
```
ilammy: Various UNIX systems have slightly different **sed** flavors with different command lines. For example, the built-in sed on macOS handles `-i` flag a bit differently. This may get annoying for developers who try running tests outside of Docker containers with Linux inside. (Though I don't know how common that is.)

I wonder if there is a better way to handle this. For example, we could teach the `configs/regenerate.sh` tool to generate configs without a version field (with `--no-version` option), or with a particular version value (`--version=0.100.500`). That way we could avoid unnecessary subprocess calls from tests.

tests/test.py
```diff
@@ -3823,6 +3842,155 @@ def testEmptyValues(self):
...
+        return
+
+    def testStartupWithoutVersionInConfig(self):
+        files = os.listdir('cmd/')
+        services = [i for i in files if os.path.isdir(os.path.join('cmd', i))]
+        self.assertTrue(services)
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            # generate configs for tests
+            subprocess.check_output(['configs/regenerate.sh', tmp_dir])
```
ilammy: Why do we regenerate configurations every time instead of keeping test input under version control?

I think it would be more simple to just keep a dozen of files ready to use, they will not suddenly change and lead to brittle tests. And if the tests _do_ fail one day then this indicates a breaking change in configuration format.

utils/version.go
```diff
@@ -43,17 +43,61 @@ const (
 	Less    ComparisonStatus = iota - 1 // -1
 	Equal                               // 0
 	Greater                             // 1
+	InvalidFlags
```
ilammy: This seems out-of-place here. Isn't this an error which should be reported accordingly?

I'd expect the comparison flags to be inlined at every call site. Therefore using an incorrect set of flags is definitely a programmer error and we can panic if the flags are invalid.

configs/regenerate.sh
Lagovas: because this script do only one thing: generate all configs and doesn't require users to pass any args. plus script uses only args that support all binaries from `cmd/*`. it's general parameters like '--help'/'--config_filg'/'--dump_config'. all other parameters may be differ between all binaries

tests/test.py
Lagovas: why:
1. to test this script too
2. to not maintain these pre-generated files. if we add new binary in the future, we will need update `configs/regenerate.sh`, then update all test configs. now here we test 3 cases with different types of changes in configs. It's 3 folders with configs that we will need to remember what difference between these configs and regenerate them. Or add +3 scripts for each of these folders that will regenerate configs with all requirements for tests. And don't forget to remove these folders if we will remove any of test-cases.

in general, for me, it was more simple to call the script from test script instead pre-generate test files

utils/version.go
Lagovas: it's a bad pattern to decide what should do unknown program. To handle panics in golang programs programmer should to wrap this function which will defer function with `recover()`` [call](https://github.com/golang/go/wiki/PanicAndRecover) if he wants to handle an error in place where it has occurred

I added this flag for cases when wasn't used any flag (pass 0) or incorrect value. Function should compare two values and return any less/equal/greater. But in other cases with undefined (oh, maybe use `undefined`?) value used some value that will show it and allow to handle it.

Anyway programmer will test his code with comparisons and catch his mistakes. He can write `if v1.CompareOnly(0, v2) <operator> <value` where operator may be == or != (>, >=, <=, < is unexpected and not supported operators as any of other implementations of `compare` in golang/c/c++). If he expect that result ` == Equal` or ` != Equal`, he will take it anyway if pass incorrect flag : )

or we need +1 opinion how will be better in this case )

tests/test.py
Lagovas: oh. first time I wanted to do it from python script but thought that it's will be too complicated with reading file into memory + replacing + dumping to file and decided to do it with sed. then thought that we can use 2-step replacing with sed: sed to temp file and rename but then remembered that he can do it with in place mode. chose second variant and forget about posix compatibility.

I don't want to extend `configs/regenerate.sh` because in general it will not be useful for users because they should use up to date config versions and script was created as helper to create all configs for current version of acra. Maybe will be good to add separate bash script for tests which will do this? Which will replace in 2-step way or forget about overhead and do it in python because config files has small size (biggest is 4.1k))...

ilammy: The version flags can be optional:

- no flags: generate configuration with current version
- `--no-version`: generate configuration without a version field
- `--version <V>`: generate configuration with specified version field

Though, in all cases we will generate a configuration _for_ the current version, just with different version field. Which is kinda hacky, I admit. So I don't know whether it's good to add these flags to regenerate.sh because we need them only for tests.

Anyway, this is a minor annoyance do it can be ignored if there is no one-liner fix for it.

I've seen (portable?) utilities like **yq** which can process YAML from command-line, but they are a bother to install. There should be Python modules which do the same, but that's another dependency to install.

Lagovas: we already have pyYaml in dependencies for tests so I will rewrite it with replacing on tests side

support `--no-version`/`--version` in `configs/regenerate.sh` will meet the same problem: replace version with some tool like sed/etc and do it in a posix-compatible way and it will be useful only for tests.
add it on acra side will be complicated due to different handling flags by each binary or write in general part of code and it will be useful only for tests too ) And with such parameters we allow to use/generate configs without versions but it is not that we want. Because we added versions and force them to simplify discovering user's problems and misconfigurations in the future. And probably will drop support configs without versions in a future

cmd/utils.go
vixentael: agree, makes sense to rename

configs/acra-encryptor.example.yaml
```diff
@@ -1,4 +1,3 @@
-schemas:
```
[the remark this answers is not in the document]
vixentael: @Lagovas I'm really concerned about this one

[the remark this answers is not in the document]
Lagovas: it's a mistake. nice catch. maybe removed locally and added via `git add` by mistake

cmd/utils.go
Lagovas: okay, will add this

configs/acra-encryptor.example.yaml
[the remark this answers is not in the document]
vixentael: ok, thank you!

PR #324, opened 2019-03-07
Update acrawriter to import Themis in a new way

vixentael:
Themis `0.10.4` brings modular frameworks on Swift, which has changed a bit a way how it's imported inside AcraWriter module.

New acrawriter.podspec is coming after merging this PR.


vixentael:
This changes doesn't affect any of Acra tests, except iOS wrapper tests.

vixentael:
Next PR: #325

PR #325, opened 2019-03-07
Update acrawriter.podspec

vixentael:
After Themis started to support Swift frameworks, `acrawriter.podspec` could be simplified.

1. Remove unneeded `s.ios.xcconfig` and `s.osx.xcconfig` configurations.
2. Stick with Themis 0.10.4 and above
3. Podspec source file will be re-linked after #324 merge

This PR will be updated after merging #324


vixentael:
Podspec was pushed successfully

PR #326, opened 2019-03-07
Add Swift Acrawriter examples

vixentael:
As soon as new` acrawriter.podspec` supports Swift frameworks, I thought it makes sense to add Swift example as well.

1. Updated ObjC example to use new acrawriter.
2. Added Swift example.

Will be updated and merged after #324 #325


examples/swift/AcraWriterSwift/AcraWriterSwift/AppDelegate.swift
```diff
@@ -0,0 +1,76 @@
...
+import UIKit
+import acrawriter
+
+@UIApplicationMain
+class AppDelegate: UIResponder, UIApplicationDelegate {
+
+    var window: UIWindow?
+
+
+    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
```
vixentael: real code is here

PR #328, opened 2019-03-11
handle casted placeholders [T1075]

Lagovas:
*support type cast for placeholders: `select $1::type1::type2 from table1 where column1=$2::type3::type4`

before placeholders were as the [separate expression](https://github.com/cossacklabs/acra/blob/master/sqlparser/ast.go#L1593) and worked only with placeholders without any casts.
to add cast support it needed to duplicate same logic as it was done for [SQLVal with extra field](https://github.com/cossacklabs/acra/blob/master/sqlparser/ast.go#L1206) for expression struct that will contain type info as bytes and extend `format` method. But this way has one disadvantage that we can't handle such expressions as `sqlval` in our censor rules and encryptor's logic. 
So, to handle it better, it to work with placeholders as with `sqlval`'s and extend rule for `typecast` to support more than one type.


sqlparser/sql.go
```diff
@@ -618,572 +619,521 @@ var yyExca = [...]int{
 	153, 272,
 	-2, 262,
 	-1, 248,
-	111, 616,
+	111, 617,
```
vixentael: o.o

vixentael: this code is fully generated, right?

Lagovas: sql.go is auto-generated from sql.y via goyacc

sqlparser/parse_test.go
```diff
@@ -1577,6 +1577,8 @@ func TestConvert(t *testing.T) {
 		input: "select convert('abc', json) from t",
 	}, {
 		input: "select convert('abc' using ascii) from t",
+	}, {
+		input: "select $1::integer::bytea from dual where test = $2::varchar::bytea",
```
storojs72: `$1::integer::bytea` - this is valid only in postgres, right?

Lagovas: yes, only postgres

PR #329, opened 2019-03-13
More user-friendly config descriptions for AcraConnector, AcraServer

vixentael:
I've added more clarity in configs for AcraConnector and AcraServer.

@Lagovas haven't I missed smth?


cmd/acra-server/acra-server.go
```diff
@@ -106,12 +106,12 @@ func main() {
 	enableHTTPAPI := flag.Bool("http_api_enable", false, "Enable HTTP API")
 
 	useTLS := flag.Bool("acraconnector_tls_transport_enable", false, "Use tls to encrypt transport between AcraServer and AcraConnector/client")
-	tlsKey := flag.String("tls_key", "", "Path to private key that will be used in TLS handshake with AcraConnector as server's key and Postgresql as client's key")
+	tlsKey := flag.String("tls_key", "", "Path to private key that will be used in TLS handshake with AcraServer as server's key and database as client's key")
```
vixentael: between AcraServer and database, right?

Lagovas: client/acra-connector <-> acra-server and acra-server <->database. So leave `with AcraConnector` as it was before : )

cmd/acra-connector/acra-connector.go
```diff
@@ -229,12 +229,12 @@ func main() {
 	tlsCert := flag.String("tls_cert", "", "Path to certificate")
 	tlsAcraserverSNI := flag.String("tls_acraserver_sni", "", "Expected Server Name (SNI) from AcraServer")
 	tlsAuthType := flag.Int("tls_auth", int(tls.RequireAndVerifyClientCert), "Set authentication mode that will be used in TLS connection with AcraServer/AcraTranslator. Values in range 0-4 that set auth type (https://golang.org/pkg/crypto/tls/#ClientAuthType). Default is tls.RequireAndVerifyClientCert")
-	noEncryptionTransport := flag.Bool("acraserver_transport_encryption_disable", false, "Use raw transport (tcp/unix socket) between acraserver and acraproxy/client (don't use this flag if you not connect to database with ssl/tls")
+	noEncryptionTransport := flag.Bool("acraserver_transport_encryption_disable", false, "Use raw transport (tcp/unix socket) between AcraServer and AcraConnector/client app (don't this flag if you don't connect to database with TLS)")
```
Lagovas: here the same. with this flag, we turn off the requirement of acra-connector between app and acra-server and client's app can create a direct connection to acra-server without acra-connector.

cmd/acra-server/acra-server.go
vixentael:
```suggestion
	tlsKey := flag.String("tls_key", "", "Path to private key that will be used in AcraServer's TLS handshake with AcraConnector as server's key and database as client's key")
```

vixentael: ah, got it!

cmd/acra-server/acra-server.go
```diff
@@ -106,12 +106,12 @@ func main() {
...
 	tlsCA := flag.String("tls_ca", "", "Path to root certificate which will be used with system root certificates to validate Postgresql's and AcraConnector's certificate")
-	tlsDbSNI := flag.String("tls_db_sni", "", "Expected Server Name (SNI) from Postgresql")
-	tlsAuthType := flag.Int("tls_auth", int(tls.RequireAndVerifyClientCert), "Set authentication mode that will be used in TLS connection with Postgresql. Values in range 0-4 that set auth type (https://golang.org/pkg/crypto/tls/#ClientAuthType). Default is tls.RequireAndVerifyClientCert")
-	noEncryptionTransport := flag.Bool("acraconnector_transport_encryption_disable", false, "Use raw transport (tcp/unix socket) between AcraServer and AcraConnector/client (don't use this flag if you not connect to database with ssl/tls")
+	tlsDbSNI := flag.String("tls_db_sni", "", "Expected Server Name (SNI) from database")
+	tlsAuthType := flag.Int("tls_auth", int(tls.RequireAndVerifyClientCert), "Set authentication mode that will be used in TLS connection with database. Values in range 0-4 that set auth type (https://golang.org/pkg/crypto/tls/#ClientAuthType). Default is tls.RequireAndVerifyClientCert")
```
Lagovas: actually here was incorrect description and this parameter related to handshake connector/client <-> acra-server and configure how acra-server should validate client's connections

cmd/acra-connector/acra-connector.go
vixentael:
```suggestion
	noEncryptionTransport := flag.Bool("acraserver_transport_encryption_disable", false, "Use raw transport (tcp/unix socket) between AcraServer and client app directly, omitting AcraConnector (don't use this flag if database connection is not TLS)")
```

vixentael: @Lagovas take a look

cmd/acra-server/acra-server.go
vixentael:
```suggestion
	tlsAuthType := flag.Int("tls_auth", int(tls.RequireAndVerifyClientCert), "Set authentication mode that will be used in TLS connection with AcraConnector. Values in range 0-4 that set auth type (https://golang.org/pkg/crypto/tls/#ClientAuthType). Default is tls.RequireAndVerifyClientCert")
```

cmd/acra-connector/acra-connector.go
Lagovas: `if app doesn't connect with TLS`. Here important that raw data from acra-server will go over encrypted connection after decryption. Less important connection between acra-server<->database because data will be encrypted with acrawriter.

PR #334, opened 2019-03-19
add dependencies control [T1069]

Lagovas:
* use golang modules as dependency management tool
* force modules usage in circleci config
* update dockerfile to build with correct dependencies
* small refactoring some bash scripts: split with newlines several commands at one line
* add Gopkg.[lock|toml] files to fix dependencies for `go dep` which may be used with old golang versions (but not sure that current library versions support old versions of golang ... )


.circleci/prepare.sh
```diff
@@ -1,12 +1,28 @@
...
+cd themis
+sudo make install
+cd $HOME
+rm -rf themis
+for version in $VERSIONS; do
+    mkdir go_root_$version;
+    cd go_root_$version;
+    wget https://storage.googleapis.com/golang/go$version.linux-amd64.tar.gz;
+    tar xf go$version.linux-amd64.tar.gz;
+    cd -;
```
ilammy: I think it would be better to use `cd $HOME` instead of `cd -` for consistency. (Or vice versa.)

Alternatively, we could use the following on lines 12–13:
```
mkdir $HOME/go_root_$version
cd $HOME/go_root_$version
```

Finally, we don't have to write trailing semicolons (`;`) now that we use newlines.

.circleci/prepare.sh
```diff
@@ -1,12 +1,28 @@
...
+cd $HOME
+git clone https://github.com/cossacklabs/themis
+cd themis
+sudo make install
+cd $HOME
+rm -rf themis
+for version in $VERSIONS; do
+    mkdir go_root_$version;
+    cd go_root_$version;
+    wget https://storage.googleapis.com/golang/go$version.linux-amd64.tar.gz;
```
ilammy: I'd write this more explicitly like this
```suggestion
    wget https://storage.googleapis.com/golang/go${version}.linux-amd64.tar.gz;
```
(as I constantly forget what are the separators in complex variable usages like this one)

Or even more readable:

```shell
archive=go${version}.linux-amd64.tar.gz
wget https://storage.googleapis.com/golang/$archive
tar xf $archive
```

PR #337, opened 2019-04-01
Fix image name in mysql-no-ssl compose file

shadinua:


Lagovas:
because our image for mysql configured to require tls and doesn't allow connections without tls

shadinua:
Yes. It was a typo, thought. There are no reason to build own no-ssl mysql image, because SSL is disabled in default one.

PR #340, opened 2019-05-24
T1157: Fix ACRA_MASTER_KEY 

AlexShmelev:
Use different master key for acra-server and acra-connector


vixentael:
great, thank you!

PR #342, opened 2019-10-04
Fix integration tests [T1353]

Lagovas:
Integration tests failed due to 2 problems:
* compilation binaries as startup
* sqlalchemy and his drivers to mysql/mariadb added new and changed old queries which they send to db before executing users query (like setup correct charset, fetch some db metadata, default transaction mode, etc) and they are not in our whitelist config which used in tests and censor drop connection with database (as it should be when he doesn't know these queries : )
When we added module support golang need more time to fetch all dependencies
Our integration test compile all needed binaries at start and then use them and has timeout 2 minutes
Now compilation with fetching dependencies sometimes takes more than 2min and raise exception.
**Solution:**
* Added fetching all dependencies before running test
* Added new queries to whitelist

P.S. was surprised when saw that failed mysql-ssl and mariadb tests. their faults should be the same but it wasn't. and I found misnaming circleci jobs : )

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -184,7 +184,7 @@ jobs:
       - store_artifacts:
           path: /home/user/tests_output
 
-  mysql:
+  mysql-ssl:
     docker:
     - image: cossacklabs/android-build
     # use the same credentials for mysql db as for postgresql (which support was added first)
@@ -225,7 +225,7 @@ jobs:
     - store_artifacts:
         path: /home/user/tests_output
 
-  mysql-ssl:
+  mysql:
     docker:
       - image: cossacklabs/android-build
       - image: mysql:5.7.25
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -184,7 +184,7 @@ jobs:
       - store_artifacts:
           path: /home/user/tests_output
 
-  mysql:
+  mysql-ssl:
```
vixentael: good catch!

.circleci/integration.sh
```diff
@@ -30,9 +30,10 @@ for version in $VERSIONS; do
     export PATH=$GOROOT/bin/:$PATH;
 
     # remove built packages with another golang version and force to rebuild
-    rm -rf $GOPATH/pkg
```
vixentael: no need to rm rf anymore?

Lagovas: `go clean` below do this

PR #346, opened 2020-01-16
refactor data serialization

Lagovas:
in our first version of parsing data from postgresql database we read per-byte and try to convert hex/octal values and then try to match zone/acrastruct

after processing we encoded output to one format configured at start: in hex or octal version. 
this way is not optimal because we have 3 version of parsers: binary, hex and octal. but logic the same and should work only with binary format.

plus we should support all versions of encoded data because it may be switched at runtime in database and return output in same format. now we support only hex+bin or octal+bin and return always in hex or octal

in this code change, we add abstraction for data in db responses with api to fetch binary data and encode to correct format which was before decoding. remove all format dependent decryptor and zone matchers (hex/octal), leave only binary which same for all dbs and formats. it used for mysql and postgresql

because  we remove redundant decryptors, some related code which use it was refactored too (actually changes was migrated from acra-ee, related general parts which planned to migrate to CE version). Refactored way

with these changes we deprecate `--pgsql_escape_bytea` and `--pgsql_hex_bytea` flags, they will not change anything but acra will support both transparently

because this change doesn't change any functionality and refactor only serialization/deserialization logic, integration tests wasn't touched and updated only unit-tests


decryptor/postgresql/pg_decryptor.go
```diff
@@ -291,9 +351,15 @@ func checkInlinePoisonRecordInBlock(block []byte, decryptor base.Decryptor, logg
 			if index == utils.NotFound {
 				return nil
 			}
-			currentIndex += index
-			if err := checkWholePoisonRecord(block[currentIndex:], decryptor, logger); err != nil {
-				return err
+			if len(block[index:]) > base.GetMinAcraStructLength() {
```
ilammy: If there is not enough space to fit even a header then why not exit early instead of continuing the loop?

decryptor/postgresql/pg_decryptor.go
```diff
@@ -625,64 +686,21 @@ func (proxy *PgProxy) ProxyDatabaseConnection(errCh chan<- error) {
...
-					err = proxy.processInlineBlockDecryption(packetCtx, packetHandler, column, proxy.decryptor, decryptionLogger)
-					if err != nil {
-						decryptionLogger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorDecryptorCantDecryptBinary).WithError(err).Errorln("Can't process block with inline mode")
-						errCh <- err
-						return
-					}
-				}
-			} else {
-				logger.Debugln("Skip decryption because length of block too small for ZoneId or AcraStruct")
+			newData, err := proxy.onColumnDecryption(packetCtx, i, column.Data)
```
ilammy: Will only column index be enough here? Why not complete column description?

decryptor/postgresql/pg_general_decryptor.go
```diff
@@ -476,15 +397,176 @@ func (decryptor *PgDecryptor) MatchZoneInBlock(block []byte) {
...
+	return decryptor.binaryDecryptor.GetTagBeginLength()
 }
 
 // SetDataProcessor replace current with new processor
 func (decryptor *PgDecryptor) SetDataProcessor(processor base.DataProcessor) {
 	decryptor.dataProcessor = processor
 }
+
+// OnColumn handler which process column data on db response and try to decrypt/detect poison record
+func (decryptor *PgDecryptor) OnColumn(ctx context.Context, data []byte) (context.Context, []byte, error) {
```
ilammy: Do I understand it correctly that this method is effectively tested by existing integration tests for poison records?

These 160+ lines of dense code look scary to merge without tests.

decryptor/postgresql/pg_general_decryptor.go
```diff
@@ -476,15 +397,176 @@ func (decryptor *PgDecryptor) MatchZoneInBlock(block []byte) {
...
+				// check that it's not poison record
+				err = checkInlinePoisonRecordInBlock(data, decryptor, logger)
+			}
+
+			if err != nil {
+				logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorDecryptorCantCheckPoisonRecord).WithError(err).Errorln("Can't check poison record in block")
+				return ctx, nil, err
+			}
+			// TODO remove to back decryption only with matched zone_id
+			//return ctx, data, nil
```
vixentael: 🤔 didn't get what this TODO means (why `return` is commented out).

encryptor/dbDataCoder_test.go
```diff
@@ -161,7 +156,8 @@ func TestPostgresqlDBDataCoder_Decode(t *testing.T) {
 }
 
 func TestPostgresqlDBDataCoder_Encode(t *testing.T) {
-	testData := []byte("some data")
+	testData := make([]byte, 100)
+	rand.Read(testData)
```
vixentael: smart!

decryptor/postgresql/pg_general_decryptor.go
```diff
@@ -476,15 +397,176 @@ func (decryptor *PgDecryptor) MatchZoneInBlock(block []byte) {
...
 func (decryptor *PgDecryptor) SetDataProcessor(processor base.DataProcessor) {
 	decryptor.dataProcessor = processor
 }
+
+// OnColumn handler which process column data on db response and try to decrypt/detect poison record
+func (decryptor *PgDecryptor) OnColumn(ctx context.Context, data []byte) (context.Context, []byte, error) {
+	logger := logging.GetLoggerFromContext(ctx)
+	span := trace.FromContext(ctx)
+	// try to skip small piece of data that can't be valuable for us
+	if (decryptor.IsWithZone() && len(data) >= zone.ZoneIDBlockLength) || len(data) >= base.GetMinAcraStructLength() {
```
vixentael: just IMHO.

For better readability, I suggest to change order:
- first, test that data len is too short and exit
- then, continue to parse AcraStruct assuming that data len is enough

This will change code structure from

```
if (data len is enough) {
 many many
 many many
 many many
 lines
}

"if data len is too short" condition branch, easy to miss
```

to

```
if data len is too short <- visible immediately
  return

// else 
many many 
many lines of code
```

decryptor/postgresql/pg_general_decryptor.go
```diff
@@ -476,15 +397,176 @@ func (decryptor *PgDecryptor) MatchZoneInBlock(block []byte) {
...
+		var err error
+		if decryptor.IsWholeMatch() {
+			newData, err = decryptor.processWholeBlockDecryption(ctx, data, decryptionLogger)
+			if err != nil {
+				decryptionLogger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorDecryptorCantDecryptBinary).WithError(err).Errorln("Can't process whole block")
+				return ctx, nil, err
+			}
+			return ctx, newData, err
+		}
+		newData, err = decryptor.processInlineBlockDecryption(ctx, data, decryptionLogger)
```
vixentael:
```suggestion
                // else "injected cell mode", inline acrastruct
		newData, err = decryptor.processInlineBlockDecryption(ctx, data, decryptionLogger)
```

decryptor/postgresql/pg_general_decryptor.go
```diff
@@ -476,15 +397,176 @@ func (decryptor *PgDecryptor) MatchZoneInBlock(block []byte) {
...
+	logger := logging.GetLoggerFromContext(ctx)
+	span := trace.FromContext(ctx)
+	// try to skip small piece of data that can't be valuable for us
+	if (decryptor.IsWithZone() && len(data) >= zone.ZoneIDBlockLength) || len(data) >= base.GetMinAcraStructLength() {
+		decryptor.Reset()
+		span.AddAttributes(trace.BoolAttribute("decryption", true))
+
+		// Zone anyway should be passed as whole block
+		// so try to match before any operations if we process with ZoneMode on
+		if decryptor.IsWithZone() && !decryptor.IsMatchedZone() {
```
vixentael: _random suggestion, don't shout @ me_

maybe it makes sense to extract this `if` into separate method `matchZone(decryptor, data, logger)`? To improve readability of this code

.circleci/check_golint.sh
```diff
@@ -4,7 +4,7 @@
 # golint print issues it found in source files
 # count lines with wc and check that N lines was in output
 result=$($GOPATH/bin/golint ./... | wc -l)
-if [[ $result -gt 10 ]]; then
+if [[ $result -gt 12 ]]; then
```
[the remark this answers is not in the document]
Lagovas: I surprised, that now tests didn't pass due to golint check with 20+ checks on master branch. As I think, maybe he updated some checks after new versions of golang... Previously we left 10 entries from golint which suggest to remove redundant words from names like `type AcraCensorInterface interface` (it suggested remove Interface suffix), or remove `JSON` from `ZoneInfoJSON` because it declared in `json` package ). Now such suggestions 12 items...

cmd/acra-server/acra-server.go
```diff
@@ -91,7 +91,7 @@ func main() {
 	keysCacheSize := flag.Int("keystore_cache_size", keystore.InfiniteCacheSize, "Maximum number of keys stored in in-memory LRU cache in encrypted form. 0 - no limits, -1 - turn off cache")
 
 	_ = flag.Bool("pgsql_hex_bytea", false, "Hex format for Postgresql bytea data (default)")
-	pgEscapeFormat := flag.Bool("pgsql_escape_bytea", false, "Escape format for Postgresql bytea data")
+	flag.Bool("pgsql_escape_bytea", false, "Escape format for Postgresql bytea data")
```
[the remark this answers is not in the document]
Lagovas: Agree, will add

decryptor/base/decryptionNotification.go
```diff
@@ -0,0 +1,71 @@
...
+// NewContextWithColumnInfo return new context with assigned column metadata
+func NewContextWithColumnInfo(ctx context.Context, info ColumnInfo) context.Context {
+	return context.WithValue(ctx, columnInfoKey{}, info)
+}
+
+// ClientZoneInfo store metadata of matched client/zone id and zonemode
+type ClientZoneInfo struct {
+	ClientID []byte
+	ZoneID   []byte
+	WithZone bool
```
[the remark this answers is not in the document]
Lagovas: `WithZone` responsible for decryption mode, when ClientID/ZoneID is metadata of decryption. So it's real case when acra-server ran in `withzone` mode, but at decryption time ZoneID is empty, and clientID is not empty (always not empty if SecureSession transport was used). To explicitly separate mode of decryption and metadata, was added field WithZone at start.

decryptor/base/decryptionNotification.go
```diff
@@ -0,0 +1,71 @@
...
+// ColumnInfo interface describe available metadata for column
+type ColumnInfo interface {
+	Index() int
+	Alias() string
+}
+
+type columnInfo struct {
+	index int
+	alias string
+}
```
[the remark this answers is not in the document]
Lagovas: When I added ColumnInfo, I knew that at the future we will be able to provide more info about columns (type, db specific info, some other context)
ClientZoneInfo it's final state of data which will be useful in that context. For now to decrypt info we need client_id or zone_id and ZoneMode.
But when I write these rows, I think that it would be better to use ClientZoneInfo as interface too...

decryptor/postgresql/packet_handler.go
```diff
@@ -123,10 +124,12 @@ func (packet *PacketHandler) sendMessageType() error {
...
 // ColumnData hold column length and data
 type ColumnData struct {
-	LengthBuf [4]byte
-	Data      []byte
-	changed   bool
-	isNull    bool
+	LengthBuf   [4]byte
+	Data        []byte
+	data        []byte
+	decodedData *utils.DecodedData
```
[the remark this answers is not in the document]
Lagovas: agree, will update it

decryptor/postgresql/pg_general_decryptor.go
Lagovas: previously acra not able to match zone and decrypt acrastruct at same iteration. but when work on acra-ee I meet some case when it was need (all these changes it's migration general parts from acra-ee). I commented to mark place in code when and what was changed to decide later do we need back it or no. I think we can remove it

decryptor/postgresql/pg_general_decryptor.go
Lagovas: will add in commit to avoid merging after commiting from github

PR #352, opened 2020-02-25
Initial Key Store v2

ilammy:
Add initial version of Key Store v2 (code name: Sonic the Hedgehog – because key _rings_). This is cleaned up, rewritten, and expanded prototype code which implements more complex data model for stored keys, aimed at better compliance with NIST SP 800-57 recommendations.

Now "key store" stores not simply 'keys' but 'key rings' that contain multiple key versions in them, as well as metadata describing keys: their purpose, validity range (cryptoperiod), and current state.

Key ring descriptions are stored as structured data described by ASN.1 grammar and serialized using DER. This allows us to sign resulting key ring data to ensure that it is not modified by a third-party.

This commit includes a "filesystem keystore" abstraction similar to the one we have right now. It provides an interface to a file-base key store with an additional intermediate interface allowing to plug arbitrary backends (may be useful to support key-value stores, for example).

Note that the implementation is somewhat complicated because we need to update key ring data atomically and prevent possible data loss, either due to power failures, application failures, or concurrent modification.

Also note that this is only an initial version, so it currently lacks some important features:

  - The key store is not integrated with the rest of the services. It does not implement the KeyStore V1 interface right now. Implementing high-level key store interface on top of this one is the next step.

  - Key directory integrity is not verified. Current implementation verifies that key ring data is intact, but does not verify that key rings themselves are not removed or entirely replaced. The data model includes a 'key directory' entity to track this, but they are not implemented currently.

----

Please lynch me. This is probably a textbook example of _second-system syndrome_: somehow our arguably elegant, file-based key store v1 turned into... this. 30 files. 4 kloc. How did I end up writing this abomination? It also looks way overengineered and overextenisble without cause. However, we _do want_ to store all this data, _do want_ to keep it consistent and authenticated, and _do *not* want_ to simply put this into some specific database because key data can be stored in external KMS. So... it seemed reasonable to me.

Anyhow, I want to believe that some of this code can be removed and simplified in the future.

Some preemptive replies for possible questions:

- We return and operate on interfaces where possible. Dynamic dispatch will allow us to easily replace the implementation completely, if the filesystem abstraction turns out to be inadequate for some storages. Existing code (keystore v1) is really tightly coupled with `filesystem.KeyStore` and I'd like to avoid this in v2.

- Testing abstract interfaces (not concrete implementations) is really weird in Go. This is reason behind `api` and `api/tests` packages: we need to decouple tests from concrete implementations and to avoid circular import references. In the end, this should allow reuse of these tests in Acra Enterprise code without copy-pasting them.

- I generously split code into files because I find files longer than 200-300 lines of code to be hard to navigate, and a possible smell of God object implemented by them. So here we are, at 30 files, and all those nested packages.

- Some methods of `filesystem.Backend` are currently not used, but I found them useful for atomic updates and I believe they come into play when key directories are implemented.

Things to pay attention to:

- Private/symmetric key data encryption. Grep by `encrypt` and `decrypt`, they are usually in their own section.

- Entire `keystore/signature` package which computes and verified signatures on stored data.

- API in general. Other modules will be using more high-level API, like the current keystore API:

  ```go
   RotateSomeKey(clientID []byte) (*keys.PublicKey, error)
   GetSomePrivateKey(clientID []byte) (*keys.PrivateKey, error)
   GetSomePublicKeys(clientID []byte) ([]*keys.PublicKey, error)
   GetSomePublicKey(clientID []byte) (*keys.PublicKey, error)
   ```

   However, if you find anything silly or questionable, please mark it.

This PR targets an integration branch, not `master`. New key store will be cooked there for a while, before it can be integrated into main code base.

----

Refactoring before merging:

- [X] Standalone `UnmarshalThing(data []byte) (*Thing, error)` functions
- [X] Remove custom-made `Split/Join()` utilities by accepting any separator

Things that require a decision before further work, but may be not addressed in this PR:

- [x] Discuss Key & KeyRing API
- [x] Think about reusing `hash.Hash`


keystore/v2/keystore/signature/notary.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// Notary cryptographically signs provided ASN.1 data.
+type Notary struct {
+	hmac KeyedHMAC
+	log  *log.Entry
+}
+
+// NewNotary makes a new notary with given encryptor.
+func NewNotary(hmac KeyedHMAC) (*Notary, error) {
```
Lagovas: why not to accept just hash.Hash?

keystore/v2/keystore/asn1/asn1.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// VerifiedContainer is an unmarshaled form of SignedContainer.
+type VerifiedContainer struct {
+	Payload    VerifiedPayload
+	Signatures []Signature `asn1:"set"`
+}
+
+// Unmarshal from bytes.
+func (container *VerifiedContainer) Unmarshal(data []byte) error {
```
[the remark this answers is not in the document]
ilammy: > can we just use one method `Unmarshal(data[]byte)(*SignedContainer, error)`? 

Yeah, that API should be better and more consistent. I don’t really remember why it is like this, I _had_ some reason for in-place unmarshaling but then simply left it as is. That reason is probably not applicable anymore. Ditto with `Unmarshal` methods for other types. I’ll look into replacing them.

keystore/v2/keystore/filesystem/backend/utils.go
```diff
@@ -0,0 +1,45 @@
...
+package backend
+
+import (
+	"bytes"
+
+	"github.com/cossacklabs/acra/keystore/v2/keystore/filesystem/backend/api"
+)
+
+// SplitPath splits a path into components by separator.
+func SplitPath(path string) []string {
```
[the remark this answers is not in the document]
ilammy: > can we replace `SplitPath` with `filepath.Split` or `strings.Split`?
>
> [is `JoinPath`] duplicate of `filepath.Join`?

These functions were motivated by the following considerations:

- Since “key paths” are not file paths we should not use `os.PathSeparator` for them. Instead we’d better define _our_ path separator as always `/`. However, writing `strings.Split(path, "/")` seems like unnecessary hardcoding, `strings.Split(path, backend.PathSeparator)` looks too verbose, so I’ve just made these two utilities.

- Another realization that I’ve had is that `strings.Split()` works on _strings_ (as in UTF-8 encoded Unicode text). However, in offline discussions I remember some ideas about using binary data as client or zone IDs. While Go allows to keeps arbitrary bytes in `string`, all those string utilities expect actual text. In order to get correct behavior we should use `bytes.Split` instead. Having separate utilities which are consistently used makes it easier to choose them and not choose `bytes` when you need `strings` or vice versa.

However, now that you’ve mentioned it, we could probably do like Windows does: have some canonical path separator but accept either of them. It just boils down to putting every paths through some

```go
var pathSeparators = strings.NewReplacer(
        "/", os.PathSeparator,
        "\\", os.PathSeparator,
)

// when you need to pass a path to OS:
os.Open(pathSeparators.Replace(path))
```

And then just let the code use `filepath.Join()` if they want, or whatever. We’d need to take special care _if_ we us binary data in paths, but maybe accepting `string` instead of `[]byte` is a good enough reminder about that.

keystore/v2/keystore/filesystem/keyRing.go
```diff
@@ -0,0 +1,255 @@
...
+
+// AllKeys returns all keys of this key ring, from newest to oldest.
+func (r *KeyRing) AllKeys() ([]api.Key, error) {
+	// Return a copy so that the caller cannot modify *our* cache. Also,
+	// return keys in reverse order, from newest to oldest. This makes
+	// AcraStruct decryption attempts more likely to succeed earlier.
+	keyCount := len(r.keys)
+	keys := make([]api.Key, keyCount)
+	for i := range r.keys {
+		keys[keyCount-i-1] = r.keys[i]
```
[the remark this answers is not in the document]
ilammy: Yeah, that’s a good idea.

Initially, I assume that there wouldn’t be many keys in a key ring so fetching all of them if okay. Eventually the keys are going to be destroyed after being rotated so they will be taken off the key ring. (And in current implementation we store entire key rings. If we have 10,000 keys in a single key ring – that’s what we are going to read and write every time we modify any one of them.)

However, we can still add some convenience filtered methods to fetch all keys in pre-active and active state, for example. We’ll just need to observe how `AllKeys()` is actually used, what filters we apply to returned lists, and then move the common code into key store.

keystore/v2/keystore/filesystem/keyStore.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	keystore := &KeyStore{
+		encryptor: cryptosuite,
+		notary:    notary,
+		fs:        backend,
+		log: log.WithFields(log.Fields{
+			"service": serviceName,
+		}),
+	}
+	runtime.SetFinalizer(keystore, (*KeyStore).finalize)
```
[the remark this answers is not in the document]
ilammy: > how you found this method and decided to use?)

Well, I know about garbage collection probably more than I would like to know, so... :)

It’s probably overkill, but it’s a safety measure nevertheless. The motivation was to ensure that we release the file locks on the key store and close the backend connections in case something somewhere panics but we recover (and in case developers forget to write `defer keystore.Close()`).

Actually, `os.File` definitely has a finalizer and various collection handles tend to have them as well (at least indirectly via some `net.TCPConn` instance they wrap), so the OS resources will be released eventually in case they leak. However, this works only for filesystem locks (they are released when the last file descriptor for the lock file is closed). This may not be the case with distributed locks implemented on top of some key-value storage. In this case we will need to perform more work than simply close a socket.

Though, key stores tend to be allocated in `main()` and closed only before exiting the process so we’d close the file descriptor anyway. And in case of distributed locks we’d still need to make them work with timed leases or something to protect against connectivity losses, so they too will be eventually released even if the holder does not do it explicitly.

So this finalizer is here mostly for the peace of mind, and for that sake of that `log.Warn()` in case a key store is not allocated in `main()` _and_ leaked without being closed. (Go does not run pending finalizers when the process exits, so if you leak a key store in main() then you won’t see that warning.)

keystore/v2/keystore/signature/notary.go
ilammy: > why not to accept just `hash.Hash`?

Mostly for extensibility so that we can add a new hash functions without changing signatures and making them confusing. It is possible to have

```go
func NewNotary(sha256 hash.Hash)
```

but then you might accidentally pass a `sha256.New()` there instead of `hmac.New(sha256.New, yourKey)`. When the time comes to add a new function, this changes into

```go
func NewNotary(sha256, keccak hash.Hash)
```

which makes it easy to accidentally swap the functions. Even worse if we just replace it:

```go
func NewNotary(keccak hash.Hash)
```

the signature is the same, old code still compiles, but behaves incorrectly. In every case you need to go and manually hunt down every call to `NewNotary()` in the code base and make sure it’s correct.

_Actually_ it should not be a problem in practice because most likely there would be exactly one call to `NewNotary()` from the only key store we use. But I’d still be consistent with `KeyEncryptor`’s idea to keep _all_ cryptography behind a interface. Though, in case of HMAC-SHA-256 it’s kind of a leaky abstraction...

Okay! I can’t say you instantly convinced me, but you got me thinking. Replacing this with just `hash.Hash` keyed by the caller should probably allow us to simply reuse the existing `KeyEncryptor` interface and avoid inventing new ones. I’m still somewhat worried about accidentally using the wrong function. And about having to write `hmac.New()` manually instead of having some `NewSCellEncryptorFromEnvironment()` to sort everything out for you. But I’ll give it a thought.

keystore/v2/keystore/api/key.go
```diff
@@ -0,0 +1,145 @@
...
+// MutableKey in a KeyRing.
+type MutableKey interface {
+	Key
+
+	// SetState changes key State to the given one, if allowed.
+	SetState(newState KeyState) error
+
+	// SetCurrent makes this key current in its key ring.
+	// Does nothing if the key is already current.
+	SetCurrent() error
```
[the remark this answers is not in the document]
ilammy: > as I understand, it requires from MutableKey implementation to have pointer to KeyRing implementation to be able to setup itself as current key, yes?

Yeah, you’re right. A key keeps a pointer to its parent key ring as well as its own sequential number so that it’s possible to tell which key it is. This is used not only to set the current key, but to request key data as well because it’s actually stored in the key ring, not in the key.

I don’t really like this indirection. (And it used to be even more silly, like the key keeping an actual _pointer_ into data stored inside the key ring — which had to be refreshed in all existing keys when a new key is added to the ring.) However, my attempts at removing it ended up in just abolishing the whole `Key` interface altogether and interacting with keys via the key ring. Maybe that would be a better, like

```go
type KeyRing interface {
        CurrentKey() (int, error)
        SetCurrentKey(seqnum int) error

        AddKey(key KeyDescription) (int, error)

        PublicKey(seqnum int, format KeyFormat) ([]byte, error)
        PrivateKey(seqnum int, format KeyFormat) ([]byte, error)
}
```

I’m not sure it’s good, but I’m not sure we’d need a separate `Key` interface either.

Well, now that you have seen the code, let’s discuss it offline.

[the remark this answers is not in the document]
ilammy: @Lagovas, I’ve coded a simplified API like shown above which does not rely on a separate `Key` instances to access keys in. All key accesses are performed via a `KeyRing` object. Here’s a [diff against this PR](https://github.com/ilammy/acra/compare/keystore-v2...simple-key-api) (changes only), and a [complete diff](https://github.com/cossacklabs/acra/compare/keystore-v2...ilammy:simple-key-api) (easier to read).

Could you please give it a look? I’ll include it into this PR if that’s fine by you.

(I’ve simply renamed the interfaces and fixed up the code so that it’s clear what changes are made. All those awkward `KeyAccess` interfaces are unlikely to be useful on their own so they can be merged into `KeyRing` directly.)

ilammy:
@storojs72,

> 1. A single key for some security functionality that Acra provides (encryption, search, secure log) has its state (unique identifier, usage) and cryptoperiod associated with it.

Correct. An individual key is used only for single purpose (as prescribed by its key ring). A key can be identified by the name of the key ring (unique within the key store) and the sequential number of the key in its key ring (unique within a key ring). Each key contains its current state and individual cryptoperiod.

Key store data model does not guarantee that key content is unique. It’s possible to have the same key content to be stored as several “keys” within different or same key ring. We can add some simple checks later, but the data model cannot make it easier than looking through the entire data for duplicates.

> 2. Key ring contains a set of single keys related to one security functionality (e.g. key ring for encryption can't contain key for encrypted search).

Right. Each key ring is intended to be used for one purpose only. Storage keys will have their own key ring, search keys will be stored in a different one.

> 3. Key directory contains a set of key rings associated with concrete client (application) and can be identified by client id + zone id.

Yep. Key rings can be partitioned by client and zone in their name. For example, `myApp/zone_c2xhc2hlcwo/storage`. It (will be) possible to cheaply enumerate all key rings under `myApp/*` if the need arises. They are also stored in a separate filesystem directory.

> 4. Key store is a high level entity in the hierarchy. It contains multiply key directories.

Yes. You can think of it as a “root key directory”.

> And now we use encryption on single key level and signature on key ring level. Is my understanding correct?

Yes. We encrypt only data of individual keys (actual secret material). Metadata (names, seqnums, cryptoperiods, etc.) is not encrypted and stored as plaintext in the key ring. Complete key ring is signed to prevent tampering. Encryption and signature use separate keys. All keys in the key store are encrypted with the same key, but with different and unique contexts. All key rings in the store are signed with the same key, but with different and unique contexts as well.

vixentael:
>  we need to update key ring data atomically and prevent possible data loss, either due to power failures, application failures, or concurrent modification.

I appreciate your attitude :)

keystore/v2/keystore/api/key.go
```diff
@@ -0,0 +1,145 @@
...
+	// PublicKey data in given format, if available.
+	PublicKey(format KeyFormat) ([]byte, error)
+	// PrivateKey data in given format, if available.
+	PrivateKey(format KeyFormat) ([]byte, error)
+	// SymmetricKey data in given format, if available.
+	SymmetricKey(format KeyFormat) ([]byte, error)
+}
+
+// MutableKey in a KeyRing.
+type MutableKey interface {
```
Lagovas: looked one more time and can't imagine situation when key's state should be changed by itself. always it is the responsibility of keyring. when new key added, keyring take new key and set state pre-active or active (depends on ValidSince value. If key is compromised, our some cli tool will accept some key that should be marked as compromised and passed to keyring/keystore to mark. Keyring set new state (and probably place key at end of keychain). If keystore process GetKey request and see that new key from keychain is outdated, it set status Deactivate, place at end of keychain and return another valid key.

Plus looked on our key statuses. I think that good to have only statuses that can be set manually after some event.
We don't need pre-active/active/deactivated statuses because they depends on time period of key (valid since, valid 
 until) and should be configured automatically by keyring/keystorage.

keystore/v2/keystore/filesystem/keyRing.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// Don't start with zero to not match zero-initialized defaults.
+const firstSeqnum = 1
+
+func (r *KeyRing) nextSeqnum() int {
+	if len(r.data.Keys) == 0 {
+		return firstSeqnum
+	}
+	return r.data.Keys[len(r.data.Keys)-1].Seqnum + 1
```
Lagovas: the newest key at the end? what if we will sort by some other logic our keys (new first, destroyed last) and their numbers will not sorted by ascending? what about to cache max(r.data.Keys.Seqnum...) value on load keyring from storage and use it? or if we will use not array struct but some linked list or tree. then retrieve last key wiill be not such easy

keystore/v2/keystore/asn1/asn1.go
```diff
@@ -0,0 +1,260 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if len(rest) != 0 {
+		return nil, ErrExtraData
+	}
+	return container, nil
+}
+
+// KeyRingReference to a child key ring, not included into KeyDirectory object directly.
```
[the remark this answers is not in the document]
ilammy: > can you explain for what and how will be used such references?

It’s a placeholder for currently unimplemented “key directories”. The idea is that each directory will contain a list of `KeyRingReference` and `KeyDirectoryReference` for its direct children. These structures contain names and signatures of objects, so signing a key directory signs its children content too:

```
KeyDirectory {
    name: "ClientAlice"
    rings: [
        KeyRingReference {
            name: "ClientAlice/storage",
            signature: "12345"
        }
    ],
    signature: "67890",
)
```

This should allow us to verify integrity of the key store as a whole, not only of individual key rings. Currently it is possible to remove a keyring entirely, or replace it with any previously signed version without key store noticing it. Key directories will prevent that because they will track what key rings they contain and what data they store.

keystore/v2/keystore/filesystem/backend/utils.go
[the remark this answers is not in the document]
Lagovas: can you comment it in the source code? because I am sure that I will forget your explanation after some time) and for other future contributors, to prevent removing it)

keystore/v2/keystore/api/key.go
ilammy: > looked one more time and can't imagine situation when key's state should be changed by itself. always it is the responsibility of keyring

Alrighty then. I’ll clean up and add to this PR that change which moves all key access API into the key ring directly.

> We don't need pre-active/active/deactivated statuses because they depends on time period of key (valid since, valid until) and should be configured automatically by keyring/keystorage.

Hm... I don’t think that validity range coincides with usage timespan. The key should not be used outside of validity range (either before or after it), but it can be active for a shorter period, as defined by some usage policy.

For example, some keys may be generated in advance and placed into _pre-active_ state, with validity range starting on 2020-04-01 and ending 2021-03-31. However, something comes up in the infrastructure and it is not possible to start using keys on 2020-04-01. Because of ${legal_reasons}, IDK. Similarly, it might be necessary to deactivate a key earlier that its scheduled validity end because a new policy says that keys should not be valid for more than _half_ a year, not a whole year as before. These cases call for explicit key activation and deactivation by administrator, not implicit activation by the key store itself.

The key store provide a mechanism (tracking key states), it’s up to users to decide on the policy of the state updates.

(That being said, it’s probably reasonable to deactivate a key or otherwise complicate its usage past the validity range. Though again I wonder whether this should be enforced by elsewhere or by the key store directly.)

keystore/v2/keystore/signature/notary.go
Lagovas: actually I dont like that we accept interface with hardcoded name of algorithm: hmac.
I think will be better if we will name interface like "Signer" or "SignAlgorithm" (or any other which will be abstract) and which will force manually implement some wrapper and disallow to use standard insecure hash functions. for example:
```
type Signer interface {
  hash.Hash
  Signer()
}
type KeyedHmac struct {
  hash.Hash
}
func NewKeyedHmac(key []byte)(*KeyedHmac, error){
  return KeyedHmac{hmac.New(sha256.New(), key)), nil
}
// Signer implement interface Signer
func (h *KeyedHmac) Signer(){}
```

Here we accept our own interface, disallow to pass something like sha256.New and force to manually write some wrapper and implement `Signer()` method to pass as object

keystore/v2/keystore/filesystem/keyRing.go
ilammy: > what if we will sort by some other logic our keys (new first, destroyed last) and their numbers will not sorted by ascending? what about to cache max(r.data.Keys.Seqnum...) value on load keyring from storage and use it?

Good point! Yes, current implementation assumes that the keys are stored with ascending sequence numbers. This assumption breaks if we start using a different order.

It’s a good idea to just store the last seqnum in key ring directly. They should not be reused. Currently we do not implement key removal, but imagine that we have this:

```
[1, 2, 3, 4, 5]
```

then the key number 5 gets removed. If we don’t record that the next seqnum should be 6, we will reuse 5 and much confusion will arise (say, because someone tracks hashes of keys tied to their seqnums).

keystore/v2/keystore/api/key.go
Lagovas: Okay, status processing and managing we will see when will integrate it into acra. for now more generalized better than more limited

keystore/v2/keystore/filesystem/keyRing.go
Lagovas: I think that it good requirement and assumption that we will not physically remove any key. We can change state to "destroyed", store empty key material (empty or zeroized array) but leave metadata as is (time period, type, ...) for history and any audit requirements

ilammy: Yeah, that was the idea behind not allowing to remove keys from a key ring: mark them as destroyed instead. Though, I could imagine some pathological use cases where a key is frequently rotated and we end up with 1,000,000 useless destroyed keys. It would be nice to allow to prune metadata for older keys. That is possible to do without disturbing the currently active key (and hence the next seqnum). So as long as we don’t allow to remove the latest key, we don’t need to store the last used seqnum explicitly.

ilammy: > or if we will use not array struct but some linked list or tree. then retrieve last key wiill be not such easy

ASN.1 does not support linked lists directly and here `r.data.Keys` refers to parsed ASN.1. So if we assume that the keys are sorted by seqnum and active keys go last then we can get away with this implementation.

It allows some leeway, such as to put “unusable” keys first, followed by active ones, grouped by their state. For example, we can store keys like this:

```
Index | Seqnum | State      
------+--------+------------
    0 |      4 | compromised
    1 |      7 | compromised
    2 |      1 | deactivated
    3 |      2 | deactivated
    4 |      3 | deactivated
    5 |      5 | active     
    6 |      6 | active     
    7 |      8 | pre-active 
    8 |      9 | pre-active 
```

This makes it cheap to add new keys and relatively cheap to filter for “active” keys. You can get cheap addition by storing last seqnum separately, but if the keys are sorted by seqnum alone you still need to look through all of them to filter by state.

keystore/v2/keystore/signature/notary.go
ilammy: Hm... I like this approach. I wonder if it’s possible to make it even more simple.

As I asked myself a question “why not?” I remembered that I could not think up a suitable generic interface. The issue is that `Signature` object stores _the algorithm_ too as ASN.1 Object Identifier, along with signature value. I was thinking about something even more generic:

```go
type Signer interface {
        Sign(data, context []byte) []Signature
        Verify(data, context []byte, signatures []Signature) error
}
```

But in this case we will need to 1) store the algorithm in `Signature`:

```go
type Signature struct {
        Algorithm SignatureAlgorithm
        Value []byte
}

type SignatureAlgorithm int
const (
    SignatureSha256 SignatureAlgorithm = iota
)
```

2) map SignatureSha256 onto ASN.1 OID, 3) map ASN.1 OIDs back into `SignatureAlgorithm ` and construct a temporary `[]Signature`. I though it was too much of a hassle if the interface is still explicit about using SHA-256 (and thus does not really hide this detail).

P.S. I named it “notary” because I constantly confuse “singer” with “signer”.

keystore/v2/keystore/api/key.go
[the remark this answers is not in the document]
ilammy: I’ve pushed the commit which moves all key access API into `KeyRing` interface.

keystore/v2/keystore/signature/notary.go
ilammy: @Lagovas, after more and more thinking, it seems I've found a signature interface that should fill the bill. Please give a look to recently pushed changes.

We put cryptography behind the interface for two main reasons:

- make it possible to easily swap algorithms if necessary
- hide direct access to encryption/signature keys

Original interfaces succeeded in the second goal but not so much in the first one. That code duplicated existing Secure Cell encryptor code and leaked the implementation detail of using SHA-256. In the end, in order to add a new signature algorithm we would have to update a whole lot of interfaces, etc. it's hardly _easily_ extensible.

New interface consists of `signature.Algorithm` which defines abstract signing algorithm. It is used by `signature.Notary` to sign ASN.1 objects and verify signatures on them. You can pass a list of signatures for the notary to use. Now we can easily add new algorithms by implementing `signature.Algorithm` and adding it to the list.

With this we no longer need to duplicate `KeyEncryptor` interface and Secure Cell encryption. We can reuse existing implementations. We also reexport existing interface for convenience.

`crypto.SignSha256` provides default implementation of signature algorithm: HMAC-SHA-256. As a bonus, this interface allows us to avoid keeping the signature key in memory (we store only keyed `hash.Hash`).

We also keep convenience utilities: `crypto.KeyStoreSuite` (which is now a struct) to avoid passing tons of arguments to key store constuctors as well as `NewSCellSuite()` function which constructs the default cryptography suite: encrypt with Secure Cell, sign with HMAC-SHA-256.

keystore/v2/keystore/crypto/signature.go
```diff
@@ -0,0 +1,57 @@
...
+// AlgorithmOID returns ASN.1 OID for this algorithm.
+func (s *SignSha256) AlgorithmOID() encodingASN1.ObjectIdentifier {
+	return asn1.Sha256OID
+}
+
+// Sign provided data in given context.
+func (s *SignSha256) Sign(data, context []byte) []byte {
+	s.hmac.Reset()
+	s.hmac.Write(context)
+	s.hmac.Write(data)
```
Lagovas: nice to use some extra value to hash between hashing 'context' and 'data' to separate their values. because then hashing `data='hello ' context='world'` == `data='h' context='ello world'` - https://play.golang.org/p/bjaphNK_9Nu

keystore/v2/keystore/signature/notary.go
```diff
@@ -103,11 +110,11 @@ func (s *Notary) Verify(data, context []byte) (*asn1.VerifiedContainer, error) {
 }
 
 func (s *Notary) signData(data, context []byte) []asn1.Signature {
-	signatures := make([]asn1.Signature, 0, 1)
-	signatures = append(signatures, asn1.Signature{
-		Algorithm: asn1.Sha256OID,
-		Signature: s.signSHA256(data, context),
-	})
+	signatures := make([]asn1.Signature, len(s.algorithms))
+	for i, algo := range s.algorithms {
```
Lagovas: can you recall when and for what we decided to use several signatures with different algos for same data?

keystore/v2/keystore/crypto/signature.go
ilammy: Normally we use this with formatted context and data so it’s unlikely that we accidentally run into a collision this way. But fair point :+1: I’ll add a separator.

keystore/v2/keystore/signature/notary.go
ilammy: > can you recall when and for what we decided to use several signatures with different algos for same data?

It’s mostly for future-proofing.

Indeed, in the nearest foreseeable future we are going to use HMAC-SHA-256 exclusively. Supporting more than one possible algorithm allows us to add new algorithms and phase out old ones without it being a breaking change requiring a flag day when upgrading Acra installation.

With this we can simply add a new algorithm and it will be gradually used after update. We can remove an outdated algorithm and existing signatures will simply be ignored. We do not have to develop any conversion utilities for resigning keystore with new algorithm.

But yeah, there is no immediate short-term benefit in it. This comes into play when SHA-256 will be declared not strong enough anymore.

Lagovas: okay, to avoid breaking changes and re-signing logs, we can try to verify with all supported algorithms. but what the benefit from signing with different algos instead just to use last/new added?

ilammy: > but what the benefit from signing with different algos instead just to use last/new added?

It somewhat increases resistance to collisions. Even if individual algorithms are easy to crack, it’s much harder to find a value which results in a simultaneous collision of multiple hash functions based on different computation principles. For example, if we add BLAKE3 along with SHA-256, then even if SHA-256 is cracked tomorrow, user key stores will be more or less shielded from that because existing key rings will have BLAKE3 signatures on them, which is radically different algorithm supposedly immune to collisions in SHA-2.

There isn’t much benefit in this if we update the current signature algorithm in a timely manner and these updates are distributed and applied to production environment in a timely manner. However, it’s easier said than done. Supporting multiple algorithms allows us to do it smoother and ahead of time.

ilammy:
> interestingly to see how it will be used and implemented as AS/AT keystore backend)

Here's [a preview](https://github.com/ilammy/acra/commit/8e95ecb0c3f2ffb1a3c813624e2e3f7aeb04e123) of what's currently cooking. It looks like _a lot_ of repetitive code. I hope to reduce duplication and prune unused methods when it can be used as a drop-in replacement for all current use cases.

Lagovas:
looks nice)

PR #354, opened 2020-03-11
Use keystore v2

ilammy:
Teach all Acra CE tools to store and retrieve their keys in Key Store v2.

The tools will use Key Store v2 only when `--keystore=v2` option is passed on the command-line. By default, Key Store v1 is used. At some point in the future v2 will become the new default. You can explicitly request v1 with `--keystore=v1`.

Key Store v2 uses _two_ master keys passed via environment variables:

  - `ACRA_MASTER_ENCRYPTION_KEY` — used to encrypt private and symmetric keys
  - `ACRA_MASTER_SIGNATURE_KEY` — used to sign key ring content

`ACRA_MASTER_KEY` variable is ignored by Key Store v2.

You can use **acra-keymaker** tool to generate suitable keys, just don't forget to put them into separate files.

### Testing

I have manually tested AcraServer, AcraConnector, AcraTranslator with Key Store v2 and they appear to work fine. Further testing of other tools is pending but I guess they'll be fine too.

Since this feature is behind a feature flag, the branch may be even merged into master from this point. However, I'd like to stabilize and test it more before that.

I have tried tricking the automated integration tests to use Key Store v2 but it's not that easy. Some issues that need to be resolved before it can be tested automatically:

- We need to pass `--keystore=v2` to enable it. I've made this a command-line argument so that this experimental feature cannot be accidentally enabled by forgetting to unset some environment variables. However, it also interferes with test harness which now needs to be taught to support Key Store v2.
- A lot of testing tools expect to be able to read keys directly from the key store (see [`test/utils.py`](https://github.com/cossacklabs/acra/blob/2b88682b775e171e218088e99725d349e35df443/tests/utils.py#L107)). This isn't that simple with v2 now. I guess the best way would be to remove those hacks from Python code and provide some utility to access Key Store using the same Go code.
- It's still to be defined how broad the testing matrix should be. Should it be enough to test some cases with v1 and v2? All of them? Should we test v1—v2 integration as well?

Anyhow, automated testing of key store v2 is out of scope for this PR.

### Implementation notes

Like Key Store v1, `keystore` package exports three basic key stores:

  - `ServerKeyStore` — used almost everywhere
  - `ConnectorKeyStore` — used by AcraConnector, differs in TransportKeyStore implementation
  - `TranslatorKeyStore` — used by AcraTranslator, differs in TransportKeyStore implementation

The implementation is spread over multiple files which makes it easier to track what keys are used where. The transport key usage is particularly convoluted:

| Method           | ServerKeyStore | ConnectorKeyStore | TranslatorKeyStore |
| ---------------- | -------------- | ----------------- | ------------------ |
| GetPrivateKey    | `server`       | `connector`       | `translator`       |
| GetPeerPublicKey | `connector`    | `server`/`translator` <sup>*</sup> | `connector`  |

<sup>*</sup> Depends on the _mode_ in which AcraConnector is running.

Other key types are more or less straightforward.

Here's also the layout of key store in the filesystem:

```
.acrakeys
├── client              client-specific keys, partitioned by ID
│   ├── Alice
│   │   ├── storage.keyring     storage key pair
│   │   └── transport           transport key pairs
│   │       ├── connector.keyring
│   │       ├── translator.keyring
│   │       └── server.keyring
│   └── Bob
│       └── ...
├── zone                zone-specific keys, partitioned by ID
│   ├── AAAAAA
│   │   └── storage.keyring     storage key pair
│   └── BBBBBB
│       └── ...
├── authentication.keyring      WebConfig access symmetric key
└── poison-record.keyring       poison record key pair
```

For now, we put keys into those key rings and do not touch them after that. The keys always remain in the _pre-active_ state. The keys also have a default cryptoperiod of 365 days which is stored, but ignored. You can rotate the keys without reencrypting data by running acra-keymaker again, but the tools currently ignore historical keys and use only the current one.

### DIFF
--- a/keystore/v2/keystore/filesystem/backend/filesystem.go
+++ b/keystore/v2/keystore/filesystem/backend/filesystem.go
@@ -36,12 +36,14 @@ const directorySubsystemName = "directory-backend"
 const (
 	keyDirPerm  = os.FileMode(0700)
 	keyFilePerm = os.FileMode(0600)
+	versionPerm = os.FileMode(0644)
 )
 
 // Errors returned by DirectoryBackend:
 var (
 	ErrNotDirectory       = errors.New("KeyStore: root directory is not a directory")
 	ErrInvalidPermissions = errors.New("KeyStore: invalid access permissions")
+	ErrInvalidVersion     = errors.New("KeyStore: invalid version file content")
 )
 
 // DirectoryBackend keeps data in filesystem directory hierarchy.
@@ -51,7 +53,11 @@ type DirectoryBackend struct {
 	lock *os.File
 }
 
-const lockFile = ".lock"
+const (
+	lockFile      = ".lock"
+	versionFile   = "version"
+	versionString = "Acra Key Store v2"
+)
 
 // CreateDirectoryBackend opens a directory backend at given root path.
 // The root directory will be created if it does not exist.
@@ -92,6 +98,11 @@ func CreateDirectoryBackend(root string) (*DirectoryBackend, error) {
 			errLog.WithError(err).Debug("failed to stat root key directory")
 			return nil, err
 		}
+	}
+	err = checkVersionFile(root)
+	if err != nil {
+		errLog.WithError(err).Debug("failed to create version file")
+		return nil, err
 	}
 	lock, err := os.Create(filepath.Join(root, lockFile))
 	if err != nil {
@@ -138,6 +149,11 @@ func OpenDirectoryBackend(root string) (*DirectoryBackend, error) {
 			Debugf("invalid access permissions on root key directory")
 		return nil, ErrInvalidPermissions
 	}
+	err = CheckDirectoryVersion(root)
+	if err != nil {
+		errLog.WithError(err).Debug("not a key store")
+		return nil, err
+	}
 	lock, err := os.Create(filepath.Join(root, lockFile))
 	if err != nil {
 		errLog.WithError(err).Debug("failed to create lock file")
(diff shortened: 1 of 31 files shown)


Lagovas:
great job, looks nice. cool that not so much changes in binaries to add new keystore version.

> A lot of testing tools expect to be able to read keys directly from the key store (see test/utils.py). This isn't that simple with v2 now. I guess the best way would be to remove those hacks from Python code and provide some utility to access Key Store using the same Go code.
yeah, it's really problem for testing v2. imho the simplest method will be to use some tool which will convert keystore v2 -> v1 or v1 -> v2 (anyway this we will need for migration existing customers). then we can generate keys by one method and converto to new format.


> It's still to be defined how broad the testing matrix should be. Should it be enough to test some cases with v1 and v2? All of them? Should we test v1—v2 integration as well?

imho if we will have some tools which can convert one keys to other then simplest method will be add some global switcher which keystore use and re-run whole integration test with second version of keystore. because 90% of testcases in integration tests uses keystores. and we should be sure that it fully works with both. To do it we need to add switching by env variable (easiest way imho) keystore version in next methods:
* [_fork_acra](https://github.com/cossacklabs/acra/blob/master/tests/test.py#L931)
* [fork_connector](https://github.com/cossacklabs/acra/blob/master/tests/test.py#L844)
* [fork_translator](https://github.com/cossacklabs/acra/blob/master/tests/test.py#L997)
And extend separate tests for cli tools which uses keystore too, such keymaker/add-zone/generatepoisonrecord

ilammy:
I have pushed a couple commits that improve key store versioning story. Now filesystem-based key stores will keep a `version` file in them which stores a marker value that can be used to identify key store content and version. With this we are able to accurately tell v1 and v2 apart. This means that for _existing_ key stores there should be no need to use `--keystore` option by most of the tools. It may only be necessary during initial key store creation with acra-keymaker or future export/import tool.

ilammy:
Re: automated testing.

After some offline discussion we have come to a conclusion that the following tests are necessary:

- full integration test suite with v1 (that's what we have now)
- full integration test suite with v2
- verify v1 ⟹ v2 migration

So in any way we will need to teach the test runner what key store they should use. This looks like a command-line switch that needs to be applied on start. It will define `--keystore` version passed to keymaker as well as the environment variables that need to be set. With that integration tests should pick it up automatically.

Migration tests should be possible to write once we have the migration tool which is currently work in progress. We could save some time by running v1 tests first, then the migration tool, then v2 tests with existing converted key store.

I will look into automated testing once the migration tool is out.

keystore/v2/keystore/filesystem/backend/filesystem.go
```diff
@@ -150,6 +166,63 @@ func OpenDirectoryBackend(root string) (*DirectoryBackend, error) {
...
+}
+
+func createVersionFile(rootDir string) error {
+	path := versionFilePath(rootDir)
+	// Make sure the file does not exist and create it with proper mode.
+	file, err := os.OpenFile(path, os.O_CREATE|os.O_EXCL|os.O_WRONLY, versionPerm)
+	if err != nil {
+		return err
+	}
+	// Close() might fail for newly written files, make sure we don't lose this error.
```
Lagovas: did you check that it works like this? https://play.golang.org/p/yNPWMyA8TAt - checked, it works only if we re-assign value for return variable

ilammy: Hm... you’re right. It does not work as expected. (And I did not test it properly, shame on me.)

I saw this pattern somewhere on the web, or reading through the standard library, or something. Maybe it *did* use return variables there.

Thank you for attention and correction. I’ll review defers to see if this is incorrectly used anywhere else...

ilammy: I've looked through the code and it seems this is the only place which is incorrect. Fun thing, I *did* use this pattern correctly elsewhere:

- https://github.com/cossacklabs/acra/blob/e22cacaf4f03a9ec475fee76af330384556fb1d2/keystore/v2/keystore/filesystem/keyStoreLoad.go#L30-L52

I guess it's just a brain fart this time.

I've fixed the code to use return variables.

PR #355, opened 2020-03-16
sync with cop branch and ee code

Lagovas:
* remove `id` arg from connection wrapper as redundant. add new function to create wrapper for client-side code with defining serverID once. implement methods to allow use secure session as gRPC transport
* tls for acra-translator
* added helpers to tls package `RegisterTLSBaseArgs`/`RegisterTLSClientArgs` to reuse in acra-ee
* improve logging in acra-translator, log clientid/zoneid + start/end of operation (added to allow manually calculate time of execution)
* decrypt rotated acrastructs in acra-translator
* use new keystore with storage component
* decryptor refactoring, sync with ee version
* make public some functions from keystore package (GetZoneKeyFilename)
* added function ZeroizeKeys used to zeroizing keys for rotated keys


README.md
```diff
@@ -344,15 +344,9 @@ Requirements: Linux or macOS with installed Docker.
 | 🔛 [Run Acra Load Balancing Example](https://github.com/cossacklabs/acra-engineering-demo/#example-8-load-balancing-with-acra) 🔛 |
 |---|
 
-## GDPR, HIPAA, CCPA
+## GDPR and HIPAA
```
ilammy: Looks like a merge conflict. The version in master is the correct one, @karenswry has recently expanded this section.

cmd/acra-translator/server/server.go
```diff
@@ -249,30 +250,51 @@ func (server *ReaderServer) Start(parentContext context.Context) {
...
+				}
+				//listener = tls.NewListener(listener, server.config.GetTLSConfig())
+				opts = append(opts, grpc.Creds(credentials.NewTLS(server.config.GetTLSConfig())))
+			} else {
+				//listener, err = network.NewSecureSessionListener(server.config.ServerID(), server.config.IncomingConnectionGRPCString(), server.keystorage)
+				//if err != nil {
+				//	grpcLogger.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorTranslatorCantHandleGRPCConnection).
+				//		Errorln("Can't create secure session listener")
+				//	return
+				//}
```
ilammy: I guess we don't need this anymore. (And line 268 too.)

keystore/filesystem/translator_keystore.go
```diff
@@ -62,13 +116,3 @@ func (store *TranslatorFileSystemKeyStore) GetPrivateKey(id []byte) (*keys.Priva
...
-
-// GetPeerPublicKey returns other party transport public key.
-func (store *TranslatorFileSystemKeyStore) GetPeerPublicKey(id []byte) (*keys.PublicKey, error) {
-	filename := getConnectorKeyFilename(id)
-	key, err := ioutil.ReadFile(filepath.Join(store.directory, getPublicKeyFilename([]byte(filename))))
-	if err != nil {
-		return nil, err
-	}
-	return &keys.PublicKey{Value: key}, nil
-}
```
ilammy: Mmm... We can remove this because Acra Translator talks to Acra Server via Acra Connector, and thus uses its peer public key. Acra Server also expects that key for inbound connections. `GetPeerPublicKey()` in the base `KeyStore` is implemented like that. Right?

keystore/keystore.go
```diff
@@ -189,3 +189,13 @@ type KeyStore interface {
...
+
+// MultiKeyStore is a KeyStore with additional interface to retrieve historical versions of the key.
+// It is used to decrypt AcraStructs with possibly rotated keys.
+type MultiKeyStore interface {
+	KeyStore
+
+	GetZonePrivateKeys(id []byte) ([]*keys.PrivateKey, error)
+
+	GetServerDecryptionPrivateKeys(id []byte) ([]*keys.PrivateKey, error)
+}
```
ilammy: **keystore-v2** branch has refactored key access interfaces a bit and that makes this interface unnecessary. These methods can be simply moved into `PrivateKeyStore` interface.

I think it's okay to merge it into master like this to avoid unnecessary changes. I'll sync and clean this up in v2, along with other updates related to multikeys.

network/tls_wrapper.go
```diff
@@ -47,35 +48,30 @@ type TLSConnectionWrapper struct {
...
+	tlsCert       string
+	tlsAuthType   int
+	tlsServerName string
+)
+
+// RegisterTLSBaseArgs register CLI args tls_ca|tls_key|tls_cert|tls_auth which allow to get tls.Config by NewTLSConfigFromBaseArgs function
+func RegisterTLSBaseArgs() {
+	flag.StringVar(&tlsCA, "tls_ca", "", "Path to root certificate which will be used with system root certificates to validate peer's certificate")
+	flag.StringVar(&tlsKey, "tls_key", "", "Path to private key that will be used for TLS connections")
+	flag.StringVar(&tlsCert, "tls_cert", "", "Path to certificate")
```
ilammy: Will TLS still work if these options are not specified? I don't remember the story with systems CA certificates lists with Go. @shadinua has seen some issues with that.

Based on `tls_auth` default value we're defaulting to required mutual TLS authentication. All servers *and* clients have to specify their certificates and keys, unless less strict mode is selected.

network/secure_session_wrapper.go
```diff
@@ -180,23 +182,100 @@ func (err *ConnectionWrapError) Temporary() bool {
...
+	}
+	return wrappedConn, SecureSessionInfo{clientID: clientID, spanContext: spnCtx}, nil
+}
+
+// Info return protocol info for Secure Session
+func (wrapper *SecureSessionConnectionWrapper) Info() credentials.ProtocolInfo {
+	return credentials.ProtocolInfo{
+		ProtocolVersion:  "1.0.0",
+		SecurityProtocol: "secure session",
+		SecurityVersion:  "1.0.0.",
```
ilammy: Where do we verify these versions? 🤔

encryptor/utils.go
```diff
@@ -0,0 +1,184 @@
...
+	Name  string
+	Table string
+	Alias string
+}
+
+var errEmptyTableExprs = errors.New("empty table exprs")
+
+// getFirstTableWithoutAlias search table name from "FROM" expression which has not any alias
+// if more than one table specified without alias then return errNotFoundTable
+func getFirstTableWithoutAlias(fromExpr sqlparser.TableExprs) (string, error) {
```
ilammy: Here and below in this file there is a lot of parsing code with no apparent tests for it.

I figure, is it tested transitively by the integration test suite?

keystore/filesystem/translator_keystore.go
Lagovas: Nice catch. It forced me to try to remember why I removed it and why it works now ))
https://github.com/cossacklabs/acra/blob/master/keystore/filesystem/translator_keystore.go#L28 - we can remove it because keystore for translator inherit this function from base server's keystore and should not be overrided. Because AT and AS expects key from connector like <id>.pub. But we need override `GetPrivateKey` method, because translator has own keys with `_translator` prefix

network/tls_wrapper.go
Lagovas: All logic with TLS initialization declared here - https://github.com/cossacklabs/acra/blob/master/network/tls_wrapper.go#L82
We always load system CA files and append tls_ca if it was defined. Plus we allow to use not strict mode of tls without mutual authentication. Then acra-connector should skip tls_key and tls_cert. But if I remember correctly, they are required for AS/AT if tls should work because server should provide any certificate.
Plus this code is not something new, I just copied all args from AS and moved to tls package to reuse it in acra-ee too.

network/secure_session_wrapper.go
Lagovas: it was added for `credentials.TransportCredentials` interface (gRPC) to use secure session as transport and authentication backend for gRPC and it require to define some info like this. And I just defined simple values like 1.0.0 with prediction that we will not change them in close future

network/tls_wrapper.go
ilammy: Thanks for explanation! 👍

README.md
Lagovas: Actually now I copied README.md from acra/master and wasn't changed - https://raw.githubusercontent.com/cossacklabs/acra/master/README.md
Now master has the same version as in PR.

cmd/acra-translator/server/server.go
Lagovas: yes, forgot to remove

keystore/keystore.go
Lagovas: Okay, left as is

encryptor/utils.go
Lagovas: You know, I looked one more time on this code, found the piece of SQL code which I used for tests when wrote it and decided to add tests too. And I found a case when it worked incorrectly)) Added query and tests for it with fix

ilammy: I know that feel :) You try to fix one thing, pick it up, and there’s a whole bunch of bugs scattering around from under it! :D

I’m glad there’s fewer of them now.

README.md
ilammy: Here’s what I see at **master**’s tip:

- 2b88682b775e171e218088e99725d349e35df443 ("Added CCPA mention to README.md")

And here’s the commit I see as a tip of your branch **lagovas/ee-updates**:

- 38220692fe92c873377a218f342152c10f4ea307 ("add tests for query parsers")

```diff
$ git diff 2b88682b775e171e218088e99725d349e35df443..38220692fe92c873377a218f342152c10f4ea307 -- README.md | head
diff --git a/README.md b/README.md
index a7ecf8c..b03506a 100644
--- a/README.md
+++ b/README.md
@@ -344,15 +344,9 @@ Requirements: Linux or macOS with installed Docker.
 | 🔛 [Run Acra Load Balancing Example](https://github.com/cossacklabs/acra-engineering-demo/#example-8-load-balancing-with-acra) 🔛 |
 |---|
 
-## GDPR, HIPAA, CCPA
+## GDPR and HIPAA
```

So it seems that your master branch has gone stale at some point, then some merges happened, and now git thinks that all conflicts have been resolved (in favor of older version).

Blame & bisect shows that this commit is the culprit:

- 1fd032522e9699a01638ca4bb4b73fc95f75057c ("megrge ce")

Well, I’ll just go ahead and push the commit with a fixup that syncs README with master.

PR #356, opened 2020-03-16
acra-migrate-keys tool

ilammy:
Create a tool for migrating keys between key store formats. Currently it supports only v1 ⟹ v2 migration.

With default settings it is enough to simply run `acra-migrate-keys` to migrate keys from `.acrakeys` (v1) into `.acrakeys.migrated` (v2). After successful migration the administrator has to manually reconfigure Acra services to use the new key store. It is possible to explicitly specify migration details via `--src...` and `--dst...`  options if necessary.

There is a `--dry_run` option which performs migration in memory, without writing results to disk. You can use it to test that everything is in place, the keys can be read and understood.

By default the migration tool will refuse to write output into already existing key store. This is because the migration process is not transactional and idempotent. Migrating the same key store twice will result in duplicate keys, and failure to migrate some keys will still migrate other successful keys. This behavior can be overridden with `--force` option, allowing the migration to proceed.

Note that master keys are expected in environment variables, as usual. You will need to set all variables for successful migration:

  - ACRA_MASTER_KEY
  - ACRA_MASTER_ENCRYPTION_KEY
  - ACRA_MASTER_SIGNATURE_KEY

More implementation details are in commit messages. The code is structured in a way that should easy its reuse in Acra EE with its extended key store support (both v1 and v2).

### DIFF
--- a/cmd/acra-migrate-keys/acra-migrate-keys.go
+++ b/cmd/acra-migrate-keys/acra-migrate-keys.go
@@ -1,0 +1,245 @@
+/*
+ * Copyright 2020, Cossack Labs Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package main is entry point for Acra key migration utility (`acra-migrate-keys`).
+// It is used to convert between key store formats used by Acra.
+//
+// https://docs.cossacklabs.com/pages/documentation-acra/#key-management
+package main
+
+import (
+	"errors"
+	"flag"
+
+	"github.com/cossacklabs/acra/cmd"
+	keystoreV1 "github.com/cossacklabs/acra/keystore"
+	filesystemV1 "github.com/cossacklabs/acra/keystore/filesystem"
+	keystoreV2 "github.com/cossacklabs/acra/keystore/v2/keystore"
+	keystoreApiV2 "github.com/cossacklabs/acra/keystore/v2/keystore/api"
+	filesystemV2 "github.com/cossacklabs/acra/keystore/v2/keystore/filesystem"
+	"github.com/cossacklabs/acra/logging"
+	"github.com/cossacklabs/acra/utils"
+	log "github.com/sirupsen/logrus"
+)
+
+var (
+	defaultConfigPath = utils.GetConfigPathByName("acra-migrate-keys")
+	serviceName       = "acra-migrate-keys"
+
+	defaultSrcDir       = keystoreV1.DefaultKeyDirShort
+	defaultDstDirSuffix = ".migrated"
+)
+
+// CommandLineParams - command-line options.
+type CommandLineParams struct {
+	Src, Dst KeyStoreParams
+	Misc     MiscParams
+}
+
+// KeyStoreParams - key store parameters.
+type KeyStoreParams struct {
+	KeyStoreVersion string
+	KeyDir          string
+	KeyDirPublic    string
+}
+
+// MiscParams - miscellaneous parameters.
+type MiscParams struct {
+	DryRun bool
+	Force  bool
+
+	LogDebug   bool
+	LogVerbose bool
+}
+
+// OpenMode - whether key store is source or destination.
+type OpenMode int
+
+// OpenMode constant values:
+const (
+	OpenSrc OpenMode = iota
+	OpenDst
+)
+
+// RegisterCommandLineParams registers command-line options for parsing.
+func RegisterCommandLineParams() *CommandLineParams {
+	params := &CommandLineParams{}
+	// Source key store
+	flag.StringVar(&params.Src.KeyStoreVersion, "src_keystore", "v1", "key store format to use: v1 (current), v2 (experimental)")
+	flag.StringVar(&params.Src.KeyDir, "src_keys_dir", defaultSrcDir, "path to source key directory")
+	flag.StringVar(&params.Src.KeyDirPublic, "src_keys_dir_public", defaultSrcDir, "path to source key directory for public keys")
+	// Destination key store
+	flag.StringVar(&params.Dst.KeyStoreVersion, "dst_keystore", "v2", "key store format to use: v1 (current), v2 (experimental)")
+	flag.StringVar(&params.Dst.KeyDir, "dst_keys_dir", "", "path to destination key directory (default \".acrakeys.migrated\")")
+	flag.StringVar(&params.Dst.KeyDirPublic, "dst_keys_dir_public", "", "path to destination key directory for public keys (default \".acrakeys.migrated\")")
+	// Miscellaneous
+	flag.BoolVar(&params.Misc.DryRun, "dry_run", false, "try migration without writing to the output key store")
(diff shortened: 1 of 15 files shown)


keystore/v2/keystore/importV1.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	default:
+		log.Debug("unknown key purpose")
+		return ErrUnknownPurpose
+	}
+	return nil
+}
+
+func zeroizeSymmetricKey(key []byte) {
+	utils.FillSlice(0, key)
```
Lagovas: sometimes we will decide to refactor FillSlice to something like that )) or we will not only zeroize but oneize, secondize too ))

tests/test.py
```diff
@@ -4253,6 +4255,7 @@ def testStartupWithoutVersionInConfig(self):
             default_args = {
                 'acra-server': ['-db_host=127.0.0.1'],
                 'acra-connector': ['-user_check_disable', '-acraserver_connection_host=127.0.0.1', '-client_id=keypair1'],
+                'acra-migrate-keys': ['--dry_run', '--src_keydir={}'.format(KEYS_FOLDER.name)],
```
Lagovas: it's cool that you found these place in our 4.4k lines file with tests where you can test starting tool ))

ilammy: Well, no silver bullet here, I just followed the stack trace:

      File "tests/test.py", line 4282, in testStartupWithOutdatedConfigVersion
        stderr = self.getOutputFromProcess(args)
      File "tests/test.py", line 1224, in getOutputFromProcess
        process = subprocess.Popen(args, stderr=subprocess.PIPE)
      File "/usr/lib/python3.5/subprocess.py", line 947, in __init__
        restore_signals, start_new_session)
      File "/usr/lib/python3.5/subprocess.py", line 1551, in _execute_child
        raise child_exception_type(errno_num, err_msg)
    FileNotFoundError: [Errno 2] No such file or directory: './acra-migrate-keys'

If not for that, I'd still be confused about what has gone wrong. So, ugh, thanks for using Python, I guess :)

keystore/v2/keystore/importV1.go
ilammy: Yeah, now that you've mentioned it, this definitely looks like something for `utils`. Should we move these functions there? I remember you've added some in #355:
https://github.com/cossacklabs/acra/pull/355/files#diff-45a2ca38910926a14a2c14c6a04192dd

keystore/filesystem/key_export.go
```diff
@@ -0,0 +1,298 @@
...
+	return utils.ReadFile(key.SymmetricPath)
+}
+
+// DefaultKeyFileClassifier is a KeyFileClassifier for standard key types.
+type DefaultKeyFileClassifier struct{}
+
+// EnumerateExportedKeys prepares a list of keys that can be exported.
+// The keys are classified with default key file classifier.
+func (store *KeyStore) EnumerateExportedKeys() ([]ExportedKey, error) {
+	return store.EnumerateExportedKeysByClass(&DefaultKeyFileClassifier{})
```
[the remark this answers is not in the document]
ilammy: Mmm... It seems like we wouldn't need any state here, and there will be at most two implementations, for CE and EE respectively. However, I'm not sure what do mean with package variables. Could you please elaborate?

I've intended for user code (i.e., migration tool) to only call `EnumerateExportedKeys()`. This other method `EnumerateExportedKeysByClass()` will be used by Acra EE code to implement `EnumerateExportedKeys()` for its basic key store, passing an extended classifier. Migration tool in Acra EE will still use `EnumerateExportedKeys()` (this time for EE key store).

ilammy:
The dependencies have been merged, but the build has been broken. I guess I'll fix it up first (should require only a few tweaks, I hope), and then rebase and retarget this PR onto the correct branch before undrafting it.

cmd/acra-migrate-keys/acra-migrate-keys.go
```diff
@@ -0,0 +1,245 @@
...
+	log.WithFields(log.Fields{"src": *params.Src.KeyStoreVersion, "dst": *params.Dst.KeyStoreVersion}).
+		Fatal("key store conversion not supported")
+}
+
+// MigrateV1toV2 transfers keys from key store v1 to v2.
+func MigrateV1toV2(srcV1 filesystemV1.KeyExport, dstV2 keystoreV2.KeyFileImportV1, params MiscParams) error {
+	log.Trace("enumerating keys for export")
+	keys, err := srcV1.EnumerateExportedKeys()
+	if err != nil {
+		log.WithError(err).Debug("failed to enumerate exported keys")
```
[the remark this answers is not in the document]
ilammy: > But `errors.New` should be defined with lowercase message according to golang standard libraries.

Oh, I just noticed that I've been writing errors like this:


```go
var (
        ErrUnknownPurpose = errors.New("KeyStore: unknown key purpose")
)
```

with a `KeyStore:` prefix.

I guess it is redundant? Internal key store loggers have a `service=keystore` field added to them so we keep this context. If these errors bubble up somewhere else they seem to be not too out-of-context to not require these prefixes.

I think I'll cook a PR for that...

ilammy:
Rebased and retargeted. Changes since draft:

- Updated log lines to start with an uppercase letter to conform with the style.
- Slightly renamed the src/dst options to be consistent with other tools.
- Slightly improved handling of command-line options to avoid dereferences everywhere.
- Added utilities for exporting both encrypted and plaintext symmetric keys (useful for EE).
- Unbroken the build ;)

keystore/filesystem/key_export.go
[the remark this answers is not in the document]
Lagovas: to avoid each time allocation of empty struct (I dont know how golang actually compile it, but can predict that at least some pointer it may allocate on stack/heap), we can allocate it once on package level and use it every time when need (like singleton). For example how it used in flag package - https://golang.org/src/flag/flag.go#L1010
```
var defaultClassifier = &DefaultKeyFileClassifier{};

func (store *KeyStore) EnumerateExportedKeys() ([]ExportedKey, error) {
	return store.EnumerateExportedKeysByClass(defaultClassifier)
}
```

keystore/v2/keystore/importV1.go
Lagovas: I think will be nice to avoid duplication if sometimes we need it one more time

keystore/filesystem/key_export.go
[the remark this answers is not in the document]
ilammy: Oh, I’ve misunderstood your original comment. I thought you meant to use global variables as a way to configure the “classifier” being used by default, with the idea that Acra EE will somehow set a new value for this value, substituting its own classifier.

But you seem to have meant the allocation optimization as above. I’ve updated the code to include it like you’ve written.

keystore/v2/keystore/importV1.go
ilammy: I’ve moved the utilities to `utils`. They will be available for general use once **keystore-v2** lands into master.

Lagovas:
lgtm

ilammy:
Hm... Filesystem key store tests just failed on this push, there was a problem in locking tests (should not be affected by these changes at all). I’ve restarted the build to see if it persists.

I’ll keep it mind, it might be possible that the locking implementation is broken in a subtle way.

cmd/acra-migrate-keys/acra-migrate-keys.go
```diff
@@ -0,0 +1,245 @@
...
+const (
+	OpenSrc OpenMode = iota
+	OpenDst
+)
+
+// RegisterCommandLineParams registers command-line options for parsing.
+func RegisterCommandLineParams() *CommandLineParams {
+	params := &CommandLineParams{}
+	// Source key store
+	flag.StringVar(&params.Src.KeyStoreVersion, "src_keystore", "v1", "key store format to use: v1 (current), v2 (experimental)")
```
vixentael: I'd not name v2 as `experimental`, because when this thing is merged to master, we'll totally forget to rename these into "old" and "new" (or similar)

PR #359, opened 2020-03-18
Log hooks

storojs72:
This PR adds following changes, necessary for audit log functionality (EE version):

1) Added `Builder` for customizing logging (as @ilammy suggested). Now it's done with the following:
```
logging.Customize()
.SetOutput(io.MultiWriter(writer...))
.SetServiceName(serviceName)
.SetFormat(loggingFormat)
.SetHooks(hooks)
.Complete()
```

2) Minor changes in functions naming of `FormattedHook` interface: 
```
PreFormat(entry *log.Entry) error
PostFormat(entry *log.Entry, formatted *bytes.Buffer) error
```

For verification of 1 and 2, please look inside the following commit: https://github.com/cossacklabs/acra/commit/349b9a12cd2901557733a0f19a5ae337dc2c4ae9

3) Added constants for supported format strings (as @Lagovas suggested)


logging/logging.go
```diff
@@ -72,25 +96,76 @@ func GetLogLevel() int {
 	return LogDiscard
 }
 
-// CustomizeLogging changes logging format
-func CustomizeLogging(loggingFormat string, serviceName string) {
-	log.SetOutput(os.Stderr)
-	log.SetFormatter(logFormatterFor(loggingFormat, serviceName))
+var (
+	errNoWriter      = errors.New("output writer is not specified")
```
Lagovas: why these error variables are private?

logging/logging.go
```diff
@@ -72,25 +96,76 @@ func GetLogLevel() int {
...
+// CustomizeLogging changes logging format
+func CustomizeLogging(loggingFormat string, serviceName string) {
+	log.SetOutput(os.Stderr)
+	log.SetFormatter(logFormatterFor(loggingFormat, serviceName, nil))
+	log.Debugf("Changed logging format to %s", loggingFormat)
+}
+
+func logFormatterFor(loggingFormat string, serviceName string, hooks []FormatterHook) log.Formatter {
+	switch strings.ToLower(loggingFormat) {
+	case "json":
```
Lagovas: use new constants?)

logging/logging.go
```diff
@@ -72,25 +96,76 @@ func GetLogLevel() int {
...
+	log.SetOutput(c.writer)
+	log.SetFormatter(logFormatterFor(c.loggingFormat, c.serviceName, c.hooks))
+	log.Debugf("Changed logging format to %s", c.loggingFormat)
+	return nil
+}
 
-	return TextFormatter()
+// CustomizeLogging changes logging format
+func CustomizeLogging(loggingFormat string, serviceName string) {
+	log.SetOutput(os.Stderr)
```
Lagovas: if we add builder, we should use it everywhere to avoid duplication code. Can we change it to `Customize().SetOutput(os.Stderr).SetFormat(loggingFormat).Complete()`? If we do so, we don't need to sync log message `Changed logging format to %s", loggingFormat` in the builder and in this function, sync way how should be defined formatter etc. Plus if we will add some more complicated logic in our builder, we should not to sync it to this function too, just update builder.

logging/log_formatters.go
```diff
@@ -185,3 +216,24 @@ func nanosecondsToMillisecondsString(nanos int64) string {
 func unixTimeWithMilliseconds(e *logrus.Entry) string {
 	return TimeToString(e.Time)
 }
+
+func formatEntry(e *logrus.Entry, formatter log.Formatter, hooks []FormatterHook) ([]byte, error) {
```
Lagovas: cool, one readable simple function which provides so many ways to inject some extra logic)

logging/logging.go
```diff
@@ -72,25 +96,76 @@ func GetLogLevel() int {
...
+func (c *CustomizeBuilder) Complete() error {
+	if c.writer == nil {
+		return errNoWriter
+	}
+	if c.serviceName == "" {
+		return errNoServiceName
 	}
+	if c.loggingFormat == "" {
+		return errNoFormat
+	}
```
ilammy: I think it would be nice to provide some sensible defaults. Like, use plaintext format, write logs to console, do not set the service field, and have no custom hooks. Whatever we do without `Customize()`. That way if one writes

```go
logging.Customize().Complete()
```

then they will get exactly the same default log instead of an unexpected error.

logging/logging.go
storojs72: I looked into @ilammy 's example of keystore builder (https://github.com/cossacklabs/acra/blob/cop/keystore/filesystem/server_keystore.go#L128)

Lagovas: I missed it in his PR *(
@ilammy , why you declared err variables private?)

logging/log_formatters.go
storojs72: @ilammy 's code)))

logging/logging.go
storojs72: Yes. Nice suggestion

logging/logging.go
storojs72: Yes. Agree

logging/logging.go
storojs72: Yes

logging/logging.go
ilammy: Uh...
・・・
\*thinking\*
・・・
I guess that’s just a natural habit of not exporting things that are not needed outside of a module right now and there are no obvious reasons for why they might be needed. I figured that no one is going to write code like

```go
again:
        builder := NewBuilder().ConfigureThis(thing).ConfigureThat(stuff)
        result, err := builder.Build()
        if err != nil {
                if err == ErrAlsoThisNotSpecified {
                        // Oh right, we need to also do this
                        builder.ConfigureAlsoThis(zork)
                        goto again
                }
                // Oh well, no more guesses
                return err
        }
        use(result)
```

so I felt no need to export those errors, because that’s probably the only case where you’d need to have access to those names (you don’t need it to log them, etc.)

Just a natural aversion to expanding the API without a need.

cmd/acra-connector/acra-connector.go
```diff
@@ -255,8 +255,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	// if log format was overridden
-	logging.CustomizeLogging(*loggingFormat, ServiceName)
+	logging.Customize().SetFormat(*loggingFormat).SetServiceName(ServiceName).SetOutput(os.Stderr).Complete()
```
ilammy: We don’t need to add `SetOutput(os.Stderr)` here and in other places because that’s what Customize() does by default now.

logging/logging.go
```diff
@@ -72,25 +95,76 @@ func GetLogLevel() int {
...
+}
 
-	return TextFormatter()
+/*
+// CustomizeLogging changes logging format
+func CustomizeLogging(loggingFormat string, serviceName string) {
+	log.SetOutput(os.Stderr)
+	log.SetFormatter(logFormatterFor(loggingFormat, serviceName, nil))
+	log.Debugf("Changed logging format to %s", loggingFormat)
+}*/
```
ilammy: Please delete the code if it’s not needed anymore, don’t leave it commented out.

If we leave it like this then no one will remember about it when we merge cop into master, and then there will be questions why a merge commit added this.

logging/logging.go
Lagovas: On my opinion, it's okay to not export them if they caught internally and don't go out from the package. But I saw that these errors returned from public methods and caller haven't any possible way to handle it by error type. He just can compare it with `nil` or try to convert to string and rely on string representation of error, and it's not okay.

ilammy: > But I saw that these errors returned from public methods and caller haven't any possible way to handle it by error type.

Point taken 👍

I’ll update existing errors in keystore as I go with v2, exporting them whenever they can leave the package boundaries via exported methods.

cmd/acra-webconfig/acra-webconfig.go
```diff
@@ -389,7 +389,6 @@ func main() {
 	host = flag.String("incoming_connection_host", cmd.DefaultWebConfigHost, "Host for AcraWebconfig HTTP endpoint")
 	port = flag.Int("incoming_connection_port", cmd.DefaultWebConfigPort, "Port for AcraWebconfig HTTP endpoint")
 	loggingFormat := flag.String("logging_format", "plaintext", "Logging format: plaintext, json or CEF")
-	logging.CustomizeLogging(*loggingFormat, ServiceName)
```
vixentael: why this is removed? the next line is a log line, so if logger is not customized, the log line will be output in unknown/default format

vixentael: this line should be here

`logging.Customize().SetFormat(*loggingFormat).SetServiceName(ServiceName).Complete()`

Lagovas:
decided in the chat that acceptable to merge

PR #360, opened 2020-03-19
Add contribution guidelines

ilammy:
Since the number of actual and possible developers is expanding, it’s probably time to describe the contribution guidelines more explicitly. Here's an initial version based on our existing documentation, primed with some friction points encountered during code review so far.

Please feel free to expand it as you see fit in the future.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,121 @@
+# Contributing to Acra
+
+Although Acra is an in-house-developed product by [Cossack Labs Limited](https://www.cossacklabs.com/),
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
...
+
+    ```go
+    // Correct:
+    var ErrCannotConnect = errors.New("cannot connect to AcraService")
+
+    // NOT like this:
+    var ErrCannotConnect = errors.New("AcraService: Cannot connect")
+    ```
+
+  - Log messages should start with an uppercase letter.
```
vixentael: 🏆

PR #361, opened 2020-03-20
Run integration tests with key store v2

ilammy:
This patch series teaches CircleCI to run integration test suite with key store v2.

The test suite now reacts to the `TEST_KEYSTORE` environment variable. If it is not set or set to `v1`, the test suite runs with current key store (v1). If that variable is set to `v2`, the tests run with new experimental key store (v2). This is meant for complete integration test suite run from scratch.

The test suite has been too tightly coupled with key store v1, hardwiring the expected environment variables, key store layout and data format, etc. All of this has been abstracted out where necessary. The tests now read keys using a new **acra-read-key** tool, which currently can read storage keys (by client and zone ID) and the poison record key. It somewhat intersects with key export tool, but works with both v1 and v2 and output keys in plaintext. It is intended only for test usage.

There are also multiple minor tweaks in tests to support key store v2. See commit messages for details.

What is lacking and may be improved later:

- Migration v1 → v2 is not tested right now whatsoever. New tests need to be written for this.
- Tests for key removal are currently disabled. Key store v2 does not support that (yet).
- Tests are running only with PostgreSQL. Other jobs pass with v2, but it's inconvenient to run them.

### DIFF
--- a/cmd/acra-read-key/acra-read-key.go
+++ b/cmd/acra-read-key/acra-read-key.go
@@ -71,7 +71,7 @@ type commandLineParams struct {
 var params commandLineParams
 
 func main() {
-	flag.StringVar(&params.KeyStoreVersion, "keystore", "", "force key store format: v1 (current), v2 (experimental)")
+	flag.StringVar(&params.KeyStoreVersion, "keystore", "", "force key store format: v1 (current), v2 (new)")
 	flag.StringVar(&params.KeyDir, "keys_dir", defaultKeyDir, "path to key directory")
 	flag.StringVar(&params.KeyDirPublic, "keys_dir_public", "", "path to key directory for public keys")
 	flag.StringVar(&params.ClientID, "client_id", "", "client ID for which to retrieve key")
--- a/cmd/acra-read-key/acra-read-key.go
+++ b/cmd/acra-read-key/acra-read-key.go
@@ -105,6 +105,8 @@ func main() {
 
 	var keyBytes []byte
 	switch params.KeyKind {
+	case "":
+		// no key requested, do nothing
 	case keyPoisonPublic:
 		keypair, err := keyStore.GetPoisonKeyPair()
 		if err != nil {
--- a/cmd/acra-read-key/acra-read-key.go
+++ b/cmd/acra-read-key/acra-read-key.go
@@ -1,0 +1,202 @@
+/*
+ * Copyright 2020, Cossack Labs Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package main is entry point for acra-read-key test utility.
+// It is used by integration tests to read arbitrary keys from the key store.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/cossacklabs/acra/cmd"
+	"github.com/cossacklabs/acra/keystore"
+	keystoreV1 "github.com/cossacklabs/acra/keystore"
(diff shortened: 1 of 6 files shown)


cmd/acra-read-key/acra-read-key.go
```diff
@@ -0,0 +1,220 @@
...
+	KeyDirPublic    string
+	ClientID        string
+	ZoneID          string
+	KeyKind         string
+}
+
+var params commandLineParams
+
+func main() {
+	flag.StringVar(&params.KeyStoreVersion, "keystore", "", "force key store format: v1 (current), v2 (experimental)")
```
vixentael:
```suggestion
	flag.StringVar(&params.KeyStoreVersion, "keystore", "", "force key store format: v1 (current), v2 (new)")
```

configs/acra-read-key.yaml
```diff
@@ -0,0 +1,28 @@
...
+# key kind to read, one of: poison-public, poison-private, storage-public, storage-private, zone-public, zone-private
+key: 
+
+# path to key directory
+keys_dir: .acrakeys
+
+# path to key directory for public keys
+keys_dir_public: 
+
+# force key store format: v1 (current), v2 (experimental)
```
vixentael:
```suggestion
# force key store format: v1 (current), v2 (new)
```

ilammy:
> My only thought is that we shouldn't call new format as 'experimental', because it looks like something that will go into master and become a new default setting :)

Well okay. This PR will use “current” and “new”, I'll update other tools to use that wording too.

IMO, there is no big difference. We'd probably need to revisit the command-line help anyway when we switch to v2 by default. For now I intended to communicate that this is a brand new, barely tested, please-don't-use-in-production setting. I do intend to merge it into master as soon as possible – that is, as soon as it has enough automated test coverage. However, this feature is still behind a `--keystore=v2` feature flag that forces users to explicitly opt-in. Who knows when we are going to cut a new release from master, or base some Acra EE work on latest master, maybe at that time key store v2 will still be not tested enough for us to be sure that it's safe for a curious user to enable this flag and not get their entire KMS wiped by accident.

Arguably, if that wiping is even possible, we'd better keep the feature in a branch, but long-running branches are harder to maintain.

tests/test.py
```diff
@@ -1091,47 +1119,68 @@ def tearDown(self):
...
+            log_entry['zone_id'] = zone_id
+
+        if poison_key:
+            public_key = read_poison_public_key(KEYS_FOLDER.name)
+            private_key = read_poison_private_key(KEYS_FOLDER.name)
+            log_entry['public_key'] = b64encode(public_key).decode('ascii')
+            log_entry['private_key'] = b64encode(private_key).decode('ascii')
+            log_entry['poison_record'] = b64encode(get_poison_record()).decode('ascii')
+
+        logging.debug("test log: {}".format(json.dumps(log_entry)))
```
Lagovas: just interesting. something broke when you found this function and it needed refactoring or you had a need to see some log and found place where it produced?)

tests/test.py
```diff
@@ -1636,6 +1689,9 @@ def testClosingConnectionsWithDB(self):
 class TestKeyNonExistence(BaseTestCase):
     def setUp(self):
         self.checkSkip()
+        # FIXME(ilammy, 2020-03-20): implement key removal for v2
```
Lagovas: add please "TODO" abbr, to collect it by IDE (jetbrains' ide have tab TODO which aggregate all todos from all files. maybe some other tools have same feature like grep by "todo" word)

ilammy: PyCharm shows FIXMEs in the TODO list too.

My idea is that TODO is something that you have a moral right to ignore, like a use case that is not useful right now but might become important later. Or some really-really rare edge case. So if you see a TODO in code you don't go all “why didn't you implement it then?”

While FIXME is something that's actually broken and, well, needs a fix. It's probably not *really* important bug that may be ignore right now. But you *should* become angry at the person who left a fixme instead of fixing their stuff.

Anyhow, I don't care particularly so I can rename it if you insist.

tests/test.py
ilammy: Yeah, this function used to read key files directly from filesystem which does not work with v2. It's call sites relied on `acra_key_name` argument being treated as a file name too (e.g., to get access to the poison record key). So I've refactored it a bit to handle all those cases.

I've looked at its output, but it was not needed for debugging this PR.

tests/test.py
Lagovas: if pycharm show fixmes too than ok. plus I thought that we can "todo" keyword to description like `# FIXME(ilammy, 2020-03-20): todo: ` if it wasn't match by IDE.

PR #362, opened 2020-03-23
Call key store v2 "new", not experimental

ilammy:
In order to avoid unnecessary connotations, use a different wording in command-line help.


ilammy:
> this is the best Acra PR ever

Time to switch Acra to obviously superior Windows line endings.

brb, cooking PR

PR #363, opened 2020-03-24
Test key store migration

ilammy:
Add a simple test to verify v1 → v2 key store migration path. Since we'll be changing the key store layout, we cannot use global shared key store and have to reinvent some wheels. Let's at least make it easy to write new tests if necessary.

The tests verify that:

  - we are able to put and get data before migration
  - the migration itself completes successfully
  - the migration keeps the data unchanged and accessible
  - new data can be added after migration

We check both default and zone mode in the same test to verify that they are not getting mixed, and that both storage and zone keys are migrated.

The tests run as a part of the test matrix, so we'll test PostgreSQL/MySQL and TLS/Secure Session variations.

Since we can't use the global key store, I had to refactor the `fork_connector()` method so that it's possible to override arbitrary command-line parameters, similar to `fork_acra()`. This is used to pass correct path to temporary key store.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -863,15 +863,18 @@ class BaseTestCase(PrometheusMixin, unittest.TestCase):
         return process
 
     def get_connector_tls_params(self):
-        return [
-            '--acraserver_tls_transport_enable',
-            '--tls_acraserver_sni=acraserver',
-        ]
+        return {
+            'acraserver_tls_transport_enable': True,
+            'tls_acraserver_sni': 'acraserver',
+        }
 
     def get_connector_prometheus_port(self, port):
         return port+1
 
-    def fork_connector(self, connector_port: int, acraserver_port: int, client_id: str, api_port: int=None, zone_mode: bool=False, check_connection: bool=True):
+    def fork_connector(self, connector_port: int, acraserver_port: int,
+                       client_id: str, api_port: int=None,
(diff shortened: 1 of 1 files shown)


tests/test.py
```diff
@@ -2273,6 +2280,243 @@ def test_clearing(self):
...
+
+    def new_key_store_path(self):
+        return os.path.join(self.test_dir.name, '.acrakeys.new')
+
+    def old_key_store_path(self):
+        return os.path.join(self.test_dir.name, '.acrakeys.old')
+
+    # Now we can proceed with the tests...
+
+    def test_migrate_v1_to_v2(self):
```
Lagovas: cool very readable test, and nice pattern to use domain/test specific functions which do meaningful action. forgot name of this pattern in tests *(

tests/test.py
```diff
@@ -2273,6 +2280,243 @@ def test_clearing(self):
...
+        self.migrate_key_store('v2')
+
+        # After we have migrated the keys, check the setup again.
+        with self.running_services():
+            # Old data should still be there, accessible via Connector.
+            selected = self.select_via_connector(row_id_1)
+            self.assertEquals(selected['data'], data_1.encode('ascii'))
+            self.assertEquals(selected['raw_data'], data_1)
+
+            # Key migration does not change encrypted data.
```

PR #364, opened 2020-03-25
Sync keystore-v2 with master

ilammy:
Pull in massive update in key store usage from "cop" branch (#355), along with some other things. Resolve resulting merge conflicts (by preserving both changes).

As announced in #355, I've merged `MultiKeyStore` interface into common `PrivateKeyStore` interface, so now all "decryptors" like acra-translator and acra-server can make use of multi-key capabilities for decryption of old data using rotated keys.

Though, currently only acra-translator supports that. The new interface is implemented for key store v2 as well, but AFAIK we don't have any automated tests for that, and I have not tested it manually yet. But it should work :crossed_fingers:

> **N.B.** Merge this PR as *fast-forward* to preserve the merge commit to avoid conflicts later when keystore-v2 is merged back into master.


.circleci/check_golint.sh
```diff
@@ -6,7 +6,7 @@
 # skip linter issues for files generated by protobuf compiler because generation out of our control
 result=$($GOPATH/bin/golint ./... | grep -v '.pb.go' | wc -l)
 
-if [[ $result -gt 12 ]]; then
+if [[ $result -gt 13 ]]; then
```
vixentael: ¯\_(ツ)_/¯

ilammy:
> some prs are so large that i rely on tests only

It ain't me! 😆 Most of the changes here come from the merge with #355.

Unfortunately, merge commits with conflict resolutions are notoriously hard to review because most of the tools handle them poorly. The only tool that I know that actually highlights the parts where the conflicts were is **gitk**. Web-based diff viewers usually don't bother and show the applied merge, with all the changes and resolutions mixed up. That's why rebasing works better with this workflow, and why long-lived feature branches are kinda bad.

[`5bbce4d`](https://github.com/cossacklabs/acra/pull/364/commits/5bbce4d760a8d8ee92c64885e1592ff652d88c24) fixes up the fallout from conflict resolutions. That's probably the interesting part.

PR #365, opened 2020-03-25
Protect key store flock with sync.Mutex

ilammy:
CI builds for keystore-v2 branch have been intermittently failing on the unit test for key store file locking. It turns out that synchronization is damn hard and I still have not learned how to write it correctly. This PR attempts to solve the issue that caused test failures.

BSD-style flocks are per-process. I have incorrectly assumed that these are "Linux kernel processes" (= per-thread), however this turned out to be incorrect. These locks are per-PID. Furthermore, they are recursive in a weird way: if a process requests a shared lock when holding an exclusive lock (or vice versa), the lock is transformed into the new type in a non-atomic way (that is, the lock is released and re-grabbed). And finally, goroutines do not map directly onto system threads, so even if flocks did work per-thread, this would still be incorrect in Go.

Effectively, this means that we should keep track of whether we have the lock locked and in which way. Add a sync.Mutex which will ensure that at most one goroutine in the process keeps `os.File` locked.

With these changes it does not make sense to have a single Unlock() call to unlock after both Lock() and RLock(). Introduce RUnlock() method that mirrors `sync.RWMutex`.

### DIFF
--- a/keystore/v2/keystore/filesystem/backend/file_lock.go
+++ b/keystore/v2/keystore/filesystem/backend/file_lock.go
@@ -1,0 +1,157 @@
+/*
+ * Copyright 2020, Cossack Labs Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package backend
+
+import (
+	"os"
+	"sync"
+	"syscall"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// fileLock is an interprocess read-write lock that serializes access to filesystem.
+//
+// This lock is implemented with BSD flock(2) and has corresponding semantics.
+//
+// Note that this is an advisory lock. That is, this lock can be used to guard
+// access to some filesystem resource by cooperating processes, but it does not
+// guarantee correct serialization of filesystem accesses by itself.
+//
+// Also note that flock(2) has subtly different semantics from POSIX record locks
+// created by fcntl(2). Read corresponding system manual pages for details.
+type fileLock struct {
+	lockFile *os.File
+	path     string
+	// Here's a thing: BSD file locks are per-process. That is, they are shared
+	// between threads like fds. Furthermore, they can be 'transparently'
+	// upgraded: if the process holds a shared lock and grabs it again as exclusive,
+	// the system will upgrade the lock to exclusive (possibly by releasing and
+	// re-grabbing it). This is not what we want, so keep a regular lock here
+	// to make sure that goroutines are correctly synchronized among themselves.
+	lockSync sync.Mutex
+}
+
+// newFileLock creates a lock anchored at given file path.
+func newFileLock(path string) (*fileLock, error) {
+	lock, err := os.Create(path)
+	if err != nil {
+		return nil, err
+	}
+	return &fileLock{lockFile: lock, path: path}, nil
+}
+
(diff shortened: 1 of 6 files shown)


keystore/v2/keystore/filesystem/backend/file_lock.go
```diff
@@ -0,0 +1,157 @@
...
+//
+// This lock is implemented with BSD flock(2) and has corresponding semantics.
+//
+// Note that this is an advisory lock. That is, this lock can be used to guard
+// access to some filesystem resource by cooperating processes, but it does not
+// guarantee correct serialization of filesystem accesses by itself.
+//
+// Also note that flock(2) has subtly different semantics from POSIX record locks
+// created by fcntl(2). Read corresponding system manual pages for details.
+type fileLock struct {
```
Lagovas: you know, this fileLock looks like well done code which good to publish at least in our code (make it public) to re-use, at most make this as a separate repository/module in github to allow use it other people too )

keystore/v2/keystore/filesystem/backend/file_lock.go
```diff
@@ -0,0 +1,157 @@
...
+	return l.lockFile.Close()
+}
+
+// We need to juggle *two* locks here -- the mutex and the file lock -- and
+// this is hard. So hard that we are using a regular sync.Mutex here rather
+// than seemingly more natural sync.RWMutex. Lock contention within the
+// process should be fairly minimal, and different processes still use the
+// read-write semantics, so it's okay.
+//
+// The reason for this complexity is that unlocking the flock can fail and
```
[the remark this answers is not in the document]
ilammy: > did you meet situations when flock failed? when it can occur?

No, I did not encounter situations like this. The documentation for flock() does not mention obvious and practical ways to shoot yourself in the foot. However, I'd imagine that it can fail in situations like

  - `Close()` followed by `Unlock()` causes flock() to be called for invalid file descriptor;
  - if someone puts key store on sshfs or other networked filesystem and it fails;
  - if key store is located on some removable media which gets removed with lock held;
  - some other weird failure in the kernel because of kernel reasons.

So these are largely theoretic possibilities. However, we have a practical issue of syscall.Flock() returning an error that we need to handle in some way. I felt that just ignoring it, hoping for the best, and – most importantly – risking with a possible deadlock is not robust enough. I'd be maybe okay with ignoring an error that causes other things to fail loudly with screams in logs, but if that's going to be a deadlock that silently blocks the service and makes it unresponsive, then it's definitely not okay.

keystore/v2/keystore/filesystem/backend/file_lock.go
ilammy: Yeah, it's probably a good idea. I did not feel that it's generic enough, as it's mostly based around our use case for key stores which results in effectively one file lock per process.

I guess there are libraries that provide file locking. I shopped around right now and [this one](https://pkg.go.dev/github.com/gofrs/flock?tab=doc) comes up, for example. Initially i thought that we can get away with something “simple” for our use case, but now that I look at all these things I think that maybe we should just add a dependency since that code is likely to be way more tested than this hack.

ilammy: Oh, I've read the source code of that library more closely and it seems it does not protect from the case that caused this PR. It still allows the flock to be accidentally upgraded or downgraded and relies entirely on the system call to provide synchronization. No luck!

ilammy: As for publishing this, I'll keep that in mind but I think it's too early.

I'd wait for a second legitimate use case in Acra code base before even exporting it. Maybe that use case will get us to improve the lock in some way. Having it exported from `keystore/v2/filesystem/backend` does not really increase discoverability, and I feel that this lock is too task-specific to be placed into `utils`.

As for making it into a library, I'm flattered and I think it's a good idea to open-source reusable bits. However, we'd probably need to look around more to see if there are *really* no similar libraries*, prepare the repository, write a lot of tests, etc. This will take time so it's obviously out of scope for this PR. I've created a task to track that though.

<sub>* So far I've found [1](https://github.com/gofrs/flock), [2](https://github.com/MichaelS11/go-file-lock), [3](https://github.com/golang/go/tree/master/src/cmd/go/internal/lockedfile/internal/filelock), all different and none of them is combined with in-process synchronization.</sub>

PR #369, opened 2020-03-27
Forbid v2 master keys to be the same

ilammy:
Keys should not be reused for multiple purposes, that's why key store v2 uses two distinct keys: one for encrypting private key data, and one for signing all key store data. Let's detect configuration errors and refuse to create or open a key store if both keys are the same.

Since the key store does not have direct access to the keys and cannot check for that, the only place where we can still do it is when the keys are retrieved from the environment. This also gives us a chance to point out to the user what exactly is wrong with the environment.

### DIFF
--- a/keystore/v2/keystore/crypto.go
+++ b/keystore/v2/keystore/crypto.go
@@ -17,7 +17,9 @@
 package keystore
 
 import (
+	"crypto/subtle"
 	"encoding/base64"
+	"errors"
 	"os"
 
 	keystoreV1 "github.com/cossacklabs/acra/keystore"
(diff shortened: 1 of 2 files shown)


ilammy:
> how Acra receives the signature key?

It needs to be generated by the user with `acra-keymaker`, just like the encryption master key. 

Key store v1 uses one key: ACRA_MASTER_KEY. Version 2 uses two keys:

- ACRA_MASTER_ENCRYPTION_KEY
- ACRA_MASTER_SIGNATURE_KEY

For security reasons they should not be derived from a single source, instead they should be two separate keys.

PR #370, opened 2020-03-30
Technical changes in logging customization

storojs72:
These changes are rather technical. 

Main point is soonest customization for logging process (as we discussed by voice - it is critical 1) to preserve format for logging in error cases, even if not all parameters for customization have been already loaded, and 2) to start logging only when format is customized). 

Since format is specified by actual formatter object (`CefFormatter`, `JsonFormatter`, `TextFormatter`), it is necessary to create this object first, and then to perform customization on it. For this reason - `SetFormat(*loggingFormat)` function (that previously returns nothing) is changed onto function `logging.CreateFormatter(*loggingFormat)` that returns actual formatter object with empty fields. Then, as long as parameters become ready (basically hooks and service name - which is actually known but applying it requires formatter object whatever) - those fields are initialised. 

Mentioned "approach" is applied for each service instead of previous customization via builder

### DIFF
--- a/cmd/acra-connector/acra-connector.go
+++ b/cmd/acra-connector/acra-connector.go
@@ -209,8 +209,6 @@ type Config struct {
 }
 
 func main() {
-	log.WithField("version", utils.VERSION).Infof("Starting service %v [pid=%v]", ServiceName, os.Getpid())
-
 	loggingFormat := flag.String("logging_format", "plaintext", "Logging format: plaintext, json or CEF")
 	keysDir := flag.String("keys_dir", keystore.DefaultKeyDirShort, "Folder from which will be loaded keys")
 	clientID := flag.String("client_id", "", "Client ID")
@@ -255,7 +253,11 @@ func main() {
 		os.Exit(1)
 	}
 
-	logging.Customize().SetFormat(*loggingFormat).SetServiceName(ServiceName).Complete()
+	// Start customizing logs here (directly after command line arguments parsing)
+	formatter := logging.CreateFormatter(*loggingFormat)
+	logging.SetServiceName(formatter, ServiceName)
+
+	log.WithField("version", utils.VERSION).Infof("Starting service %v [pid=%v]", ServiceName, os.Getpid())
 	log.Infof("Validating service configuration...")
 
 	if err = checkDependencies(); err != nil {
--- a/cmd/acra-connector/acra-connector.go
+++ b/cmd/acra-connector/acra-connector.go
@@ -255,7 +255,7 @@ func main() {
 
 	// Start customizing logs here (directly after command line arguments parsing)
 	formatter := logging.CreateFormatter(*loggingFormat)
-	logging.SetServiceName(formatter, ServiceName)
+	formatter.SetServiceName(ServiceName)
 	log.SetOutput(os.Stderr)
 
 	log.WithField("version", utils.VERSION).Infof("Starting service %v [pid=%v]", ServiceName, os.Getpid())
--- a/cmd/acra-connector/acra-connector.go
+++ b/cmd/acra-connector/acra-connector.go
@@ -256,6 +256,7 @@ func main() {
 	// Start customizing logs here (directly after command line arguments parsing)
 	formatter := logging.CreateFormatter(*loggingFormat)
 	logging.SetServiceName(formatter, ServiceName)
+	log.SetOutput(os.Stderr)
 
 	log.WithField("version", utils.VERSION).Infof("Starting service %v [pid=%v]", ServiceName, os.Getpid())
 	log.Infof("Validating service configuration...")
--- a/cmd/acra-translator/acra-translator.go
+++ b/cmd/acra-translator/acra-translator.go
@@ -88,6 +88,7 @@ func main() {
 	logging.SetServiceName(formatter, ServiceName)
 	log.SetOutput(os.Stderr)
 
+	log.WithField("version", utils.VERSION).Infof("Starting service %v [pid=%v]", ServiceName, os.Getpid())
 	log.Infof("Validating service configuration...")
 	cmd.ValidateClientID(*secureSessionID)
 
--- a/cmd/acra-translator/acra-translator.go
+++ b/cmd/acra-translator/acra-translator.go
@@ -83,7 +83,10 @@ func main() {
 		os.Exit(1)
 	}
 
-	logging.Customize().SetFormat(*loggingFormat).SetServiceName(ServiceName).Complete()
+	// Start customizing logs here (directly after command line arguments parsing)
+	formatter := logging.CreateFormatter(*loggingFormat)
+	logging.SetServiceName(formatter, ServiceName)
+	log.SetOutput(os.Stderr)
 
 	log.Infof("Validating service configuration...")
 	cmd.ValidateClientID(*secureSessionID)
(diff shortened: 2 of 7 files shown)


cmd/acra-connector/acra-connector.go
```diff
@@ -255,7 +253,11 @@ func main() {
 		os.Exit(1)
 	}
 
-	logging.Customize().SetFormat(*loggingFormat).SetServiceName(ServiceName).Complete()
+	// Start customizing logs here (directly after command line arguments parsing)
+	formatter := logging.CreateFormatter(*loggingFormat)
+	logging.SetServiceName(formatter, ServiceName)
```
ilammy: It's weird for **acra-connector** to be the only one without
```go
log.SetOutput(os.Stderr)
```
Is there a reason for that?

cmd/acra-translator/acra-translator.go
```diff
@@ -83,7 +83,10 @@ func main() {
 		os.Exit(1)
 	}
 
-	logging.Customize().SetFormat(*loggingFormat).SetServiceName(ServiceName).Complete()
+	// Start customizing logs here (directly after command line arguments parsing)
+	formatter := logging.CreateFormatter(*loggingFormat)
+	logging.SetServiceName(formatter, ServiceName)
+	log.SetOutput(os.Stderr)
```
ilammy: Similarly, **acra-translator** seems to be the only one without

```go
log.WithField("version", utils.VERSION).Infof("Starting service %v [pid=%v]", ServiceName, os.Getpid())
```

during startup.

logging/log_formatters.go
```diff
@@ -32,55 +32,35 @@ import (
...
-func TextFormatter(hooks []FormatterHook) logrus.Formatter {
+func TextFormatter() FormatterWrapper {
 	return &AcraTextFormatter{
 		Formatter: &logrus.TextFormatter{
 			FullTimestamp:    true,
 			TimestampFormat:  time.RFC3339,
 			QuoteEmptyFields: true,
 		},
-		Hooks: hooks,
+		Hooks: nil,
```
ilammy: Fields are initialized with `nil` by default so this can be omitted. However, I don't particularly mind explicit initialization either.

logging/log_formatters.go
```diff
@@ -126,6 +106,14 @@ type AcraTextFormatter struct {
 	Hooks []FormatterHook
 }
 
+func (f *AcraTextFormatter) SetServiceName(serviceName string) {
+	// service name is ignored by plaintext formatter, so just do nothing
+}
```
ilammy: Oh... This is news to me. I thought it is mapped onto a log field or something, but now I've noticed that indeed, it's not really used in any way by the plaintext formatter.

By the way, is there a particular reason for why we have the “service name” thing at all, @Lagovas, @shadinua? Shouldn't the log connector be aware of the log source so that it's not necessary to set it? Or is it only because CEF has a mandatory field for that? Why JSON follows suit then?

logging/logging.go
```diff
@@ -71,6 +69,15 @@ func IsDebugLevel(logger *log.Entry) bool {
 	return logger.Level == log.DebugLevel
 }
 
+// FormatterWrapper wraps log.Formatter interface and adds functions for customizations.
+// Intention for this interface is to provide ability to customize logging by accustomed:
+// `logging.SetServiceName` / `logging.SetHooks` from main function of Acra services
+type FormatterWrapper interface {
```
ilammy: I'd call it just `Formatter` since the users don't need to actively *wrap* anything into it. It's just that `logging.Formatter` is a `logrus.Formatter` with a couple of additional methods that we need. Acra formatters already implement this interface.

Nice use of interface for customization, by the way! 👍

logging/logging.go
```diff
@@ -84,79 +91,41 @@ func SetLogLevel(level int) {
...
+// CreateFormatter creates formatter object
+func CreateFormatter(format string) FormatterWrapper {
+	var formatter FormatterWrapper
+	switch strings.ToLower(format) {
+	case JsonFormatString:
+		formatter = JSONFormatter()
+	case CefFormatString:
+		formatter = CEFFormatter()
+	default:
+		formatter = TextFormatter()
```
ilammy: I see that it's the behavior we had before, but is it okay to let the users say `--log_format "cat videos"` and get the default one without any error?

logging/log_formatters.go
Lagovas: it's new for me too) yes, maybe we added it first when needed cef format. but anyway it's good to log source of logs especially when we have several running services to distinguish them

logging/log_formatters.go
Lagovas: omit default values is good to keep fewer lines of code, less time to verify formatting, etc ) but anyway didn't find any suggestions/conventions on [effective go](https://golang.org/doc/effective_go.html), so up to you

logging/logging.go
```diff
@@ -40,9 +38,9 @@ const (
 )
 
 const (
-	LoggingFormatPlaintext = "plaintext"
-	LoggingFormatJSON      = "json"
-	LoggingFormatCEF       = "cef"
+	PlaintextFormatString = "plaintext"
```
Lagovas: why renamed?

logging/logging.go
```diff
@@ -84,79 +91,41 @@ func SetLogLevel(level int) {
...
+	return formatter
 }
 
-// SetOutput specifies where logs should be written (stderr, file, etc.)
-func (c *CustomizeBuilder) SetOutput(w io.Writer) *CustomizeBuilder {
-	c.writer = w
-	return c
+// SetServiceName adds service-name label to log entries
+// (plaintext formatter ignores it)
+func SetServiceName(formatter FormatterWrapper, serviceName string) {
```
Lagovas: why not add method `SetServiceName` to formatter object? Why we use OOP with functions and explicit first argument to object?)

cmd/acra-connector/acra-connector.go
storojs72: no. Just my oversight, thank you

cmd/acra-translator/acra-translator.go
storojs72: thank you

logging/logging.go
storojs72: Yes, probably it is better to throw some error like `unexpected logs format has been specified`

logging/logging.go
storojs72: Because IDE highlighted those constants for me as they start same as package name

logging/logging.go
storojs72: The reason is just stylistic. Previously we used to apply customizations like `log.SetOutput()`, `logging.SetServiceName()`, etc. So I decided to keep this style.

storojs72:
- added missing `log.SetOutput` to connector;
- added missing `log.WithField("version", utils.VERSION).Infof("Starting service %v [pid=%v]", ServiceName, os.Getpid())` to translator;
- renamed `FormatterWrapper` -> `Formatter`

logging/logging.go
storojs72: Added https://github.com/cossacklabs/acra/pull/370/commits/4247aed0c2c46a91927634a72af3928a45511183 after @ilammy 's explanation https://github.com/cossacklabs/acra-enterprise/pull/41#discussion_r400994241

PR #372, opened 2020-04-02
Improve build scripts

shadinua:
* Makefile now has:
  - clean, readable and maintainable structure (I hope)
  - strict division by sections and namespaces
  - OS detection
  - correct error handling for most cases
* implemented build/install:
  - build target builds the binaries in the ./build subdirectory
  - install target builds and installs binaries into system/user $GOPATH directory
  - test/test_all run tests
  - clean removes all created artifacts
* refined packaging:
  - single target pkg
  - auto detection of package type
  - deb and rpm targets are temporarily left for backward compatibility
  - supported Debian >= 9, Ubuntu >= 16.04, RHEL >= 7, CentOS >= 7
* refined docker build flow:
  - useful scripts added for easy automation
  - dockerfiles refined
  - labels refined
* implemented help system:
  - try this: make help
  - auto generated description for targets
  - auto generated description for properties/variables

Rules for adding docker tags:
* VCS hash — always
* VCS version — always
* VCS branch¹ — always
* + `latest` — when VCS branch is `stable`
* + `current` — when VCS branch is `master`
* + specified in `DOCKER_EXTRA_(BUILD|PUSH)_TAGS`

¹ — VCS branch is preliminarily adapted to comply with naming conventions. For a typical custom branch name it looks like replacing 'username/T000_branch_name' to 'username.T000-branch-name'. If, despite the naming conventions, the resulting branch name begins with a minus symbol, the word 'branch' will be added to the beginning.

### DIFF
--- a/CHANGELOG_DEV.md
+++ b/CHANGELOG_DEV.md
@@ -1,0 +1,20 @@
+## 0.85.0 - 2020-04-02
+### Added
+- Support of RHEL >= 7
+- Configurable build and install parameters in Makefile (see `make help`)
+- Self-documented Makefile
+- `go_install.sh` script
+- Makefile `pkg` target with automatic detection of OS (use it instead of `rpm` and `deb`)
+### Changed
+- Build image use Debian 10 instead of Debian 9
+- Application names in Docker image description got `CE` suffix
+- Refined logic of automatic image tagging
+### Removed
+- Makefile targets `dist`, `temp_copy`
+- `docker_push` target replaced with `docker-push`
+- default argument `--db_host=postgres` from `acra-server` docker image, specifying it explicitly is more secure
+- default argument `--acraserver_connection_host=acra-server` from `acra-connector` image
+### Deprecated
+- `docker` target in Makefile (will be removed in 0.87.0), use `docker-build` instead
+- docker images `acra-authmanager` and `acra-keymaker` (will be removed in 0.87.0); all tools are now packaged into the `acra-tools` image
+- Makefile targets `rpm` and `deb` are aliases for `pkg` and will be removed in future
--- a/Makefile
+++ b/Makefile
@@ -296,9 +296,8 @@ endif
 		--package '$(BUILD_DIR_ABS)/$(PKG_TYPE)/$(PKG_NAME)' \
 		--version '$(PKG_VERSION)' \
 		--category security \
-		--depends openssl \
 		--depends libthemis \
-		--conflicts acra \
+		--conflicts acra-ee \
 		$(PKG_TYPE_SPECIFIC_ARGS) \
 		'$(BUILD_DIR_ABS)/$(PKG_TYPE).struct/bin=$(PKG_INSTALL_PREFIX)'
 	@find '$(BUILD_DIR_ABS)' -name \*.$(PKG_TYPE)
--- a/Makefile
+++ b/Makefile
@@ -1,190 +1,310 @@
-ifneq ($(BUILD_PATH),)
-    BIN_PATH = $(BUILD_PATH)
-else
-    BIN_PATH = build
-endif
+#===== Variables ===============================================================
 
-ifeq ($(PREFIX),)
-    PREFIX = /usr
-endif
+#----- Application -------------------------------------------------------------
 
-TEMP_GOPATH = temp_gopath
-ABS_TEMP_GOPATH := $(shell pwd)/$(TEMP_GOPATH)
+APP_NAME := acra
+APP_LICENSE_NAME = Apache License Version 2.0
 
-ifneq ($(GIT_BRANCH),)
-    BRANCH = $(GIT_BRANCH)
-else
-    BRANCH = master
-endif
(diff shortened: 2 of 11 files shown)


CHANGELOG_DEV.md
```diff
@@ -0,0 +1,20 @@
+## 0.85.0 - 2020-04-02
```
ilammy, line 1: I wonder if there’s a point in maintaining a “sloppy” CHANGELOG_DEV.md if we already have a proper CHANGELOG.md. I’d probably put all of this there right away.

Makefile
```diff
@@ -1,190 +1,310 @@
...
+pkg: build
+ifeq ($(OS_NAME),unknown)
+	$(error OS is not detected)
+endif
+ifeq ($(OS_CODENAME),unknown)
+	$(error OS codename is not detected)
+endif
+ifeq ($(OS_VERSION_MAJOR),unknown)
+	$(error OS version is not detected)
+endif
```
ilammy, line 262: I have a feeling that all variable-setting lines could be placed in the makefile directly instead of being executed only on `make pkg` via `$(eval)`. I don’t think they take that much time but eval makes them harder to read and understand.

There is a point in raising an error when the current OS cannot be detected only on `make pkg`, when we actually care about the OS. This should be done only during `make pkg`, of course.

Also, it would probably look more readable if this were

```make
pkg: build
        tools/make_pkg.sh
```
Now that we add a bulk of code here it’s even more prominent.

Makefile
```diff
@@ -1,190 +1,310 @@
...
+		--url '$(APP_VENDOR_URL)' \
+		--description $(APP_SHORT_DESCRIPTION) \
+		--vendor '$(APP_VENDOR_NAME)' \
+		--maintainer '$(shell printf "$(APP_VENDOR_NAME) <$(APP_VENDOR_EMAIL)>")' \
+		--package '$(BUILD_DIR_ABS)/$(PKG_TYPE)/$(PKG_NAME)' \
+		--version '$(PKG_VERSION)' \
+		--category security \
+		--depends openssl \
+		--depends libthemis \
+		--conflicts acra \
```
ilammy, line 301: I believe this should be
```suggestion
		--conflicts acra-ee \
```

Makefile
```diff
@@ -1,190 +1,310 @@
...
+		--license '$(APP_LICENSE_NAME)' \
+		--url '$(APP_VENDOR_URL)' \
+		--description $(APP_SHORT_DESCRIPTION) \
+		--vendor '$(APP_VENDOR_NAME)' \
+		--maintainer '$(shell printf "$(APP_VENDOR_NAME) <$(APP_VENDOR_EMAIL)>")' \
+		--package '$(BUILD_DIR_ABS)/$(PKG_TYPE)/$(PKG_NAME)' \
+		--version '$(PKG_VERSION)' \
+		--category security \
+		--depends openssl \
+		--depends libthemis \
```
ilammy, line 300: Do Acra tools depend on OpenSSL directly, not only through Themis?

CHANGELOG_DEV.md
shadinua: `CHANGELOG.md`:
* contains summary for each release
* is actual for `stable` branch only
`CHANGELOG_DEV.md`:
* contains a bit more technical details
* something like communication between developers
* is actual for each branch where it present

Makefile
shadinua: We discussed the approach previously. While it "not so huge and ugly" it looks preferrable to keep it inside the Makefile (300 lines logically divided into blocks by ≈25 lines already on the verge, but still allow to keep all in one file).

Makefile
shadinua: Sure. Thanks!

Makefile
shadinua: I think we can remove the direct dependency.

Makefile
ilammy: > We discussed the approach previously.

“If you have an important point to make, hit it once. Then come back and hit it again. Then hit it a third time.” 😉

(Yes, it’s just annoying to remember to write `\` at the end of lines, properly escape various special symbols, build && giant && chains && of && commands && like && in && Dockerfiles, etc.)

CHANGELOG_DEV.md
ilammy: Almost everything you add here is going to be important for release notes too.

The developers could benefit from links to pull requests that we maintain in CHANGELOG.md. That makes it possible to see all the gory technical details if necessary.

I also believe it is more predictable to have a single CHANGELOG.md which is actual for the branch it is in. That kinda simplifies things. If I want to read a changelog for stable, I check out stable and read it there. If I want master version, I check out master. If I want to peek at some long-running development branch then I just go there. I still need to do this with CHANGELOG_DEV, but now I need to peek into a different file and disregard the regular CHANGELOG which just sits there being a pretty flower. I don’t know. It just looks as if we did all the development in a separate directory `src_dev` and then merged it into `src` during the release.

PR #374, opened 2020-04-10
Key management multi-tool: "acra-keys"

ilammy:
After much discussion we have settled on the name of a multi-tool that we are going to use for managing key stores: `acra-keys`. Rename and refactor the current acra-key-tool into acra-keys.

acra-keys uses subcommand-based interface. For example, to read a key:

    acra-keys --client_id=Alice read storage-public

No new commands are added here, only `read` and `destroy` are supported right now.

The code has also been cleaned up a bit and split into a package to allow later reuse in Acra EE.

There is also some refactoring in common code for configuration parsing. Now it does not interfere with subcommands and positional arguments, and the flag pretty-printing code can be reused.


shadinua:
Will it work if we place the arguments in the logical order `binary` → `command` → `sub-command` → `arguments`?
```
acra-keys read storage-public --client_id=Alice
```

ilammy:
Right now — no. Unfortunately, the argument parser we use cannot handle positional arguments before options (or intermixed with them), so it expects non-options like subcommands and their positional arguments to come after the options.

Currently the command-line grammar is something like this:

```
command-line ::= "acra-keys" option* sub-command option* argument*

option ::= ("-" | "--") identifier
        |  ("-" | "--") identifier (" " | "=") string

sub-command ::= string

argument ::= string
```

I'm not really happy with it. In particular, I don't like that client/zone ID is specified with command-line option which cannot be associated with positional arguments. This may come important in command lines for key export, for example, where users may want to export multiple keys for several zones or clients in one batch. How do you specify that?..

```
acra-keys export --client_id=Alice storage-public --client_id=Bob connector-public --client_id=Carol
```

Ding-dong, quiz time! What keys are client IDs associated with and why?

Having `--client_id` option may be useful for limiting action to the specified client ID, like if you want to list the keys, or nuke all keys for the specified client. However, it's not that convenient for identifying specific keys. I was thinking along the way of multipart names:

```
acra-keys export "storage-public/client:Alice Liddell"  "storage-public/client:Mad Hatter"
```

or just straight up numeric IDs (as suggested by @Lagovas and @vixentael in discussion):

```
acra-keys export 40CA2721 5E7F7E65
```

but neither of them seems to be really convenient and obviously superior.

Another option is to use a configuration file for anything non-trivial (i.e., not a single key) and it can be as complex as necessary. It makes sense to have premade configurations if you are going to export and import the same keys over and over for rotation. Unfortunately, our current config parser can only handle simple key-value list that maps directly onto command-line options, so it might not be suitable.

shadinua:
```
acra-keys export --client_id=Alice storage-public --client_id=Bob connector-public --client_id=Carol
```
Indeed, it looks somewhat intimidating :) As far as I understand, there are combined multiple options (in terms described above). 

What if we try to bring everything to the following form:
```
acra-keys <sub-command> [--option1[=value1] [--options2[=value2] [...] ] ]
```
In this case, we always strictly adhere to a certain order, regardless of the sub-command, we save the format of the passing options.

> Having --client_id option may be useful for limiting action to the specified client ID, like if you want to list the keys, or nuke all keys for the specified client. However, it's not that convenient for identifying specific keys.

We could implement this in such alternative ways:
* specify multiple options: `--key-include="[key_type_1:]key_name_1" --key-include="[key_type_2:]key_name_2" ...`
* ... or pass them in file: `--keys-include-file=/path/to/file_with_keys`

Also we could preserve passing IDs: `--key-include="id:key_id_1" --key-include="id:key_id_2" ...`.

cmd/acra-keys/keys/command-line.go
```diff
@@ -0,0 +1,255 @@
...
+
+// Params provide global access to command-line parameters.
+var Params *CommandLineParams = &CommandLineParams{}
+
+// Register configures command-line parameter parsing.
+func (params *CommandLineParams) Register() {
+	flag.StringVar(&params.KeyStoreVersion, "keystore", "", "force key store format: v1 (current), v2 (new)")
+	flag.StringVar(&params.KeyDir, "keys_dir", DefaultKeyDirectory, "path to key directory")
+	flag.StringVar(&params.KeyDirPublic, "keys_dir_public", "", "path to key directory for public keys")
+	flag.StringVar(&params.ClientID, "client_id", "", "client ID for which to retrieve key")
```
Lagovas, line 103: can we use all these global params separately for each command to provide <acra-keys> <command> <params> interface? copy them to each flagset. yes, we will duplicate some of them, but it is more common in use and we will not have hard to understand cli as discussed in "conversation" tab
golang has one popular library - https://github.com/spf13/cobra which do similar. we can look on some patterns and implementation details to use in acra.

cmd/acra-keys/keys/destroy-key.go
```diff
@@ -0,0 +1,75 @@
...
+	KeyTransportConnector,
+	KeyTransportServer,
+	KeyTransportTranslator,
+}
+
+// DestroyKey destroys data of the requsted key.
+func DestroyKey(params *CommandLineParams, keyStore keystore.KeyMaking) error {
+	switch params.DestroyKeyKind {
+	case KeyTransportConnector:
+		if params.ClientID == "" {
```
Lagovas, line 35: why not to move this check out of switch and validate with function like `func validateClientID(keyKind <type, clientID []byte) bool`? Each check different from each other by name of kind of key which implement `String()` method and correctly formatted by .Errorf.

cmd/acra-keys/keys/read-key.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		key, err := keyStore.GetServerDecryptionPrivateKey([]byte(Params.ClientID))
+		if err != nil {
+			log.WithError(err).Error("Cannot read client storage private key")
+			return nil, err
+		}
+		return key.Value, nil
+
+	case KeyZonePublic:
+		if Params.ZoneID == "" {
```
Lagovas, line 87: same comment for zoneid checks. we can write one function which will accept zoneid and key kind, place it in some utils.go and re-use for all commands.

cmd/acra-keys/keys/destroy-key.go
ilammy: > why not to move this check out of switch and validate with function

I'm not really sure it's worth it. Each key kind has its own requirements. Some need client ID, some need zone ID, some don't need anything. I've placed the checks into individual `case` clauses to keep them near the code that uses client/zone ID so it's more obvious why we check for that.

Though, it in this case we mostly do this to check that the user has specified `--client_id` on the command line when it's required by the key kind. So it makes sense to move this to command-line validation. Though, we'd need to parse the subcommand in order to know what key kinds we care about. So it will be more natural to have some

```go
func (params *CommandLineParams) CheckForKeyKind(keyKind string) error {
    // if something required is not set on the command line, return an error
}
```

How does that sound?

It does not seem to be very reusable outside of acra-keys (which is the only tool that cares about key kinds directly), but it should reduce duplication of such checks when more subcommands are added.

cmd/acra-keys/keys/command-line.go
ilammy: > can we use all these global params separately for each command to provide interface? 

I'm sorry, I don't follow your idea. Could you please elaborate? Do you mean to have a separate _complete_ flag set for each subcommand instead of splitting some global options into a root flag set and having subcommand-specific options in their own flag sets? Or have only a global flag set with all options of all subcommand, just some of them will not be used?

> golang has one popular library - https://github.com/spf13/cobra 

Honestly, I'd like to pull in something like that as a new dependency to make command line parsing more flexible. Using `flags` seems to be fine for flat configuration and an occasional option or two for a single-purpose tool. However, anything more complex becomes a mess.

What do you think about adding it as a dependency?
(In a separate PR, only for acra-keys at first.)

Lagovas: > I'm sorry, I don't follow your idea. Could you please elaborate? Do you mean to have a separate complete flag set for each subcommand instead of splitting some global options into a root flag set and having subcommand-specific options in their own flag sets? Or have only a global flag set with all options of all subcommand, just some of them will not be used?

exclude `--client_id`/`--zone_id` from global options and require them per each command. to avoid cases with `--client_id=client <command> --client_id=client2`

> What do you think about adding it as a dependency?
> (In a separate PR, only for acra-keys at first.)

actually personally I didn't do it because I wasn't sure that this library ok for our license politic) because this library uses too many other third-parties with different licenses) so I haven't anything against it if it's okay with license.

ilammy: > because this library uses too many other third-parties with different licenses

Oh... For some reason I thought it's some minimal argument parser or something. However, after I did

```
$ cat go.mod
module example.com/module

go 1.13

require github.com/spf13/cobra v1.0.0
```
and then
```
$ cat go.sum | awk '{print $1}' | sort -u | wc -l
      98
```
<details>
<summary>Dependency list</summary>

```
$ cat go.sum | awk '{print $1}' | sort -u
cloud.google.com/go
github.com/BurntSushi/toml
github.com/OneOfOne/xxhash
github.com/alecthomas/template
github.com/alecthomas/units
github.com/armon/consul-api
github.com/beorn7/perks
github.com/cespare/xxhash
github.com/client9/misspell
github.com/coreos/bbolt
github.com/coreos/etcd
github.com/coreos/go-semver
github.com/coreos/go-systemd
github.com/coreos/pkg
github.com/cpuguy83/go-md2man/v2
github.com/davecgh/go-spew
github.com/dgrijalva/jwt-go
github.com/dgryski/go-sip13
github.com/fsnotify/fsnotify
github.com/ghodss/yaml
github.com/go-kit/kit
github.com/go-logfmt/logfmt
github.com/go-stack/stack
github.com/gogo/protobuf
github.com/golang/glog
github.com/golang/groupcache
github.com/golang/mock
github.com/golang/protobuf
github.com/google/btree
github.com/google/go-cmp
github.com/gorilla/websocket
github.com/grpc-ecosystem/go-grpc-middleware
github.com/grpc-ecosystem/go-grpc-prometheus
github.com/grpc-ecosystem/grpc-gateway
github.com/hashicorp/hcl
github.com/inconshreveable/mousetrap
github.com/jonboulle/clockwork
github.com/julienschmidt/httprouter
github.com/kisielk/errcheck
github.com/kisielk/gotool
github.com/konsorten/go-windows-terminal-sequences
github.com/kr/logfmt
github.com/kr/pretty
github.com/kr/pty
github.com/kr/text
github.com/magiconair/properties
github.com/matttproud/golang_protobuf_extensions
github.com/mitchellh/go-homedir
github.com/mitchellh/mapstructure
github.com/mwitkow/go-conntrack
github.com/oklog/ulid
github.com/pelletier/go-toml
github.com/pkg/errors
github.com/pmezard/go-difflib
github.com/prometheus/client_golang
github.com/prometheus/client_model
github.com/prometheus/common
github.com/prometheus/procfs
github.com/prometheus/tsdb
github.com/rogpeppe/fastuuid
github.com/russross/blackfriday/v2
github.com/shurcooL/sanitized_anchor_name
github.com/sirupsen/logrus
github.com/soheilhy/cmux
github.com/spaolacci/murmur3
github.com/spf13/afero
github.com/spf13/cast
github.com/spf13/cobra
github.com/spf13/jwalterweatherman
github.com/spf13/pflag
github.com/spf13/viper
github.com/stretchr/objx
github.com/stretchr/testify
github.com/tmc/grpc-websocket-proxy
github.com/ugorji/go
github.com/xiang90/probing
github.com/xordataexchange/crypt
go.etcd.io/bbolt
go.uber.org/atomic
go.uber.org/multierr
go.uber.org/zap
golang.org/x/crypto
golang.org/x/lint
golang.org/x/net
golang.org/x/oauth2
golang.org/x/sync
golang.org/x/sys
golang.org/x/text
golang.org/x/time
golang.org/x/tools
google.golang.org/appengine
google.golang.org/genproto
google.golang.org/grpc
gopkg.in/alecthomas/kingpin.v2
gopkg.in/check.v1
gopkg.in/resty.v1
gopkg.in/yaml.v2
honnef.co/go/tools
```

</details>

In this case I'd say, “No, thanks”, _regardless_ of whether the licenses of all these libraries are acceptable or not. I don't think we should multiply our indirect dependency list by 10 because of that library.

ilammy:
@Lagovas, I've updated argument parsing, please give it a look.

Required options like `--client_id` and `--zone_id` for certain key kinds are now validated in one place, without much duplication in subcommand code.

These options are now only global. `read` and `destroy` subcommands take no options of their own, so reading keys can be performed only like
```
acra-keys --client_id Alice read storage-public
```
with `--client_id` coming before subcommand name (`read`).

PR #375, opened 2020-04-13
Support storage key rotation

ilammy:
Teach remaining services working with AcraStructs – **acra-server, acra-rollback, acra-rotate** – to use historical keys for decryption. Previously only acra-translator supported that. Now it should be truly possible to rotate keys without reencrypting the data.

While we're here, clean up key wiping as well. Now we wipe keys on acra-server as soon as we don't need them.

Tests for acra-rotate are quite involved so I do not want to copy-paste them with key rotation added. I have pulled a library for data-driven testing ([`ddt`](https://ddt.readthedocs.io/en/latest/index.html)) to cover cases with and without key rotation. It's not pretty because of some inheritance issues, but it works. @Lagovas, if you have any ideas how to make it better (ideally, without a massive refactoring), I'm all ears.

Unfortunately, right now it's not possible to rotate zone storage keys alone, without creating a new zone, so it's impossible to test that. The code is updated to support zones, but there are TODOs instead of tests.

I don't want to waste time on merge conflict resolution and would like to test this with key store v2 as well, so these changes target the **keystore-v2** branch, not master. It is possible to apply them onto master, but with some conflicts around the `KeyStore` methods for multiple keys.

This PR also does not cover _poison records_. They use rotatable keys, but currently do not really support rotation (including the case via AcraTranslator). This will be added later since it's more involved and requires updates in the key store to be able to retrieve historical poison keys.

### DIFF
--- a/cmd/acra-rotate/rotator.go
+++ b/cmd/acra-rotate/rotator.go
@@ -59,11 +59,7 @@ func (rotator *keyRotator) rotateAcrastructWithZone(zoneID, acrastruct []byte) (
 		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't get private key")
 		return nil, err
 	}
-	defer func() {
-		for _, key := range privateKeys {
-			utils.ZeroizePrivateKey(key)
-		}
-	}()
+	defer utils.ZeroizePrivateKeys(privateKeys)
 	decrypted, err := base.DecryptRotatedAcrastruct(acrastruct, privateKeys, zoneID)
 	if err != nil {
 		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't decrypt AcraStruct")
@@ -99,11 +95,7 @@ func (rotator *keyRotator) rotateAcrastructWithClientID(clientID, acrastruct []b
 		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't get private key")
 		return nil, err
 	}
-	defer func() {
-		for _, key := range privateKeys {
-			utils.ZeroizePrivateKey(key)
-		}
-	}()
+	defer utils.ZeroizePrivateKeys(privateKeys)
 	decrypted, err := base.DecryptRotatedAcrastruct(acrastruct, privateKeys, nil)
 	if err != nil {
 		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't decrypt AcraStruct")
--- a/cmd/acra-rotate/rotator.go
+++ b/cmd/acra-rotate/rotator.go
@@ -101,7 +101,7 @@ func (rotator *keyRotator) rotateAcrastructWithClientID(clientID, acrastruct []b
 	}
 	defer func() {
 		for _, key := range privateKeys {
-			utils.FillSlice(0, key.Value)
+			utils.ZeroizePrivateKey(key)
 		}
 	}()
 	decrypted, err := base.DecryptRotatedAcrastruct(acrastruct, privateKeys, nil)
--- a/cmd/acra-rotate/rotator.go
+++ b/cmd/acra-rotate/rotator.go
@@ -54,13 +54,17 @@ func (rotator *keyRotator) rotateAcrastructWithZone(zoneID, acrastruct []byte) (
 	logger := log.WithFields(log.Fields{"ZoneId": string(zoneID)})
 	logger.Infof("Rotate AcraStruct")
 	// rotate
-	privateKey, err := rotator.keystore.GetZonePrivateKey(zoneID)
+	privateKeys, err := rotator.keystore.GetZonePrivateKeys(zoneID)
 	if err != nil {
 		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't get private key")
 		return nil, err
 	}
-	defer utils.FillSlice(0, privateKey.Value)
-	decrypted, err := base.DecryptAcrastruct(acrastruct, privateKey, zoneID)
+	defer func() {
+		for _, key := range privateKeys {
+			utils.ZeroizePrivateKey(key)
+		}
+	}()
+	decrypted, err := base.DecryptRotatedAcrastruct(acrastruct, privateKeys, zoneID)
 	if err != nil {
 		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't decrypt AcraStruct")
 		return nil, err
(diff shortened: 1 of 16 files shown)


cmd/acra-rotate/rotator.go
```diff
@@ -90,13 +94,17 @@ func (rotator *keyRotator) rotateAcrastructWithClientID(clientID, acrastruct []b
...
+	privateKeys, err := rotator.keystore.GetServerDecryptionPrivateKeys(clientID)
 	if err != nil {
 		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't get private key")
 		return nil, err
 	}
-	defer utils.FillSlice(0, privateKey.Value)
-	decrypted, err := base.DecryptAcrastruct(acrastruct, privateKey, nil)
+	defer func() {
+		for _, key := range privateKeys {
+			utils.FillSlice(0, key.Value)
```
vixentael, line 104: shouldn't it be `utils.ZeroizePrivateKey(key)`?

tests/test.py
```diff
@@ -2582,6 +2585,37 @@ def test_moved_key_store(self):
...
+        acra_struct = create_acrastruct(data.encode('ascii'), public_key)
+        self.engine1.execute(
+            test_table.insert(),
+            {'id': row_id, 'data': acra_struct, 'raw_data': data})
+
+        # After that rotate the storage key for the client,
+        # but don't touch the encrypted data.
+        create_client_keypair(client_id, only_storage=True)
+
+        # Request via AcraConnector should be successful.
```
vixentael, line 2609: here i'd put new data, then read old, and new.

to prove that both old data (old key) and new data (new key) is readable

tests/test.py
```diff
@@ -2818,6 +2853,40 @@ def test_without_placeholder(self):
...
+                'data': create_acrastruct(data.encode('ascii'), server_public1),
+                'id': get_random_id()
+            }
+            rows.append(row)
+        self.engine_raw.execute(test_table.insert(), rows)
+
+        # Rotate storage keys for 'keypair1'
+        create_client_keypair('keypair1', only_storage=True)
+
+        # Run acra-rollback for the test table
```
vixentael, line 2876: same here, put new data after rotation

cmd/acra-rotate/rotator.go
ilammy: > shouldn't it be `utils.ZeroizePrivateKey(key)`?

Yeah, I missed the conversion here. Thanks.

tests/test.py
ilammy: > here i'd put new data, then read old, and new.

Makes sense to check that the new key can be used as well. I'll add it.

cmd/acra-rotate/rotator.go
```diff
@@ -54,13 +54,17 @@ func (rotator *keyRotator) rotateAcrastructWithZone(zoneID, acrastruct []byte) (
...
-	privateKey, err := rotator.keystore.GetZonePrivateKey(zoneID)
+	privateKeys, err := rotator.keystore.GetZonePrivateKeys(zoneID)
 	if err != nil {
 		logger.WithField("acrastruct", hex.EncodeToString(acrastruct)).WithError(err).Errorln("Can't get private key")
 		return nil, err
 	}
-	defer utils.FillSlice(0, privateKey.Value)
-	decrypted, err := base.DecryptAcrastruct(acrastruct, privateKey, zoneID)
+	defer func() {
+		for _, key := range privateKeys {
```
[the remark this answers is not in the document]
ilammy: > imho it's good to add some wrapper which do all of this

Yeah, that thought crossed my mind after I've written this loop for the third time. I'll add the utility.

ilammy:
I've updated a bunch of `Decryptor` utilities that manually decrypt AcraStructs. In particular this should fix MySQL support that uses that code path instead of `DecryptAcrastruct()`, and tests should be green again.

decryptor/base/decryptor.go
```diff
@@ -97,7 +97,7 @@ type DataDecryptor interface {
 	// acrastruct using secure message
 	// return decrypted data or data as is if fail
 	// db specific
-	ReadSymmetricKey(*keys.PrivateKey, io.Reader) ([]byte, []byte, error)
+	ReadSymmetricKey([]*keys.PrivateKey, io.Reader) ([]byte, []byte, error)
```
Lagovas, line 100: probably it will cause compilation errors in acra-ee due to using this interface from acra-ce which changed here...
better to add new instead of changing existing. but if do so then we should provide new extended interface to not cause same compilation errors in acra-ee and use it here instead old base DataDecryptor...
or change it here and don't forget to change in acra-ee....

decryptor/postgresql/pg_hex_decryptor.go
```diff
@@ -142,12 +142,14 @@ func (decryptor *PgHexDecryptor) ReadSymmetricKey(privateKey *keys.PrivateKey, r
 	}
 	pubkey := &keys.PublicKey{Value: decryptor.decodedKeyBlockBuffer[:base.PublicKeyLength]}
 
-	smessage := message.New(privateKey, pubkey)
-	symmetricKey, err := smessage.Unwrap(decryptor.decodedKeyBlockBuffer[base.PublicKeyLength:])
-	if err != nil {
-		return nil, decryptor.keyBlockBuffer[:n], base.ErrFakeAcraStruct
+	for _, privateKey := range privateKeys {
```
Lagovas, line 145: hmm, I thought I had removed hex/escape decoders.... will see on this parts again at some time. maybe forgot to delete old code...

decryptor/base/decryptor.go
ilammy: Hm... Makes sense.

For `KeyStore` interface we have kept around the old interface that retrieves the latest private key. It should not be used now but it may still be necessary for Acra EE until keystore v2 support is added there.

I seemed to think that decryption is implementation detail of Acra CE which does not leak to Acra EE so it's possible to change the interface without compatibility considerations. However, now I've looked at the actual code and see that Acra EE has additional decryptors.

We'd better introduce new interfaces that support multiple keys alongside with single-keyed ones.

ilammy: After thinking a bit about it, I have decided that it will be easier to make it a flag day than bother with strict backwards compatibility. I have retained meaning of the old interface methods, only adding new ones. However, this still changes the interface definition, so any Acra EE code that implements this interface (and we have some copypasta there) will break. This is intentional as it makes it trivial to track which places in Acra EE need to be updated. 

However, once merged into Acra CE's `master` this will also break any long-running branches in Acra EE which base on older Acra EE's `master`. These branches will have to be rebased onto the latest Acra EE master after the merge is complete. Unless they change something in `Decryptor` usage, this should be a trivial conflict-less rebase.

An alternative is to go full ham on compatibility, introduce new methods in new `DecryptorEx` interface, teach Acra CE to work with both of them, eventually teach Acra EE to work with both of them too, and then almost immediately remove them. This will allow us to not bother people working on long-running outdated branches but has no other benefit once those branches are merged into master.

Since we strive to practice continuous integration instead of having multiple branches, I believe it will be easier to tell the developers to quickly rebase and carry on instead of going out of the way with compatibility shims to not bother them completely.

decryptor/base/decryptor.go
```diff
@@ -97,7 +97,8 @@ type DataDecryptor interface {
 	// acrastruct using secure message
 	// return decrypted data or data as is if fail
 	// db specific
-	ReadSymmetricKey([]*keys.PrivateKey, io.Reader) ([]byte, []byte, error)
+	ReadSymmetricKey(*keys.PrivateKey, io.Reader) ([]byte, []byte, error)
+	ReadSymmetricKeyRotated([]*keys.PrivateKey, io.Reader) ([]byte, []byte, error)
```
Lagovas, line 101: adding this method doesn't touch acra-ee? it wasn't used directly anywhere? if so, then lgtm

ilammy: > adding this method doesn't touch acra-ee?

It _does_ affect Acra EE. I mean, it will once this is merged into master *and* Acra EE is updated to use that commit.

Because we add new methods to interface that Acra EE reimplements from scratch in its own code (instead of inheriting Acra CE's implementation), Acra EE will need to be updated to implement new methods too. I plan to prepare a *keystore-v2* branch in Acra EE repo with relevant changes before CE's keystore-v2 lands into master. Then this update can be coordinated.

PR #376, opened 2020-04-30
Key store v2 export/import API

ilammy:
This PR is work-in-progress on key export and import. Here we add low-level API for exporting and importing keys, and implement the export code path which is easier. Importing will come later as it's a bit harder. Meanwhile I'd like to get feedback on whatever is ready at the moment.

First of all we get rid of an unused key type for public keys only. I intended it to be used when a service needs to store only public keys of its peers. However, it's not convenient to use when programming, and it's not really a new "type" – it's just the same Themis key pair, just without the private part. So, instead of inventing a new type, just reuse the existing key pair and store only the public key. Update the code to allow skipping private keys and return a more appropriate error when a private key is requested for a key pair that does not have one.

The next commit adds the data structure used for exported keys. It's a simple list of key ring data which is going to be encrypted and signed for export. Resulting binary blob is what's going to be stored for backup and transferred between machines. Its structure is the following:

- `SignedContainer` signed with an exporting key (different from the key used by the key store)
  - binary blob, result of Themis Secure Cell encryption with another exporting key for encryption 
    - `EncryptedKeys` structure containing `KeyRing` data
      Individual key rings and keys inside them are not encrypted. Instead, the entire list is encrypted to conceal its contents.

Finally, we add some API that's going to be used for exporting and importing key rings. For export you give a list of key rings you need and get an encrypted blob in return. Then this blob can be used to import the key rings inside it. In both cases you need to explicitly pass a "crypto suite" which will be used for the transfer, it is different from the suite used by the key store to access the key data. 

Also, importing operation is more involved. We have to decide what to do if the key ring already exists. I currently have no concrete vision of what situations we might run into. For example, the same backup may be imported twice, or an outdated backup applied, or an incremental one, or there is an overlap in data. Whatever it is, I want the interface to be extensible enough to be able to handle that situations when we identify them.

This PR implements only the exporting path, and it's more like a backend support. `acra-keys` tool still has to be taught to use this API, and for that we'll need to first provide an API to retrieve all available key rings.

### DIFF
--- a/keystore/v2/keystore/api/keyStore.go
+++ b/keystore/v2/keystore/api/keyStore.go
@@ -17,6 +17,11 @@
 // Package api describes API of Acra Key Store version 2.
 package api
 
+import (
+	"github.com/cossacklabs/acra/keystore/v2/keystore/asn1"
+	"github.com/cossacklabs/acra/keystore/v2/keystore/crypto"
+)
+
 // KeyStore securely keeps all of the client key data.
 type KeyStore interface {
 	// OpenKeyRing opens an existing key ring with given purpose.
@@ -25,6 +30,11 @@ type KeyStore interface {
 	// Close this keystore, releasing associated resources.
 	// This generally renders opened KeyRings unusable.
 	Close() error
+
+	// ExportKeyRings packages specified key rings for export.
+	// Key ring data is encrypted and signed using given cryptosuite.
+	// Resulting container can be imported into existing or different key store with ImportKeyRings().
+	ExportKeyRings(paths []string, cryptosuite *crypto.KeyStoreSuite) ([]byte, error)
 }
 
 // MutableKeyStore interface to KeyStore allowing write access.
@@ -34,4 +44,31 @@ type MutableKeyStore interface {
 	// OpenKeyRingRW opens a modifiable key ring with given purpose.
 	// A new key ring will be created if it does not exist yet.
 	OpenKeyRingRW(purpose string) (MutableKeyRing, error)
+
+	// ImportKeyRings unpacks key rings packaged by ExportKeyRings.
+	// The provided cryptosuite is used to verify the signature on the container and decrypt key ring data.
+	// Optional delegate can be used to control various aspects of the import process, such as conflict resolution.
+	ImportKeyRings(exportData []byte, cryptosuite *crypto.KeyStoreSuite, delegate *KeyRingImportDelegate) error
+}
+
+// ImportDecision constants describe how to proceed with import conflict resolution.
+type ImportDecision int
+
+// ImportDecision options.
+const (
+	// Do not modify existing key ring, abort with given error.
+	ImportAbort ImportDecision = iota
+	// Do not modify existing key ring, proceed with importing others.
+	ImportSkip
+	// Overwrite existing key ring with new data.
+	ImportOverwrite
+)
+
+// KeyRingImportDelegate controls details of key ring import process.
+type KeyRingImportDelegate struct {
+	// This callback is executed when an imported key ring already exists.
+	// If not set, import process will be aborted.
+	// Current key ring content is encrypted, new key ring content is in plain.
+	// Don't look at the key material, decide based on validity ranges and sequence numbers.
+	DecideKeyRingOverwrite func(currentData, newData *asn1.KeyRing) (ImportDecision, error)
 }
--- a/keystore/v2/keystore/api/keyStore.go
+++ b/keystore/v2/keystore/api/keyStore.go
@@ -48,7 +48,7 @@ type MutableKeyStore interface {
 	// ImportKeyRings unpacks key rings packaged by ExportKeyRings.
 	// The provided cryptosuite is used to verify the signature on the container and decrypt key ring data.
 	// Optional delegate can be used to control various aspects of the import process, such as conflict resolution.
-	ImportKeyRings(exportData []byte, cryptosuite *crypto.KeyStoreSuite, delegate *KeyRingImportDelegate) error
+	ImportKeyRings(exportData []byte, cryptosuite *crypto.KeyStoreSuite, delegate KeyRingImportDelegate) error
 }
 
 // ImportDecision constants describe how to proceed with import conflict resolution.
@@ -65,10 +65,9 @@ const (
 )
 
 // KeyRingImportDelegate controls details of key ring import process.
-type KeyRingImportDelegate struct {
-	// This callback is executed when an imported key ring already exists.
-	// If not set, import process will be aborted.
+type KeyRingImportDelegate interface {
+	// This method is executed when an imported key ring already exists.
 	// Current key ring content is encrypted, new key ring content is in plain.
 	// Don't look at the key material, decide based on validity ranges and sequence numbers.
-	DecideKeyRingOverwrite func(currentData, newData *asn1.KeyRing) (ImportDecision, error)
+	DecideKeyRingOverwrite(currentData, newData *asn1.KeyRing) (ImportDecision, error)
 }
--- a/keystore/v2/keystore/api/tests/key.go
+++ b/keystore/v2/keystore/api/tests/key.go
@@ -125,9 +125,8 @@ func testKeyFormatLookup(t *testing.T, newKeyStore NewKeyStore) {
 		ValidUntil: time.Now().Add(time.Hour),
 		Data: []api.KeyData{
 			api.KeyData{
-				Format:     api.ThemisKeyPairFormat,
-				PublicKey:  []byte("my public key"),
-				PrivateKey: []byte("my private key"),
+				Format:    api.ThemisKeyPairFormat,
+				PublicKey: []byte("my public key"),
 			},
 		},
 	})
@@ -135,9 +134,13 @@ func testKeyFormatLookup(t *testing.T, newKeyStore NewKeyStore) {
 		t.Fatalf("failed to add key: %v", err)
 	}
 
-	_, err = ring.PublicKey(key, api.ThemisPublicKeyFormat)
-	if err != api.ErrFormatMissing {
-		t.Errorf("found public key with key pair format: %v", err)
+	_, err = ring.PublicKey(key, api.ThemisKeyPairFormat)
+	if err != nil {
+		t.Errorf("cannot find public key with key pair format: %v", err)
+	}
+	_, err = ring.PrivateKey(key, api.ThemisKeyPairFormat)
+	if err != api.ErrNoKeyData {
+		t.Errorf("cannot find private key with key pair format: %v", err)
 	}
 
 	_, err = ring.SymmetricKey(key, api.ThemisSymmetricKeyFormat)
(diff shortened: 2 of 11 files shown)


keystore/v2/keystore/filesystem/export.go
```diff
@@ -0,0 +1,221 @@
...
+	*ringData, err = ring.exportASN1()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (s *KeyStore) importKeyRing(newRingData *asn1.KeyRing, delegate *api.KeyRingImportDelegate) error {
+	keyRing := newKeyRing(s, string(newRingData.Purpose))
+	err := s.readKeyRing(keyRing)
```
Lagovas, line 67: can you add comment that here we check existing keyring with same purpose?) I wrote block of questions what we try to do until understood) because it's clear only when read block with `err == nil` below why we read empty keyring with same purpose)

keystore/v2/keystore/api/keyStore.go
```diff
@@ -34,4 +44,31 @@ type MutableKeyStore interface {
...
+	// Do not modify existing key ring, abort with given error.
+	ImportAbort ImportDecision = iota
+	// Do not modify existing key ring, proceed with importing others.
+	ImportSkip
+	// Overwrite existing key ring with new data.
+	ImportOverwrite
+)
+
+// KeyRingImportDelegate controls details of key ring import process.
+type KeyRingImportDelegate struct {
```
Lagovas, line 68: why to not use interface instead struct?

keystore/v2/keystore/filesystem/export.go
```diff
@@ -0,0 +1,221 @@
...
+	switch err {
+	case nil:
+		// If the key store successfuly returned an existing key ring with the same name,
+		// we have to resolve this conflict somehow. Present both current and new versions
+		// to the delegate and let it decide how to proceed.
+		decision := api.ImportAbort
+		err := ErrKeyRingExists
+		if delegate != nil && delegate.DecideKeyRingOverwrite != nil {
```
Lagovas, line 75: that is why struct seems like not a comfortable choice in use. imho better to provide interface for strategy pattern which will be used on importing when keyring already exists. And provide several types of strategy: "skip"/"overwrite"/"abort". And here just call `delegate.ImportExistingData` or something like that. Which delegate to use will be chosen on start by keystore configuration and some fabric method like:
```
switch importStrategy {
case api.ImportOverwrite:
  return NewImportOverwriteStrategy()
case api.ImportSkip:
  return NewImportSkipStrategy()
case api.AbortOnOverwrite:
  return NewAbortStrategy()
}
```

then here will be more clean code, one switch branch on start and zero in runtime execution here.

keystore/v2/keystore/filesystem/export.go
```diff
@@ -0,0 +1,221 @@
...
+		Keys:    make([]asn1.Key, len(r.data.Keys)),
+	}
+	copy(exported.Keys, r.data.Keys)
+	defer func() {
+		if err != nil {
+			zeroizeKeyRing(&exported)
+		}
+	}()
+	for i := range exported.Keys {
+		decrypted, err := r.decryptAllKeyData(exported.Keys[i].Data, exported.Keys[i].Seqnum)
```
Lagovas, line 155: why we decrypt data for export? to encrypt it with "export" key and allow to encrypt with another key on import operation?

keystore/v2/keystore/filesystem/key.go
```diff
@@ -261,23 +261,19 @@ func (r *KeyRing) addKeyData(data api.KeyData, key *asn1.Key) error {
...
+	case api.ThemisKeyPairFormat:
 		if len(data.PublicKey) == 0 {
 			return api.ErrNoKeyData
 		}
 		newData.PublicKey = data.PublicKey
-
-	case api.ThemisKeyPairFormat:
-		if len(data.PublicKey) == 0 || len(data.PrivateKey) == 0 {
-			return api.ErrNoKeyData
+		if len(data.PrivateKey) != 0 {
```
Lagovas, line 269: if `len(data.PrivateKey) == 0` we do nothing, but if empty public key, we return ErrNoKeyData? Why not same behavior for private key for `ThemisKePairFormat`?

ilammy:
> @ilammy added 3 commits 19 days ago
> @Lagovas reviewed 11 days ago

Shamefur dispray 😡

keystore/v2/keystore/api/tests/key.go
```diff
@@ -190,13 +193,12 @@ func testKeyInvalidInputs(t *testing.T, newKeyStore NewKeyStore) {
...
 			api.KeyData{
-				Format:     api.ThemisPublicKeyFormat,
-				PrivateKey: []byte("ompf"),
+				Format: api.ThemisKeyPairFormat,
 			},
 		},
 	})
 	if err != api.ErrNoKeyData {
-		t.Errorf("cannot add public key without data: %v", err)
+		t.Errorf("cannot add key pair without any data: %v", err)
```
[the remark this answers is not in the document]
ilammy: > we can add key without any data for cases when key material removed but we store metadata for history, yes?

Not really. This test verifies that it is not allowed to add *only* key metadata, without any key material. The reasoning is that you should not be backdating metadata alone.

However, it is possible (or, well, should be possible) to add a key with metadata then at some point destroy the key material so that only metadata remains.

Though, it’s an open question how this should interact with key export/import. I don’t know whether we should allow exporting destroyed keys and importing them back (which will be metadata-only import).

keystore/v2/keystore/api/tests/keyStore.go
```diff
@@ -19,9 +19,110 @@ package tests
...
+	if err != nil {
+		t.Errorf("failed to add public key: %v", err)
+	}
+}
+
+func testKeyStoreExport(t *testing.T, newKeyStore NewKeyStore) {
+	s := newKeyStore(t)
+	setupDemoKeyStore(s, t)
+
+	_, err := s.ExportKeyRings(exportRingAll, newExportStoreSuite(t))
```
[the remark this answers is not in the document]
ilammy: > why not check that exported data is the same as expected?

It’s all encrypted so it’s hard to verify the contents. Right now this test is more of a placeholder to check that you can export stuff without an error. It will be properly verified in the next PR that adds import. Then we’d try importing the exported data and check that we get the key data we should get.

keystore/v2/keystore/filesystem/export.go
ilammy: > can you add comment that here we check existing keyring with same purpose?

Will do later. Right now `importKeyRing` is more of a draft, it will be expanded in the next PR. I hope it will be more readable then.

keystore/v2/keystore/filesystem/export.go
ilammy: > why we decrypt data for export? to encrypt it with "export" key and allow to encrypt with another key on import operation?

Yes, that is correct. Suppose you’re transferring keys between key stores A and B. Each of them uses their own encryption keys and at the moment of export from A we don’t know where the keys will be imported to. During export we’ll decrypt them with key A, encrypt them back with a temporary key T. During import the T key will be transferred to the B place (in some unspecified, secure way), then the keys will be decrypted with T and encrypted for B with B’s key.

keystore/v2/keystore/filesystem/key.go
ilammy: > if `len(data.PrivateKey) == 0` we do nothing, but if empty public key, we return ErrNoKeyData? Why not same behavior for private key for ThemisKePairFormat?

Because it makes sense to have a public part of the key pair stored without a corresponding private key. We definitely don’t want to store neither of them. It also makes little sense to allow to store the private key alone, given how we use it with Themis. Themis allows to specify only private key for signature verification in Secure Message but we don’t really use it that way in Acra.

keystore/v2/keystore/api/keyStore.go
ilammy: > why to not use interface instead struct?

Hm... The idea was that there could be more than one aspect of key import process that the user would like to control, so I wanted to make it possible to specify them independently, possibly without specifying some parts they don’t care about.

However, now the users typically have to construct a `KeyRingImportDelegate` and still fill it in with some predefined strategies. Your comment made me realize that we can really just use an interface for that and provide a bunch of premade strategies for the most common use cases. If that does not fit the bill then go ahead, implement the entire `KeyRingImportDelegate` interface the way you want it.

I’ll change this to an interface instead.

keystore/v2/keystore/filesystem/export.go
ilammy: > And here just call delegate.ImportExistingData or something like that. 

That would look a bit neater from our side but that will couple the delegate implementation to the key store implementation, without a particularly good reason. If we use enumeration values then the delegate can be reused with different key store implementations without any issues.

ilammy:
I've reviewed and updated the branch with the rest of key import implementation, and it turned out to be not that big. So I've just added it here instead of creating a new PR for the import specifically. @Lagovas, please give this PR another look.

The tests... well... they leave much to be desired in the conciseness department, but they do verify what they need to verify.

keystore/v2/keystore/filesystem/keyRingTX.go
```diff
@@ -133,6 +133,34 @@ func (tx *txAddKey) Rollback(ring *KeyRing) error {
...
+func (tx *txSetKeys) Apply(ring *KeyRing) error {
+	tx.oldKeys = ring.data.Keys
+	tx.oldCurrent = ring.data.Current
+	ring.data.Keys = tx.newKeys
+	ring.data.Current = tx.current
+	return nil
+}
+
+func (tx *txSetKeys) Rollback(ring *KeyRing) error {
+	if ring.data.Current != tx.oldCurrent {
```
Lagovas, line 153: is not it should be equal to `tx.current`? In `Apply` method we setup `current` for ring. In `Rollback` we should expect any other value from oldCurrent or just setup `oldCurrent`.
p.s. what means "OOO" in `errTxOOORollback`?)

vixentael: > p.s. what means "OOO" in errTxOOORollback?)

cat on keyboard?

ilammy: @Lagovas, I believe you’re right, that line is incorrect. If the current subkey does not change then we’ll get an error here.

> p.s. what means "OOO" in errTxOOORollback?)

Out-of-order.

> cat on keyboard?

I wish it were. The only cat I have is a ceramic one, it can definitely spam one keypress if it falls of the shelf onto the keyboard.

----

I’ve added these checks to verify that we do not roll back changes out-of-order, but this code path is not well tested (if tested at all). So I guess we can drop all these lines and always roll back to whatever values we think are correct ones?

Lagovas: then better to rename to full name `errTxOutOfOrderRollback`. Because i'm not sure that anyone of us will be able to repeat what it means after 1-3 months)

and yes, rollback should return state of something to state when it was at start. In this case we store state at start of the transaction and know which state it should take. And it doesn't need to be changed always on rollback. For example with databases, state of table can be unchanged but the transaction rollbacked by some other condition and it's okay. There the same, we don't need check that something changed.

PR #377, opened 2020-05-26
"acra-keys list" subcommand

ilammy:
Introduce a new subcommand `list` which could be used to enumerate all keys stored in the key store. This will be useful during key export to specify which keys need to be exported.

The basic usage looks like this:

```
$ acra-keys list
Key purpose                  | Client/Zone ID | Key ID
-----------------------------+----------------+-----------------------------------
client storage key           | ilammy         | client/ilammy/storage
AcraConnector transport key  | ilammy         | client/ilammy/transport/connector
AcraServer transport key     | ilammy         | client/ilammy/transport/server
AcraTranslator transport key | ilammy         | client/ilammy/transport/translator
```

Fully machine-readable output is also support with new `--json` option:

```
$ acra-keys --json list | jq
[
  {
    "ID": "client/ilammy/storage",
    "Purpose": "client storage key",
    "ClientID": "aWxhbW15"
  },
  {
    "ID": "client/ilammy/transport/connector",
    "Purpose": "AcraConnector transport key",
    "ClientID": "aWxhbW15"
  },
  {
    "ID": "client/ilammy/transport/server",
    "Purpose": "AcraServer transport key",
    "ClientID": "aWxhbW15"
  },
  {
    "ID": "client/ilammy/transport/translator",
    "Purpose": "AcraTranslator transport key",
    "ClientID": "aWxhbW15"
  }
]
```

(Note that client and zone IDs are actually binary so in JSON they are encoded in base64.)

The new command is supported only for key store v2. For older key stores acra-keys will print an error and exit:

```
ERRO[0000] "list" is not implemented for key store v1 in Acra Community Edition 
INFO[0000] You can convert key store v1 into v2 with "acra-migrate-keys" 
INFO[0000] Read more: https://docs.cossacklabs.com/pages/documentation-acra/#key-management 
FATA[0000] Failed to read key list                       error="not implemented"
```

### DIFF
--- a/cmd/acra-keys/keys/list-keys.go
+++ b/cmd/acra-keys/keys/list-keys.go
@@ -52,6 +52,7 @@ const (
 func printKeysTable(keys []keystore.KeyDescription, writer io.Writer) error {
 	maxPurposeLen := len(purposeHeader)
 	maxExtraIDLen := len(extraIDHeader)
+	maxKeyIDLen := len(idHeader)
 	for _, key := range keys {
 		if len(key.Purpose) > maxPurposeLen {
 			maxPurposeLen = len(key.Purpose)
@@ -62,12 +63,17 @@ func printKeysTable(keys []keystore.KeyDescription, writer io.Writer) error {
 		if len(key.ZoneID) > maxExtraIDLen {
 			maxExtraIDLen = len(key.ZoneID)
 		}
+		if len(key.ID) > maxKeyIDLen {
+			maxKeyIDLen = len(key.ID)
+		}
 	}
 
-	fmt.Fprintf(writer, "%-*s  %-*s  %s\n", maxPurposeLen, purposeHeader, maxExtraIDLen, extraIDHeader, idHeader)
+	fmt.Fprintf(writer, "%-*s | %-*s | %s\n", maxPurposeLen, purposeHeader, maxExtraIDLen, extraIDHeader, idHeader)
 
-	separator := make([]byte, maxPurposeLen+maxExtraIDLen+len(idHeader)+4)
+	separator := make([]byte, maxPurposeLen+maxExtraIDLen+maxKeyIDLen+6)
 	utils.FillSlice(byte('-'), separator)
+	separator[maxPurposeLen+1] = byte('+')
+	separator[maxPurposeLen+maxExtraIDLen+4] = byte('+')
 	fmt.Fprintln(writer, string(separator))
 
 	for _, key := range keys {
@@ -78,7 +84,7 @@ func printKeysTable(keys []keystore.KeyDescription, writer io.Writer) error {
 		if key.ZoneID != nil {
 			extraID = string(key.ZoneID)
 		}
-		fmt.Fprintf(writer, "%*s  %*s  %s\n", maxPurposeLen, key.Purpose, maxExtraIDLen, extraID, key.ID)
+		fmt.Fprintf(writer, "%-*s | %-*s | %s\n", maxPurposeLen, key.Purpose, maxExtraIDLen, extraID, key.ID)
 	}
 	return nil
 }
--- a/cmd/acra-keys/keys/list-keys.go
+++ b/cmd/acra-keys/keys/list-keys.go
@@ -17,6 +17,7 @@
 package keys
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 
@@ -26,7 +27,20 @@ import (
 
 // PrintKeys prints key list prettily into the given writer.
 func PrintKeys(keys []keystore.KeyDescription, writer io.Writer, params *CommandLineParams) error {
+	if params.UseJSON {
+		return printKeysJSON(keys, writer)
+	}
 	return printKeysTable(keys, writer)
+}
+
+func printKeysJSON(keys []keystore.KeyDescription, writer io.Writer) error {
+	json, err := json.Marshal(keys)
+	if err != nil {
+		return err
+	}
+	json = append(json, byte('\n'))
+	_, err = writer.Write(json)
+	return err
 }
 
 const (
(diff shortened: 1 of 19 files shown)


shadinua:
It’s actually great that the option `--json` is provided.

But I'd suggest to make a little improvement to basic output to make it easily parseable by simple shell tools. Undoubtedly, JSON is very useful with high level languages. But it is common to use such tools with primitive scripts.

Even if we don’t explicitly declare that basic plain format is parseable, I think it is good practice to do it such de-facto. While second and third columns may be easily splitted by spaces, the first one doesn't have any explicit markers.

As one of the options, we could easily add `:` to the end of each "Key purpose" column.

ilammy:
> Even if we don’t explicitly declare that the basic plain format is parseable, I think it is good practice to do it such de-facto.

Makes sense. Thank you for your input, that's a valuable consideration.

I thought about using tab-separated values as the format but it has some rendering issues so I opted for manually tabulating it in software with spaces. We may still use tabs along with spaces, but that's still a bit nonobvious.

Client IDs technically can contain spaces too ([`[a-zA-Z0-9 _-]`](https://github.com/cossacklabs/acra/blob/61d0b8a25baa043f90f1155e8056b6ea57a1ed6a/keystore/keystore.go#L68-L81)). Zone IDs normally are [`[a-zA-Z]`](https://github.com/cossacklabs/acra/blob/61d0b8a25baa043f90f1155e8056b6ea57a1ed6a/zone/utils.go#L29-L37) since we're generating them. But I guess we'd better use some separator there as well.

We can make it a fancier table, for example:

```
Key purpose                  | Client/Zone ID | Key ID
-----------------------------+----------------+-----------------------------------
client storage key           | ilammy         | client/ilammy/storage
AcraConnector transport key  | ilammy         | client/ilammy/transport/connector
AcraServer transport key     | ilammy         | client/ilammy/transport/server
AcraTranslator transport key | ilammy         | client/ilammy/transport/translator
```

How does that look, everyone?

Then it can be parsed with something like 

```
acra-keys list | tail +3 | awk -F ' *\\| *' '{print $1, "-", $2, "-", $3}'
```

(If all columns are left-justified then it's trivial to trim trailing spaces.)

shadinua:
Now it's quite simple to parse it. Offhand, here is one more way:
```
acra-keys list | tail -n+3 | cut -d'|' -f3 | tr -d ' '
```
But, of course, there are many ways to do this now. :)

Looks pretty nice. Great, thanks!

cmd/acra-keys/keys/list-keys.go
```diff
@@ -0,0 +1,90 @@
...
+	"io"
+
+	"github.com/cossacklabs/acra/keystore"
+	"github.com/cossacklabs/acra/utils"
+)
+
+// PrintKeys prints key list prettily into the given writer.
+func PrintKeys(keys []keystore.KeyDescription, writer io.Writer, params *CommandLineParams) error {
+	if params.UseJSON {
+		return printKeysJSON(keys, writer)
```
Lagovas, line 31: can we add test which will parse keys printed in json format? text format maybe hard to parse but json we can check and validate that it really machine readable.

keystore/v2/keystore/filesystem/keyStore.go
```diff
@@ -138,6 +139,34 @@ func (s *KeyStore) OpenKeyRingRW(path string) (api.MutableKeyRing, error) {
 	return ring, nil
 }
 
+// ListKeyRings enumerates all key rings present in this key store.
+func (s *KeyStore) ListKeyRings() (rings []string, err error) {
```
Lagovas, line 143: can we be sure that we return only paths to rings? that in this folder wasn't created an extra file which not related to keystore?

keystore/v2/keystore/keyStore.go
```diff
@@ -71,6 +92,81 @@ func NewTranslatorKeyStore(keyStore api.MutableKeyStore) *TranslatorKeyStore {
...
+		}, nil
+	}
+	if path == authKeyPath {
+		return &keystore.KeyDescription{
+			ID:      path,
+			Purpose: PurposeAuthentication,
+		}, nil
+	}
+
+	components := strings.Split(path, string(filepath.Separator))
```
Lagovas, line 126: can you add comment with example of path for each component? like '/<client_id>/<something_else>/<storage>'? because looking on it hard to understand which strings expected here
plus I forgot how we store our keys. Is it same tree structure of files in keystore v2 or it one file in asn1 format which contains binary tree structure inside?

keystore/v2/keystore/filesystem/keyStore.go
ilammy: > can we be sure that we return only paths to rings?

Right now: not so much. If the user adds some extra files then we'll pick them up, try parsing them, and fail either when listing keys (due to unexpected naming), or later during export (due to mismatching content).

There are ideas for improved integrity – key directories – which are cryptographically signed lists of key rings stored in a key store. Once that is implemented we can rely on the key directories to list the key rings and ignore (or report) any extra files.

We also store some bookkeeping files for locking and versioning, those are filtered out.

keystore/v2/keystore/keyStore.go
ilammy: > Is it same tree structure of files in keystore v2 or it one file in asn1 format which contains binary tree structure inside?

Each key ring is stored in an individual ASN.1 file. Directory structure mirrors the IDs, so if you have `client/lagovas/storage` then this keyring is stored in a file `${keystore}/client/lagovas/storage.keyring`.

cmd/acra-keys/keys/list-keys.go
ilammy: > can we add test which will parse keys printed in json format?

I've added a couple of tests for `PrintKeys`. For default format it simply verifies that the output is the expected one. For JSON we try to parse the output as JSON and check it successfully parses into the same data as the input.

PR #378, opened 2020-06-01
acra-keys export & import subcommands

ilammy:
This PR adds `export` and `import` subcommands that handle key export and import respectively.

#### Exporting keys

The keys can be exported like this:

```shell
$ acra-keys list
Key purpose                  | Client/Zone ID | Key ID
-----------------------------+----------------+-----------------------------------
client storage key           | ilammy         | client/ilammy/storage
AcraConnector transport key  | ilammy         | client/ilammy/transport/connector
AcraServer transport key     | ilammy         | client/ilammy/transport/server
AcraTranslator transport key | ilammy         | client/ilammy/transport/translator
```

```shell
$ acra-keys export --key_bundle_file encrypted-keys.dat --key_bundle_secret access-keys.json --all
INFO[0000] Exported key data is encrypted and saved here: encrypted-keys.dat 
INFO[0000] New encryption keys for import generated here: access-keys.json 
INFO[0000] DO NOT transport or store these files together 
INFO[0000] Import the keys into another key store like this:
	acra-keys import --key_bundle_file "encrypted-keys.dat" --key_bundle_secret "access-keys.json"
```

(Instead of `--all` you can enumerate specific key IDs from `acra-keys list` output.)

#### Export results

The `encrypted-keys.dat` file contains encrypted and signed key data:

```shell
$ hd encrypted-keys.dat | head -5
00000000  30 82 03 78 30 82 03 43  02 01 04 02 01 02 17 11  |0..x0..C........|
00000010  32 30 30 36 30 31 31 32  34 37 31 34 2b 30 33 30  |200601124714+030|
00000020  30 04 82 03 26 00 01 01  40 0c 00 00 00 10 00 00  |0...&...@.......|
00000030  00 fa 02 00 00 6c 4e e6  1c e5 30 56 bb 4e f7 cb  |.....lN...0V.N..|
00000040  cd 8d de 49 d2 de 4d c4  f8 ce 32 9d 2b b5 c7 3d  |...I..M...2.+..=|
```

The `access-keys.json` contains freshly generated keys used to encrypt and sign the exported data, in plaintext:

```shell
$ cat access-keys.json
{"encryption":"EIufOjLKGXrJRMKpwtYdvbIM1LSaOkYpnEmTVimxnO4=","signature":"upXcpYdj58rpEhX7b8Xnq8asr2718i/z/DlMXgMpojo="}
```

We do not give the user an option to use their own transport keys to encrypt the output. That way we are sure that they keys are never reused.

These two data pieces should be transported separately to the target machine and imported there.

#### Importing keys

The keys can be imported like this, as suggested during export:

```shell
$ acra-keys import --key_bundle_file "encrypted-keys.dat" --key_bundle_secret "access-keys.json"
INFO[0000] successfully imported 4 keys                 
Key purpose                  | Client/Zone ID | Key ID
-----------------------------+----------------+-----------------------------------
client storage key           | ilammy         | client/ilammy/storage
AcraConnector transport key  | ilammy         | client/ilammy/transport/connector
AcraServer transport key     | ilammy         | client/ilammy/transport/server
AcraTranslator transport key | ilammy         | client/ilammy/transport/translator
```

We list the affected keys for the benefit of the user.

Right now it is not possible to import the same data twice, it results in an error:

```shell
$ acra-keys import --key_bundle_file "encrypted-keys.dat" --key_bundle_secret "access-keys.json"
FATA[0000] Failed to import keys                         error="imported key ring already exists"
```

### Peculiarities and open questions

- Are the command-line option names and the overall interface good?

  ~~I'm concerned, for example, that the users might be confused by the `--keys` option.~~
  (Resolved after discussion.)

- Is it okay to save the data encryption keys into a file?

  It's convenient for later import, but that's an unencrypted file with keys after all. They are ephemeral keys, but nevertheless: if they are stored together with the exported data then it's like having the exported keys in plaintext. It's very easy to do the stupid thing and store them together.

### Incomplete features

The following features are currently missing and will be added later:

- It is not possible to export only public keys which is required during initial Acra setup. In fact, I'd rather have `acra-keys export` export only public keys by default, with private/symmetric key export being opt-in (e.g., for backups). I'm working on some `--with-private-keys` switch for that.

- It is not possible to import _updated_ keys (e.g., rotated ones) into existing key store which already has an older version. This feature is supported by the low-level API but the high-level CLI does not have an option for that yet. This will be added later.

- Only key store v2 is supported for both export and import. We could add v1 support in Acra EE on request, but given that v1 → v2 conversion is available and (should) have no downsides, it's okay to leave it out.

### DIFF
--- a/cmd/acra-keys/acra-keys.go
+++ b/cmd/acra-keys/acra-keys.go
@@ -39,11 +39,20 @@ func main() {
 	switch keys.Params.Command {
 	case keys.CmdListKeys:
 		listKeys(keys.Params)
+	case keys.CmdExportKeys:
+		exportKeys(keys.Params)
 	case keys.CmdReadKey:
 		printKey(keys.Params)
 	case keys.CmdDestroyKey:
 		destroyKey(keys.Params)
 	}
+}
+
+func warnKeystoreV2Only(command string) {
+	log.Error(fmt.Sprintf("\"%s\" is not implemented for key store v1 in Acra Community Edition", command))
+	log.Info("You can convert key store v1 into v2 with \"acra-migrate-keys\"")
+	// TODO(ilammy, 2020-05-19): production documentation does not describe migration yet
+	log.Info("Read more: https://docs.cossacklabs.com/pages/documentation-acra/#key-management")
 }
 
 func listKeys(params *keys.CommandLineParams) {
@@ -55,10 +64,7 @@ func listKeys(params *keys.CommandLineParams) {
 	keyDescriptions, err := keyStore.ListKeys()
 	if err != nil {
 		if err == keystore.ErrNotImplemented {
-			log.Error(fmt.Sprintf("\"%s\" is not implemented for key store v1 in Acra Community Edition", keys.CmdListKeys))
-			log.Info("You can convert key store v1 into v2 with \"acra-migrate-keys\"")
-			// TODO(ilammy, 2020-05-19): production documentation does not describe migration yet
-			log.Info("Read more: https://docs.cossacklabs.com/pages/documentation-acra/#key-management")
+			warnKeystoreV2Only(keys.CmdListKeys)
 		}
 		log.WithError(err).Fatal("Failed to read key list")
 	}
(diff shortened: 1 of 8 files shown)


shadinua:
The interface looks pretty clear.

> I'm concerned, for example, that the users might be confused by the --keys option.

It took some time to realize the naming approach. But after that it also seems understandable.
However, it may confuse an user because usually the `keys` argument means keys for business data. And here we swap concepts. Not sure if this will be better, but I’ll offer a couple of options how we can name that pair of arguments (business flow keys / keys for exported bundle):
* `--keys-bundle` / `--keys-bundle-access`
* `--keys` / `--keys-access`
* `--keys-archive` / `--keys-archive-access`

vixentael:
The naming confuses me as well, I’ve put some options into slack

ilammy:
After some discussions and brainstorming we have converged on the following renaming:

- `--data` → `--key_bundle_file` (the one with encrypted exported keys from the storage)
- `--keys` → `--key_bundle_secret` (new secret used to decrypt the exported key bundle)

PR #380, opened 2020-06-02
Getter for log formatting hooks

storojs72:
This PR introduces additional abstraction `AcraFormatter` that is necessary for refactoring (https://github.com/cossacklabs/acra-enterprise/pull/50#discussion_r429401067)

### DIFF
--- a/logging/log_formatters.go
+++ b/logging/log_formatters.go
@@ -27,12 +27,17 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+type AcraFormatter interface {
+	Formatter
+	GetHooks() []FormatterHook
+}
+
 // ---------- custom Loggers
 // inspired by "github.com/bshuster-repo/logrus-logstash-hook"
 // ----------
 
 // TextFormatter returns a default logrus.TextFormatter with specific settings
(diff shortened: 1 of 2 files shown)


logging/log_formatters.go
```diff
@@ -27,12 +27,17 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+type AcraFormatter interface {
+	Formatter
+	GetHooks() []FormatterHook
+}
```
ilammy, line 33: Hm... We already have our own [`Formatter` interface](https://github.com/cossacklabs/acra/blob/e6c6e00e832511601f938dd43e5edf7c0ac78843/logging/logging.go#L72-L79) in the **cop** branch. It extends `logrus.Formatter`. Since that interface allows to `SetHooks()`, I think it will be more natural to add `GetHooks()` there as well, avoiding the need to introduce an additional interface.

storojs72: Yeah. Agree. Thank you. Done in https://github.com/cossacklabs/acra/pull/380/commits/686c2f9a8fe7c408f646d6246d0c5e4b3efaad73

PR #383, opened 2020-06-24
Add docker-compose demo in transparent mode (T1114)

shmel-RT:
This PR add 8 new `transparent` demo compose files and fix tls_auth level problem in ssl demo


docker/acra-server_transparent.dockerfile
```diff
@@ -0,0 +1,14 @@
+# Create internal synonym for previuosly built image
+ARG DOCKER_REGISTRY_PATH
+ARG VCS_REF
+FROM cossacklabs/acra-server:latest
+
+VOLUME ["/keys"]
+EXPOSE 9090 9393
+
+COPY configs/acra-encryptor.yaml /acra-encryptor.yaml
```
Lagovas, line 9: where this file used? In `CMD` not used, in `ENTRYPOINT` too. I looked do Acra load config by some default path but it didn't and starts with empty path to config. So by default it will not be used. Do we can load this file in final file which depends on it, in docker-compose by volume definition or something like that?

docker/docker-compose.mysql-nossl-server-ssl-connector_transparent.yml
```diff
@@ -0,0 +1,169 @@
...
+            --client_id=${ACRA_CLIENT_ID:-testclientid}
+            --auth_keys=/keys/httpauth.accounts
+            --http_api_enable
+            --incoming_connection_api_string=tcp://0.0.0.0:9090
+            --config_file=/config/acra-server.yaml
+            --acraconnector_tls_transport_enable
+            --tls_ca=/ssl.ca/example.cossacklabs.com.crt
+            --tls_cert=/ssl.server/acra-server.crt
+            --tls_key=/ssl.server/acra-server.key
+            --acraconnector_transport_encryption_disable
```
Lagovas, line 121: this option is redundant and will not used. It works in next [manner](https://github.com/cossacklabs/acra/blob/master/cmd/acra-server/acra-server.go#L234):
* use tls if `acraconnector_tls_transport_enable` used
* don't encrypt anything if `acraconnector_transport_encryption_disable`
* use secure session otherwise

same for connector

docker/docker-compose.mysql-nossl-server-ssl-connector_transparent.yml
```diff
@@ -0,0 +1,169 @@
...
+            --client_id=${ACRA_CLIENT_ID:-testclientid}
+            --incoming_connection_string=tcp://0.0.0.0:9494
+            --http_api_enable
+            --incoming_connection_api_string=tcp://0.0.0.0:9191
+            --acraserver_tls_transport_enable
+            --tls_ca=/ssl.ca/example.cossacklabs.com.crt
+            --tls_cert=/ssl.proxy/acra-connector.crt
+            --tls_key=/ssl.proxy/acra-connector.key
+            --tls_acraserver_sni=acra-server
+            --acraserver_transport_encryption_disable
```
Lagovas, line 161: same here

docker/docker-compose.mysql-ssl-server-ssl-connector.yml
```diff
@@ -122,6 +122,7 @@ services:
             --tls_key=/ssl.server/acra-server.key
             --tls_db_sni=mysql
             --acraconnector_transport_encryption_disable
+            --tls_auth=1
```
Lagovas, line 125: why `tls_auth=1` instead 4?

docker/docker-compose.mysql-nossl-server-ssession-connector_transparent.yml
```diff
@@ -0,0 +1,191 @@
...
+
+    # MySQL container
+    mysql:
+        image: mariadb:10.3
+        # INSECURE!!! You MUST define your own DB name and credentials
+        environment:
+            MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PASSWORD:-hy53uBLmBuihUK29}
+            MYSQL_DATABASE: ${MYSQL_DATABASE:-test}
+            MYSQL_USER: ${MYSQL_USER:-test}
+            MYSQL_PASSWORD: ${MYSQL_PASSWORD:-test}
```
ilammy, line 84: Is there any good secure default that we can use so that we don't have insecure-by-default examples? Like, load the credentials from a file or whatever.

Also, why exactly is this insecure when having Acra master key in the same file is apparently secure? Well, in some other places we *do have* warnings stating that using our hardcoded and published master keys is also insecure.

I'd suggest to change these legitimately looking base64 strings into something prominent, like `NOT-FOR-PRODUCTION-PLEASE-CHANGE` for the avoidance of doubt.

docker/docker-compose.mysql-ssl-server-ssl-connector.yml
ilammy: This why I don't really like numeric options for enumerations: it's not immediately obvious that 4 is an option, and what exactly it does.

When you're doing a code review of configuration changes, you either have to remember it, or be diligent enough to find the source code of the tool, find out that this number refers to [`ClientAuthType`](https://golang.org/pkg/crypto/tls/#ClientAuthType), and then you might think whether `RequireAndVerifyClientCert` is a better choice than `RequestClientCert`.

docker/docker-compose.mysql-nossl-server-ssession-connector_transparent.yml
```diff
@@ -0,0 +1,191 @@
...
+        volumes:
+            - ./.acrakeys:/keys
+        command: >-
+            --client_id=${ACRA_CLIENT_ID:-testclientid}
+            --generate_acraconnector_keys
+            --keys_output_dir=/keys/acra-connector
+            --keys_public_output_dir=/keys/acra-server
+    # Create keys:
+    # - ./.acrakeys/acra-server/${ACRA_CLIENT_ID}_storage
+    # - ./.acrakeys/acra-writer/${ACRA_CLIENT_ID}_storage.pub
```
shadinua, line 45:
```suggestion
    # - ./.acrakeys/acra-server/${ACRA_CLIENT_ID}_storage.pub
```

docker/docker-compose.mysql-nossl-server-ssession-connector_transparent.yml
shadinua: The primary goals of this set of docker demos are:
* quick start with one command
* show different behaviours
* show main features
* to be illustrative and understandable
* to show relations between components

We cannot make such easy one-command demos _really_ secure, because it is not only about one docker-compose configuration. It should be a set of approaches in this configuration, in the applciation, and in the infrastructure design.

But each step to make it a bit secure will make it much less readable and much more complex.

Apparently, secure infrastructure and application design is out of scope for this simple demos. So let's write with big red letters obvious warnings and will not became too dispersed.

docker/docker-compose.mysql-nossl-server-ssl-connector_transparent.yml
```diff
@@ -0,0 +1,169 @@
...
+        volumes:
+            - ./.acrakeys:/keys
+        command: >-
+            --client_id=${ACRA_CLIENT_ID:-testclientid}
+            --generate_acraconnector_keys
+            --keys_output_dir=/keys/acra-connector
+            --keys_public_output_dir=/keys/acra-server
+    # Create keys:
+    # - ./.acrakeys/acra-server/${ACRA_CLIENT_ID}_storage
+    # - ./.acrakeys/acra-writer/${ACRA_CLIENT_ID}_storage.pub
```
shadinua, line 45:
```suggestion
    # - ./.acrakeys/acra-server/${ACRA_CLIENT_ID}_storage.pub
```

docker/docker-compose.mysql-ssl-server-ssl-connector_transparent.yml
```diff
@@ -0,0 +1,175 @@
...
+        volumes:
+            - ./.acrakeys:/keys
+        command: >-
+            --client_id=${ACRA_CLIENT_ID:-testclientid}
+            --generate_acraconnector_keys
+            --keys_output_dir=/keys/acra-connector
+            --keys_public_output_dir=/keys/acra-server
+    # Create keys:
+    # - ./.acrakeys/acra-server/${ACRA_CLIENT_ID}_storage
+    # - ./.acrakeys/acra-writer/${ACRA_CLIENT_ID}_storage.pub
```
shadinua, line 45:
```suggestion
    # - ./.acrakeys/acra-server/${ACRA_CLIENT_ID}_storage.pub
```

docker/docker-compose.mysql-ssl-server-ssl-connector.yml
shadinua: > why tls_auth=1 instead 4?

Completely agree. While we may use `tls_auth=1` for TLS down to end application _in bounds of these simple demos_, to avoid of additional complex configuration of a test application, there are no reasons to make less secure connection between AcraServer and AcraConnector.

docker/docker-compose.mysql-nossl-server-ssl-connector_transparent.yml
shmel-RT: I see. Going to change this in next commits

docker/acra-server_transparent.dockerfile
shmel-RT: The **docker way** telling us to use layers to add something to container instead of creating more volume entrances for each file that we need in container

Lagovas: okay, but it's not docker container for frequent use by many users. it's demo dockerfile which used for several demo docker-compose files. Why I admit it. I looked on file `docker/docker-compose.mysql-nossl-server-ssl-connector_transparent.yml` and tried to see how it configured transparent mode or is it configured. I see next arguments for `acra-server` section:
```
command: >-
...
--encryptor_config_file=/acra-encryptor.yaml
```

And try to find this find in context of docker-compose and didn't find where it declared or generated. And it force me to search other entries for files. And it require to open related dockerfile and read it. Instead find it in context of docker-compose file in `volume` section where we add ssl certificates and all related configs with specific installation (with/out ssl, mysql/postgresql/etc).

@shadinua , what you think about it? How will be better to read new readers such configurations?

docker/docker-compose.mysql-ssl-server-ssl-connector_transparent.yml
```diff
@@ -123,10 +123,8 @@ services:
             --tls_cert=/ssl.server/acra-server.crt
             --tls_key=/ssl.server/acra-server.key
             --tls_db_sni=mysql
-            --acraconnector_transport_encryption_disable
             --encryptor_config_file=/acra-encryptor.yaml
-            --tls_auth=1
-            -v
+            -d
```
shadinua, line 127: Why do we need to start it as a daemon instead of no-detach mode, which is expectable docker-way?

docker/test-app.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
```
shadinua, line 1: 1. What are the main goals for this script? Why can't we use AMT, for example?

2. Let's not put scripts in the root of docker directory structure — it's already becoming too large and requires to be reorganized.

We have `docker/_scripts` and `docker/ssl`, I'd suggest to:
- rename `docker/_scripts` → `docker/scripts` (please don't forget to change paths inside scripts)
- create the `docker/tests` subdirectory and put the file there

shmel-RT: AMT is a little "overpowered' for that. this script is like simple little client application  that make read write requests to DB with defined encryption config

docker/docker-compose.mysql-ssl-server-ssl-connector_transparent.yml
shmel-RT: This parameter stands for debug mode in acra-server and just an development artifact, thx for noticing, will fix it

docker/acra-server_transparent.dockerfile
shadinua: It looks like both approaches is reasonable for some cases.

But I'd agree with @Lagovas: since we design demo, here we should shift the balance to clarity. Anyway, if there are we have Acra configuration file inside the repository, it would be better to simply mount it to container and explicitly specify in argument list.

shadinua:
```
$ docker-compose -f ./docker/docker-compose.mysql-ssl-server-ssl_transparent.yml up

...

acra-server_1           | time="2020-07-09T14:37:35Z" level=info msg="Load encryptor configuration from /acra-encryptor.yaml ..."
acra-server_1           | time="2020-07-09T14:37:35Z" level=error msg="Can't read config for encryptor" code=507 error="open /acra-encryptor.yaml: no such file or directory"
acra-server_1           | time="2020-07-09T14:37:35Z" level=error msg="Can't load encryptor config" error="open /acra-encryptor.yaml: no such file or directory"
mysql_1                 | 2020-07-09 14:37:35+00:00 [Note] [Entrypoint]: Initializing database files
docker_acra-server_1 exited with code 1
docker_acra-server_1 exited with code 1
docker_acra-server_1 exited with code 1
```

docker/acra-server_transparent.dockerfile
shmel-RT: Ok. Going to change this setup in next commit, tnx for detailed response!

vixentael:
This definitely needs more work and merge changes from master. I close this PR. Let's open new one.

PR #386, opened 2020-07-13
Improve acra-migrate-key reusability

ilammy:
* **Common acra-migrate-keys package**

Move command-line processing into a shared `migratekeys` package which could be reused in Acra EE to provide the same command-line interface as Acra CE without a lot of copypasta.

We'll still need to use copy-paste inheritance for key store initialization and the migration process, but that's a necessary evil
because the key store instances are different. Common command-line parameters are the same, on the other hand.

* **Improve reusability of `filesystem.KeyExport`**

The main difference between Acra CE and Acra EE in key exporting is the list of keys that need to be exported and the method by which they are exported. Acra EE will need to extend and/or override these methods.

The exporting approach can be readily overridden in `ExportPublicKey()` methods, etc. That will be easy to do. However, the key list is different. Current Acra CE code does not expose the key list itself, instead it provides the `EnumerateExportedKeys()` method which queries the list and processes it using the classifier to determine which key paths correspond to which keys.

The classification process itself is the same for both Acra CE and EE, but in the current form it cannot be reused in Acra EE: it is too coupled with path enumertion. Acre CE supports only raw filesystem, but Acra EE supports more options and cannot use `ioutil.ReadDir()` directly.

Introduce a new interface for enumerating the key paths alone, and move the existing `EnumerateExportedKeys()` methods into free functions. This way Acra EE can provide its own path enumeration, its own extended classifier, but reuse the rest of the export workflow.

### DIFF
--- a/cmd/acra-migrate-keys/migrate-keys/command-line.go
+++ b/cmd/acra-migrate-keys/migrate-keys/command-line.go
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2020, Cossack Labs Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package migratekeys implements common procedures for "acra-migrate-keys" tool.
+package migratekeys
+
+import (
+	"flag"
+
+	keystoreV1 "github.com/cossacklabs/acra/keystore"
+	"github.com/cossacklabs/acra/utils"
+	log "github.com/sirupsen/logrus"
+)
+
+var (
+	defaultConfigPath = utils.GetConfigPathByName("acra-migrate-keys")
+	serviceName       = "acra-migrate-keys"
+
+	defaultSrcDir       = keystoreV1.DefaultKeyDirShort
+	defaultDstDirSuffix = ".migrated"
+)
+
+// CommandLineParams - command-line options.
+type CommandLineParams struct {
+	Src, Dst KeyStoreParams
+	Misc     MiscParams
+}
+
+// KeyStoreParams - key store parameters.
+type KeyStoreParams struct {
+	KeyStoreVersion string
+	KeyDir          string
+	KeyDirPublic    string
+}
+
+// MiscParams - miscellaneous parameters.
+type MiscParams struct {
+	DryRun bool
+	Force  bool
+
+	LogDebug   bool
+	LogVerbose bool
+}
+
+// OpenMode - whether key store is source or destination.
+type OpenMode int
+
+// OpenMode constant values:
+const (
+	OpenSrc OpenMode = iota
+	OpenDst
+)
+
+// Params stores command-line parameters of "acra-migrate-keys" tool.
+var Params CommandLineParams
+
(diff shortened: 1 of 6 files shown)


cmd/acra-migrate-keys/migrate-keys/command-line.go
```diff
@@ -0,0 +1,104 @@
...
+	KeyDir          string
+	KeyDirPublic    string
+}
+
+// MiscParams - miscellaneous parameters.
+type MiscParams struct {
+	DryRun bool
+	Force  bool
+
+	LogDebug   bool
```
Lagovas, line 54: maybe better to use LogLevel as int instead several flags for each level?

ilammy: Hm... Okay, let's see what can be done while keeping the CLI the same. I think it may be possible by introducing several unexported fields for command-line parameters, then setting the minimum log level based on the parameters.

ilammy:
I've added a bunch of changes which make the key store version required arguments. Now the user has to specify `--src_keystore` and `--dst_keystore` arguments with explicit versions. (The `--src_keys_dir` and `--dst_keys_dir` still default to `.acrakeys` and `.acrakeys.migrated` respectively.)

There are also some minor changes in the common command-line parsing module.

cmd/acra-migrate-keys/migrate-keys/command-line.go
```diff
@@ -70,35 +78,54 @@ var Params CommandLineParams
...
+	}
+	if params.Src.KeyStoreVersion == "" || params.Dst.KeyStoreVersion == "" {
+		return ErrMissingFormat
 	}
 
+	if params.Misc.logDebug {
+		params.Misc.LogLevel = log.DebugLevel
+	}
+	if params.Misc.logVerbose {
+		params.Misc.LogLevel = log.TraceLevel
```
Lagovas, line 119: in other our tools we set InfoLevel for `-v` flag and `DebugLevel` for `-d` flag. In most cases and other tools debug level provide more information than "verbose", because debug level it's for developers and verbose for end users. In logrus TraceLevel is most verbose level. If we want to provide turn on TraceLevel for developers then maybe do it by `-d` flag?

p.s. plus when I first looked on this PR, I thought to change debug+verbose flags to one "level" flag because thought that MiscParams and these flags will be used in code with some if/else branches. And only now when you provided both versions I see, that we always use -d/-v and these flags in `MiscParams` used once only when configure logging at start. And it's okay previous version )

ilammy: Hm... Right... I've just reviewed the tools: as for log levels, `-v` enables Info logs, and `-d` enables Debug logs. This is typically performed only after the initial startup, so by default only Warning and higher are printed.

In the context of `acra-migrate-keys`, Info is used for regular output of status messages. So it seems I thought that mapping Trace to verbose makes more sense. The trace level is much more verbose as key store implementations starts spewing out lots of traces, while on Debug level normally there are only a couple of status reports and more detailed information about errors (if they arise).

However, it is true that the verbose level does not make much sense to users, it's only useful for developers and debugging.

Well, it seems to be harmless so I guess we can leave it as is. Any objections? @shadinua?

Lagovas: imho, we have 2 ways. change level usage in code (trace -> debug, debug -> info) to be consistent with other code (I don't remember where we use TraceLevel in other our components), or just change mapping of flags to logLeve and configure TraceLevel for `-d` flag and DebugLevel for `-v` flag, to be consistent with other our tools by cli.

ilammy: Alrighty. I've reverted the change which exposes logrus log level after parameters parsing (not really useful since no one is currently interested in that option). I have also swapped the `-d` and `-v` flags. Now `-v` is somewhat useful to users (outputs Debug messages which can be used to diagnose key store access errors). The `-d` flag goes full verbose on all minor tasks that the key store does: open this key, try decrypting that, process that operation, etc.

PR #393, opened 2020-07-29
Add non-exiting signal handler

storojs72:
This PR just adds new utility function `RegisterWithoutExiting` to system signal handler that behaves similar to `Register` but it does not call `os.Exit(0)` in the end. As we discussed, it is desirable to have better control on points of exiting from the Acra services and ideally to minimize them to the single place of exiting. `Register` is still in place, so if some of the services rely on it's `os.Exit(0)` it will not break their behavior.

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -85,6 +85,11 @@ func (handler *SignalHandler) AddCallback(callback SignalCallback) {
 
 // Register should be called as goroutine
 func (handler *SignalHandler) Register() {
+	handler.RegisterWithoutExiting()
+	os.Exit(0)
+}
+
+func (handler *SignalHandler) RegisterWithoutExiting() {
 	signal.Notify(handler.ch, handler.signals...)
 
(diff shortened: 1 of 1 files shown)


cmd/utils.go
```diff
@@ -85,6 +85,11 @@ func (handler *SignalHandler) AddCallback(callback SignalCallback) {
 
 // Register should be called as goroutine
 func (handler *SignalHandler) Register() {
+	handler.RegisterWithoutExiting()
+	os.Exit(0)
+}
+
+func (handler *SignalHandler) RegisterWithoutExiting() {
```
ilammy, line 92: The linter is going to complain about missing documentation comment.

storojs72: Added missing comment in https://github.com/cossacklabs/acra/pull/393/commits/9194a4c48e5a4d754c0a750034d92487515603db

storojs72:
I think we can merge it now and return to naming later

PR #395, opened 2020-07-30
Move acra-keys parameters into interfaces

ilammy:
Another issue with `acra-keys` interfaces for Acra EE is that it uses `CommandLineParameters` struct directly. This means that Acra EE is forced to use Acra CE command-line parameter set and cannot add anything else there. This is problematic because Acra EE *needs* to add more key store configuration parameters.

Instead of accepting `CommandLineParameters` and accessing its fields directly, introduce a bunch of interface with accessors, delimiting options for various commands. This decouples business logic of `acra-keys` from concrete storage of parameters (and unrelated parameters). It will also allow further refactoring of `CommandLineParameters` – I'd like to split it into multiple structs later to provide a better command-line help and key store parameter CLI.

This is only implementation refactoring. Actual CLI stays the same.

Dependencies:

- [x] #394 Drop KeyStoreFactory (pending merge and rebase)

### DIFF
--- a/cmd/acra-keys/keys/acra-keys.go
+++ b/cmd/acra-keys/keys/acra-keys.go
@@ -97,7 +97,7 @@ func ListKeysCommand(params ListKeysParams, keyStore keystore.ServerKeyStore) {
 }
 
 // ExportKeysCommand implements the "export" command.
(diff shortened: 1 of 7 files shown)


ilammy:
Rebased after dependency merge without change. Will merge once CI passes.

PR #396, opened 2020-08-04
Decouple acra-keys subcommands

ilammy:
This PR refactors `acra-keys` overall architecture, making it more extensible, replacing concrete Acra CE types with interfaces, and reducing code concentration.

The overall idea is to have each subcommand to define its own name, registered parameters, and business logic (which includes constructing a key store and doing the actual business). That way Acra EE will be able to extend only those parts it needs – argument parsing and key store construction – while reusing everything else from Acra CE.

Additionally, key store configuration is moved into individual subcommands, allowing them to control the command-line parameters. This will be particularly important for supporting the `migrate-keys` subcommand which requires *two* sets of key store parameters.

### CLI changes

This PR changes the command-line interface of `acra-keys`. Now previously common options  like `--keys_dir` are no longer global, instead they are options for individual subcommands. That is, instead of writing

    acra-keys --keys_dir=my-acrakeys export --all

you should now write

    acra-keys export --keys_dir=my-acrakeys --all

passing options to the subcommand, not `acra-keys` itself.

### Implementation notes

Many subcommands have shared parameter sets. To avoid duplication those are grouped into `Common...Parameters` mix-ins which can be shared between subcommand implementations.

Global utility `cmd.Parse()` is extended with `cmd.ParseFlagsWithConfig()` with can work with provided `FlagSet` instead of global `flag.CommandLine`.


cmd/acra-keys/keys/keystore.go
```diff
@@ -39,6 +40,39 @@ type KeyStoreParameters interface {
...
+	return p.keyDirPublic
+}
+
+// Register registers key store flags with the given flag set.
+func (p *CommonKeyStoreParameters) Register(flags *flag.FlagSet) {
+	p.RegisterPrefixed(flags, "", "")
+}
+
+// RegisterPrefixed registers key store flags with the given flag set, using given prefix and description.
+func (p *CommonKeyStoreParameters) RegisterPrefixed(flags *flag.FlagSet, flagPrefix, desciptionSuffix string) {
```
vixentael, line 68:
```suggestion
func (p *CommonKeyStoreParameters) RegisterPrefixed(flags *flag.FlagSet, flagPrefix, descriptionSuffix string) {
```

cmd/acra-keys/keys/keystore.go
```diff
@@ -39,6 +40,39 @@ type KeyStoreParameters interface {
...
+	if desciptionSuffix != "" {
```
vixentael, line 69:
```suggestion
	if descriptionSuffix != "" {
```

cmd/acra-keys/keys/keystore.go
```diff
@@ -39,6 +40,39 @@ type KeyStoreParameters interface {
...
+		desciptionSuffix = " " + desciptionSuffix
```
vixentael, line 70:
```suggestion
		descriptionSuffix = " " + descriptionSuffix
```

cmd/acra-keys/keys/keystore.go
```diff
@@ -39,6 +40,39 @@ type KeyStoreParameters interface {
...
+	}
+	flags.StringVar(&p.keyDir, flagPrefix+"keys_dir", DefaultKeyDirectory, "path to key directory"+desciptionSuffix)
```
vixentael, line 72:
```suggestion
	flags.StringVar(&p.keyDir, flagPrefix+"keys_dir", DefaultKeyDirectory, "path to key directory"+descriptionSuffix)
```

cmd/acra-keys/keys/keystore.go
```diff
@@ -39,6 +40,39 @@ type KeyStoreParameters interface {
...
+	flags.StringVar(&p.keyDirPublic, flagPrefix+"keys_dir_public", "", "path to key directory for public keys"+desciptionSuffix)
```
vixentael, line 73:
```suggestion
	flags.StringVar(&p.keyDirPublic, flagPrefix+"keys_dir_public", "", "path to key directory for public keys"+descriptionSuffix)
```

configs/acra-keys.yaml
```diff
@@ -1,7 +1,4 @@
 version: 0.85.0
-# client ID for which to retrieve key
-client_id:
```
vixentael, line 3: has this moved somewhere or we just got rid of it?

cmd/acra-keys/keys/keystore.go
ilammy: O, autocomplete! Misspell once – write like that everywhere.

configs/acra-keys.yaml
ilammy: > has this moved somewhere or we just got rid of it?

It's still there in a sense that you can still, for example, add `keys_dir` to the configuration file so that you don't have to remember to add it to each individual subcommand. However, the handling of `--dump_config` currently takes into account only the global options. All interesting options have moved to subcommands, so as a result the template for configuration file became a bit barren.

I'm still thinking about what can be done for this. We can't just update the template manually because the tests verify that `--dump_config` outputs exactly this config. Making `acra-keys` an exception is not pretty. We need some way to compose options from multiple subcommands, deduplicate them, output all options into a configuration file, and make sure that – when used in the config – options of one subcommand do not break other subcommands.

vixentael: > We need some way to compose options from multiple subcommands

if this is complicated / long / unclear -> goto2; break;

goto2: make `acra-keys` an exception.

ilammy: Well, I have an idea for this, but it will require some additional refactoring in `--dump_config` handling, to make it possible to construct a configuration from multiple flag sets. I'd like to make it in a different PR and settle for an incomplete config for now.

vixentael: awesome, thank you!

PR #397, opened 2020-08-05
Full config for acra-keys

ilammy:
As noted by @vixentael in #396, `acra-keys` currently has incomplete configuration file template. The configuration still works, but the users won't know that they can set a particular value in the configuration file instead of passing it over and over on the command line.

The issue here is that the `DumpConfig` function used to handle the `--dump_config` flag is very tightly coupled with the global flag set (`flag.CommandLine`). However, `acra-keys` has most of its flags in flags sets of individual subcommands, not the global one.

To fix this issue, introduce a bunch of sister functions to `DumpConfig` and friends, which can accept an arbitrary list of flag sets to work with. `acra-keys` will use these functions to pass its flag sets. Other tools continue using the old functions which keep their behavior.

Also, `--dump_config` flag causes the process to immediately exit after dumping the config. This is an okay thing to do in `cmd.Parse` which works with only with the global flag set, but if we work with multiple flag sets, we would like to decide which flag sets we include. Make `ParseFlagsWithConfig` to return a special error (like `flag.ErrHelp`) when `--dump_config` is requested, and handle it the way we want to. The old `Parse` will still exit as before.

In the end, this allows the `acra-keys.yaml` to be properly populated with all available configuration options.

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -181,13 +181,30 @@ func PrintFlags(flags *flag_.FlagSet) {
 	})
 }
 
+func visitFlagSets(flagSets []*flag_.FlagSet, visit func(flag *flag_.Flag)) {
+	seenNames := make(map[string]struct{})
+	for _, flags := range flagSets {
+		flags.VisitAll(func(flag *flag_.Flag) {
+			if _, visited := seenNames[flag.Name]; !visited {
+				visit(flag)
+				seenNames[flag.Name] = struct{}{}
+			}
+		})
+	}
+}
+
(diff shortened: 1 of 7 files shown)


cmd/utils.go
```diff
@@ -176,13 +181,30 @@ func PrintFlags(flags *flag_.FlagSet) {
 	})
 }
 
+func visitFlagSets(flagSets []*flag_.FlagSet, visit func(flag *flag_.Flag)) {
```
vixentael, line 184: neat

PR #398, opened 2020-08-10
MySQL decryption notifications

ilammy:
This is a sizeable but coherent changeset improving various aspects of MySQL processing in Acra CE, with the goal of providing a better platform for Acra EE features relying on additional data processing, such as pseudonymization.

The overarching idea is that PostgreSQL proxy backend implements the `ColumnDecryptionNotifier` interface which enables subscribers to observe and modify per-column data during is processing in responses and requests. MySQL does not support this interface. Moreover, there are some other issues related to its use which are fixed up here to keep the build green.

- Integer support in MysqlDBDataCoder

First of all MySQL request parser did not support integer fields in requests. This breaks any tests for MySQL using integers because Acra simply passes such requests through unchanged. Integers are represented with textual strings, so they are easy to add.

- Common ColumnDecryptionNotifier implementation
- Use common observer in PostgreSQL decryptor
- Implement ColumnDecryptionNotifier for MySQL proxy

The bulk of `ColumnDecryptionNotifier` implementation is completely independent of the database engine and its protocol. These commits introduce a shared utility that makes it trivial to implement the notification interface. The utility is based on the current PostgreSQL code, with some improvements.

- Observe decryption events for text data rows
- Observe decryption events for binary rows: decryptable strings
- Observe decryption events for binary rows: other strings
- Observe decryption events for binary rows: fixed-size numbers

These commits add calls to `OnColumnDecrypted` notifications, letting the subscribers know that a column is being decrypted. MySQL protocol is tricky because there are several representations of data, including text and binary. The subscribers expect the data to come as text, so we need to convert binary data into text, let the subscribers process it, and then reencode the result back.

This results is somewhat awful copypasta. If anyone has suggestions on how to reduce it, please voice them. I don't really like it.

- Require DecryptionSubscribers to ID themselves

This commit incorporates some extensions made by Acra EE, allowing to reduce code duplication in Acra EE, and to avoid some subtle issues related to context handling.

- Always update data after processing
- Special handling of NULL value encoding
- Propagate DataProcessor in mysql.Decryptor

Adding the new processing flow requires some additional attention to the data, as it needs to be returned into MySQL packet in the correct format. See commit messages for more details.

### DIFF
--- a/decryptor/base/decryptionNotification.go
+++ b/decryptor/base/decryptionNotification.go
@@ -103,3 +103,79 @@ type ColumnDecryptionNotifier interface {
 	SubscribeOnAllColumnsDecryption(subscriber DecryptionSubscriber)
 	Unsubscribe(DecryptionSubscriber)
 }
+
+// ColumnDecryptionObserver is a simple ColumnDecryptionNotifier implementation.
+type ColumnDecryptionObserver struct {
+	perColumn  map[int][]DecryptionSubscriber
+	allColumns []DecryptionSubscriber
+}
+
+// NewColumnDecryptionObserver makes a new observer.
+func NewColumnDecryptionObserver() ColumnDecryptionObserver {
+	// Reserve some memory for a typical amount of subscribers.
+	return ColumnDecryptionObserver{
+		perColumn:  make(map[int][]DecryptionSubscriber, 10),
+		allColumns: make([]DecryptionSubscriber, 0, 5),
+	}
+}
+
+// SubscribeOnColumnDecryption subscribes for notifications about the column, indexed from left to right starting with zero.
+func (o *ColumnDecryptionObserver) SubscribeOnColumnDecryption(column int, subscriber DecryptionSubscriber) {
+	subscribers := o.perColumn[column]
+	for _, existing := range subscribers {
+		if existing == subscriber {
+			return
+		}
+	}
+	o.perColumn[column] = append(subscribers, subscriber)
+}
+
+// SubscribeOnAllColumnsDecryption subscribes for notifications on each column.
+func (o *ColumnDecryptionObserver) SubscribeOnAllColumnsDecryption(subscriber DecryptionSubscriber) {
+	for _, existing := range o.allColumns {
+		if existing == subscriber {
+			return
+		}
+	}
+	o.allColumns = append(o.allColumns, subscriber)
+}
+
+// Unsubscribe a subscriber from all notifications.
+func (o *ColumnDecryptionObserver) Unsubscribe(subscriber DecryptionSubscriber) {
+	for column, observers := range o.perColumn {
+		for i, existing := range observers {
+			if existing == subscriber {
+				o.perColumn[column] = append(observers[:i], observers[i+1:]...)
+				break
+			}
+		}
+	}
+	for i, existing := range o.allColumns {
+		if existing == subscriber {
+			o.allColumns = append(o.allColumns[:i], o.allColumns[i+1:]...)
+			break
+		}
+	}
+}
+
+// OnColumnDecryption notifies all subscribers about a change in given column, passing the context and data to them.
+// Returns the data and error returned by subscribers.
+// If a subscriber returns an error, it is immediately returned and other subscribers are not notified.
+func (o *ColumnDecryptionObserver) OnColumnDecryption(ctx context.Context, column int, data []byte) ([]byte, error) {
+	var err error
+	// Avoid creating a map entry if it does not exist.
+	subscribers, _ := o.perColumn[column]
+	for _, subscriber := range subscribers {
+		ctx, data, err = subscriber.OnColumn(ctx, data)
+		if err != nil {
+			return data, err
+		}
+	}
+	for _, subscriber := range o.allColumns {
+		ctx, data, err = subscriber.OnColumn(ctx, data)
+		if err != nil {
+			return data, err
+		}
+	}
+	return data, nil
+}
--- a/decryptor/base/decryptionNotification.go
+++ b/decryptor/base/decryptionNotification.go
@@ -1,6 +1,10 @@
 package base
 
-import "context"
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+)
 
 // ColumnInfo interface describe available metadata for column
 type ColumnInfo interface {
@@ -95,6 +99,7 @@ func ColumnInfoFromContext(ctx context.Context) (ColumnInfo, bool) {
 // DecryptionSubscriber interface to subscribe on column's data in db responses
 type DecryptionSubscriber interface {
 	OnColumn(context.Context, []byte) (context.Context, []byte, error)
+	ID() string
 }
 
 // ColumnDecryptionNotifier interface to subscribe/unsubscribe on OnColumn events
@@ -168,12 +173,14 @@ func (o *ColumnDecryptionObserver) OnColumnDecryption(ctx context.Context, colum
 	for _, subscriber := range subscribers {
 		ctx, data, err = subscriber.OnColumn(ctx, data)
 		if err != nil {
+			logrus.WithField("subscriber", subscriber.ID()).WithError(err).Errorln("OnColumn error")
 			return data, err
 		}
 	}
 	for _, subscriber := range o.allColumns {
 		ctx, data, err = subscriber.OnColumn(ctx, data)
 		if err != nil {
+			logrus.WithField("subscriber", subscriber.ID()).WithError(err).Errorln("OnColumn error")
 			return data, err
 		}
 	}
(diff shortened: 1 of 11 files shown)


ilammy:
@Lagovas, I know you're not really available for review at the moment, but I would like you to give a look to this PR later, possibly after it has been merged. You know this codebase better than anyone else in the world, after all.

@iamnotacake, I've added you here just for awareness but constructive comments and WTFs are welcome ;)

decryptor/mysql/response_proxy.go
```diff
@@ -477,64 +514,230 @@ func (handler *Handler) processBinaryDataRow(ctx context.Context, rowData []byte
...
 			continue
+
 		default:
 			return nil, fmt.Errorf("found unknown FieldType <type=%d> <name=%s> in MySQL response packet", fields[i].Type, fields[i].Name)
 		}
 	}
 	return output, nil
 }
 
+func (handler *Handler) processFixedSizeNumberField(ctx context.Context, columnIndex int, column *ColumnDescription, encoded []byte) ([]byte, error) {
```
vixentael, line 601: that's a lot of going on in this method..

decryptor/mysql/utils.go
```diff
@@ -132,6 +133,10 @@ func PutLengthEncodedInt(n uint64) []byte {
 
 // PutLengthEncodedString https://dev.mysql.com/doc/internals/en/string.html#packet-Protocol::LengthEncodedString
 func PutLengthEncodedString(b []byte) []byte {
+	// Special case for NULL encoding, represented by "nil" values in Go.
+	if b == nil {
```
vixentael, line 137: good catch! 👍

decryptor/mysql/response_proxy.go
```diff
@@ -386,30 +414,33 @@ func (handler *Handler) processTextDataRow(ctx context.Context, rowData []byte,
 	handler.logger.Debugln("Process data rows in text protocol")
 	for i := range fields {
 		fieldLogger = handler.logger.WithField("field_index", i)
-		value, _, n, err = LengthEncodedString(rowData[pos:])
+		value, n, err = LengthEncodedString(rowData[pos:])
 		if err != nil {
 			return nil, err
 		}
+		value, err = handler.onColumnDecryption(ctx, i, value)
```
Lagovas, line 421: we try to decrypt here and below in `if handler.isFieldToDecrypt` branch? Maybe better to remove `if` check, and leave base logic with `onColumnDecryption`. And implement required interface by `handler` which will subscribe on `OnColumnDecryption` and do logic with `if handler.isFieldToDecrypt(fields[i])`?

We can encapsulate 
```
if handler.isFieldToDecrypt(fields[i]) {
			decryptedValue, err := handler.decryptor.DecryptBlock(value)
			if err == nil && decryptedValue != nil && len(decryptedValue) != len(value) {
				base.AcrastructDecryptionCounter.WithLabelValues(base.DecryptionTypeSuccess).Inc()
				fieldLogger.Debugln("Update with decrypted value")
				output = append(output, PutLengthEncodedString(decryptedValue)...)
				value = decryptedValue
			} else {
				if err != errPlainData {
					span.AddAttributes(trace.BoolAttribute("failed_decryption", true))
					base.AcrastructDecryptionCounter.WithLabelValues(base.DecryptionTypeFail).Inc()
				}
				fieldLogger.Debugln("Leave value as is")
				output = append(output, rowData[pos:pos+n]...)
			}
			pos += n
			continue
		} else {
			fieldLogger.Debugln("Field is not binary")
		}
```
to some `OnColumnEncryption` handler, subscribe him and leave here only `value, err = handler.onColumnDecryption(ctx, i, value)` to leave only one decryption

decryptor/mysql/response_proxy.go
```diff
@@ -477,64 +514,230 @@ func (handler *Handler) processBinaryDataRow(ctx context.Context, rowData []byte
...
 			return nil, fmt.Errorf("found unknown FieldType <type=%d> <name=%s> in MySQL response packet", fields[i].Type, fields[i].Name)
 		}
 	}
 	return output, nil
 }
 
+func (handler *Handler) processFixedSizeNumberField(ctx context.Context, columnIndex int, column *ColumnDescription, encoded []byte) ([]byte, error) {
+	var value []byte
+	var err error
+	// Parse encoded number value into ASCII string (because that's what subscribers expect).
```
Lagovas, line 604: hm... It good if client's driver sends numbers only using SQL and as strings. Then yes, our proxy work with numbers as with strings. I don't remember good, but probably mysql driver can send values with prepared statement as binary values. And then they will come to us as binary values, not strings. And will be encrypted as binaries. And then if they will back to user as result on some query, probably they will be encoded in text format and proxy will not decrypt them. Or we should encode all formats to one (binary/text to text or binary/text to binary), decrypt/encrypt and then encode to format which we took from client/db...

decryptor/postgresql/pg_decryptor.go
```diff
@@ -89,17 +88,16 @@ const (
...
+	clientConnection     net.Conn
+	dbConnection         net.Conn
+	TLSCh                chan bool
+	ctx                  context.Context
+	queryObserverManager base.QueryObserverManager
+	tlsConfig            *tls.Config
+	censor               acracensor.AcraCensorInterface
+	decryptor            base.Decryptor
+	tlsSwitch            bool
+	decryptionObserver   base.ColumnDecryptionObserver
```
Lagovas, line 100: cool, thanks for that and unifying)

decryptor/base/decryptionNotification.go
```diff
@@ -95,6 +99,7 @@ func ColumnInfoFromContext(ctx context.Context) (ColumnInfo, bool) {
 // DecryptionSubscriber interface to subscribe on column's data in db responses
 type DecryptionSubscriber interface {
 	OnColumn(context.Context, []byte) (context.Context, []byte, error)
+	ID() string
```
Lagovas, line 102: for what it used? ID of what it should return?

decryptor/mysql/response_proxy.go
ilammy: > Maybe better to remove if check, and leave base logic with onColumnDecryption. 

I thought about it but left as is for now. As I understand, that's how PostgreSQL code handles decryption and it would be nice to unify that. I'll look into the diff, if it's reasonable I'll include a commit into this PR. It this requires more that just some code removal and fixups then I guess that should be a separate PR.

decryptor/mysql/response_proxy.go
ilammy: > It good if client's driver sends numbers only using SQL and as strings.

Yeah, that's bothered me as well since binary protocol is a thing, and I'm not that that we handle it at all for MySQL. The integration tests pass, but that's not a guarantee it will work.

The code chunk here does exactly what you say for the decryption path: text is handled as text, binary data is converted into text for processing and then encoded back to the same binary form as in the response sent. However, the encryption path expects and works only with text, as I understand.

decryptor/base/decryptionNotification.go
ilammy: > for what it used? ID of what it should return?

ID of the subscriber, it's only used for error reporting right now.

decryptor/mysql/response_proxy.go
Lagovas: create please task in phabricator to do it in the future, to separate cases when data encrypted with text format and queried and returned in binary and vice versa with describing the place where it should be done and it's okay for now imho

decryptor/mysql/response_proxy.go
ilammy: Indeed, it is possible. I have pushed a commit which replaces hardwired decryption code in MySQL with using the OnColumn notifications. The common decryptor will now take care of that: check if a field can contain AcraStructs, try to decrypt them, update success/failure counters, etc. This also works for Acra EE.

Lagovas: looks much better, less code) missed that already exists base code which do decryption and we will need create on more subscriber with checks over mysql binary types according to supported types by `isFieldToDecrypt`.

ilammy: > we will need create on more subscriber

Uh... Do we? The `OnColumn` is already executed for both text fields and binary fields. For text fields it is executed regardless of the field type: tokenization may be interested in integers, not just byte arrays, and decryption skips most of non-bytes fields because they are too small. For binary fields there are separate core paths, but again, decryption will be applied to numbers and will be a no-op. It's probably not ideal, but I guess it shouldn't impact performance too much.

decryptor/mysql/response_proxy.go
ilammy: > create please task in phabricator to do it in the future

Tracked in T1757.

decryptor/mysql/response_proxy.go
Lagovas: looked more thoughtful, questions disappeared and have not more questions. found where and how configured decryption for mysql in acra-ce and acra-ee

PR #400, opened 2020-08-17
Move "acra-migrate-keys" => "acra-keys migrate"

ilammy:
Since we have a multitool `acra-keys`, get rid of dedicated acra-migrate-keys and move its features into the common tool. Hopefully, this simplifies user experience, and makes it easier for us to reuse code in Acra EE.

- Drop list of subcommand names

Preparatory refactoring.

- Basic "acra-keys migrate" subcommand
- Command line of "acra-keys migrate"

Add `acra-keys migrate` subcommand and its command-line. The options are the same as before, just passed to the subcommand. The master keys are to be passed via `SRC_ACRA_MASTER_KEY` and `DST_ACRA_MASTER_KEY` too. Basically, you can replace `acra-migrate-keys` with

```shell
#!/bin/sh
acra-keys migrate "$@"
```

- Opening key stores for migration
- Key migration routine
- Key migration implementation

These commits implement the actual tool.

- Use "acra-keys migrate" in tests
- Drop acra-migrate-keys
- Update documentation references

Cleaning up: updating tests to use the new tool, dropping the old one, etc.


cmd/acra-keys/keys/migrate-keys.go
```diff
@@ -0,0 +1,285 @@
...
+
+// Environment variables from which master keys are read.
+const (
+	SrcMasterKeyVarName = "SRC_" + keystoreV1.AcraMasterKeyVarName
+	DstMasterKeyVarName = "DST_" + keystoreV1.AcraMasterKeyVarName
+)
+
+// Command-line errors for "acra-keys migrate":
+var (
+	ErrMissingFormat = errors.New("key store format not specified")
```
vixentael, line 62:
```suggestion
	ErrMissingFormat = errors.New("keystore format not specified")
```
not sure which wording we use

cmd/acra-keys/keys/migrate-keys.go
```diff
@@ -0,0 +1,285 @@
...
+
+	return nil
+}
+
+// Execute this subcommand.
+func (m *MigrateKeysSubcommand) Execute() {
+	if m.SrcKeyStoreVersion() == "v1" && m.DstKeyStoreVersion() == "v2" {
+		keyStoreV1, err := m.openKeyStoreV1(SrcMasterKeyVarName, m.SrcKeyStoreParams())
+		if err != nil {
+			log.WithError(err).Fatal("Failed to open keystore v1 (src)")
```
vixentael, line 168: let's use consistent wording: either `keystore` or `key store` everywhere?

(my personal pref is `keystore`, but i'm ok with both)

cmd/acra-keys/keys/migrate-keys.go
```diff
@@ -0,0 +1,285 @@
...
+
+// RegisterFlags registers command-line flags of "acra-keys migrate".
+func (m *MigrateKeysSubcommand) RegisterFlags() {
+	m.flagSet = flag.NewFlagSet(CmdMigrateKeys, flag.ContinueOnError)
+	m.src.RegisterPrefixed(m.flagSet, "", "src_", "(old key store, source)")
+	m.dst.RegisterPrefixed(m.flagSet, "", "dst_", "(new key store, destination)")
+	m.flagSet.StringVar(&m.srcVersion, "src_keystore", "", "key store format to use: v1 (current), v2 (new)")
+	m.flagSet.StringVar(&m.dstVersion, "dst_keystore", "", "key store format to use: v1 (current), v2 (new)")
+	m.flagSet.BoolVar(&m.dryRun, "dry_run", false, "try migration without writing to the output key store")
+	m.flagSet.BoolVar(&m.force, "force", false, "write to output key store even if it exists")
```
vixentael, line 114: 🔑 👀

configs/acra-keys.yaml
```diff
@@ -29,6 +29,30 @@ key_bundle_secret:
 # export private key data
 private_keys: false
 
+# try migration without writing to the output key store
```
vixentael, line 32: don't forget to re-generate this file after fixing wording

cmd/acra-keys/keys/migrate-keys.go
ilammy: It seems that both are used with similar frequency in strings, but code use is heavily skewed in favor of "keystore" as a single word because that's how the module is named.

Okay, I'll submit a separate PR to change it all into keystore. As much as I dislike such changes.

configs/acra-keys.yaml
ilammy: > don't forget to re-generate this file after fixing wording

CI checks this and will not allow humans to forget.

ilammy:
> place a task to describe acra-keys migrate in docs

Already there, just need a little tweak:
https://github.com/cossacklabs/product-docs/blob/master/acra/keys/versions/migrate-v1-to-v2.md
which I forgot to submit.

PR #402, opened 2020-08-19
Test harness improvements

ilammy:
This PR contains several loosely related changes to the test harness which ultimately enable Acra EE to run Acra CE test suite with Acra EE binaries. They also improve some aspects of the test suite, making its manual use during development a bit easier.

If TEST_CLEAN_BINARIES environment variable is set to 0, integration test suite will leave the test binaries in the current directory, speeding up future runs as they won't need to recompile them all. The binaries will still be recompiled as necessary if there are changes in the source tree, so you can use this for incremental test runs.

The `configs/regenerate.sh` script now updates all configuration files at once, it is no longer necessary to manually add new binaries there. Additionally, if will use the binaries in the current directory if available which speeds up the integration test suite as it no longer needs to recompile them.

You can read other gory details in the comments.


configs/regenerate.sh
```diff
@@ -1,13 +1,24 @@
...
-for cmd in "${binaries[@]}"; do
- if [[ "$#" == "1" ]]; then
-    args="--dump_config --config_file=$1/acra-${cmd}.yaml"
- fi
- go run ./cmd/acra-${cmd}/*.go `echo ${args}`
+for cmd in $(ls ./cmd/ | grep ^acra-); do
+    if [[ $# == 1 ]]; then
+        extra="--config_file=$1/${cmd}.yaml"
+    fi
+    if [[ -f ./$cmd ]]; then
```
Lagovas, line 19: can you add some comments for what it and where it used? because after several months someone may remove it as unnecessary piece of code)

tests/test.py
```diff
@@ -464,32 +488,21 @@ def drop_tables():
     metadata.drop_all(engine_raw)
 
 
+# Set this to False to not rebuild binaries on setup.
+CLEAN_BINARIES = bool(os.environ.get('TEST_CLEAN_BINARIES', True))
+# Set this to False to not build binaries in principle.
+BUILD_BINARIES = True
```
Lagovas, line 494: why not allow to set value from env too? as I understand, it's for manual debugging only?

Lagovas (approved):

tests/test.py
```diff
@@ -4306,6 +4323,8 @@ def testStartupWithDifferentConfigsPatchVersion(self):
                 'acra-connector': {'connection': 'connection_string',
                                    'args': ['-keys_dir={}'.format(KEYS_FOLDER.name)],
                                    'status': 1},
+                'acra-heartbeat': {'args': ['--connection_string=please-fail'],
```
vixentael, line 4326: very nice connection string

vixentael (approved):

tests/test.py
ilammy: > why not allow to set value from env too? as I understand, it's for manual debugging only?

Yes, I did not add an environment variable for skipping builds because I did not think it was useful for manual debugging. It makes sense to build binaries and not clean them to avoid rebuilding, but it does not make sense – when working on Acra CE – to avoid building Acra CE binaries entirely. This hack is only necessary to run Acra CE tests with Acra EE binaries, where we want Acra CE tests to not build their binaries and use Acra EE ones.

configs/regenerate.sh
ilammy: > can you add some comments for what it and where it used?

Sure, will do. You're right that a comment is more visible than `git blame`.

ilammy: Done. Now there is a comment explaining that this line exists to speed up test execution.

tests/test.py
ilammy: While I've been experimenting with `TEST_CLEAN_BINARIES` I have realized that `bool(os.environ.get(...))` does not really work as expected. Setting it to `0` or `false` does not parse into boolean `False`. In fact, it looks at whether the variable is set and whether it's empty. I have been setting and unsetting the variable for testing so I did not notice.

The same approach is used to get the value of TEST_MYSQL variable which I reused. Well, this is not nice, so I have added a utility to get boolean variables in a more convenient way:

```python
CLEAN_BINARIES = utls.get_bool_env('TEST_CLEAN_BINARIES', default=True)
```

Now `0`, `false`, `no`, `n`, and empty string are treated as `False`, while `1`, `true`, `yes,` and `y` are treated as `True`.

PR #405, opened 2020-08-26
Merge keystore v2 into master

ilammy:
Recently it has been revealed that **keystore-v2** branch has improved support for key rotation (#375) but **master** does not have it. We were all under impression that it does, including in Acra EE. I guess it's time to merge key store v2 into mainline. It has been stabilized enough and has a decent test coverage, despite some rough edges.

All key store v2 functionality is under a feature flag. Unless the user explicitly opts in by using `--keystore=v2` option during key store creation, Acra will use key store v1. Note that the `--keystore` option is now **required** for `acra-keymaker.`

**Don't squash, create a merge commit** to preserve change history.


vixentael (approved):
TLDR: Acra services can work with v1 or v2 keystore. Which keystore to use depends on keystore properties. Key store is generated by `acra-keymaker` during key gen. `acra-keymaker` requires explicit usage of `--keystore=v2|v1` flag.

Keystore v1->v2 migration is possible by running `acra-keys migrate`

PR #406, opened 2020-08-26
Using waitGroup mechanism on ReaderServer level

storojs72:
This PR is a next step in our efforts for better controlling shutdown process of our services (AcraTranslator for this PR), according to T1661. Basically, I added a low-level waitGroup for every goroutine spawned by ReaderServer (main component of AcraTranslator). Additionaly a timeout is used in order to avoid infinite blocking if some of background goroutine "hangs on" for unexpected reason. Next step will be refactoring of EE AcraTranslator's main function where "more global" waitGroup should be used (for controlling background goroutines, spawned there).


cmd/acra-translator/server/server.go
```diff
@@ -298,6 +312,29 @@ func (server *ReaderServer) Start(parentContext context.Context) {
...
+		defer close(c)
+		wg.Wait()
+	}()
+	select {
+	case <-c:
+		return false // completed normally
+	case <-time.After(timeout):
+		return true // timed out
+	}
 }
```
ilammy, line 338: This neat piece of code is going to be useful in other places as well. I'd move it utilities, calling something like `WaitWithTimeout`, `TimedWait`, `TimeoutWait`, or something.

cmd/acra-translator/server/server.go
```diff
@@ -298,6 +312,29 @@ func (server *ReaderServer) Start(parentContext context.Context) {
 		}()
 	}
 	<-parentContext.Done()
+
+	// global 'cancel' has been called. Now we should wait (not more than specified duration) until all
+	// background goroutines spawned by readerServer will finish their execution
+	if timeoutExpired(&server.waitGroup, time.Second) {
```
ilammy, line 318: I'd extract `time.Second` into a constant like this
```go
// time to wait until goroutine shutdown
const terminationTimeout = time.Second
```
and put it somewhere at the beginning of the file so that it's easy to tweak if needed.

ilammy (approved):
I'm okay with this change technically.

Though my inner functional programmer urges to extract this workgroup thingie into a reusable piece of code, I'm not sure it will look very nice. So I guess I'll submit to a Go maxim: copypaste if you need it.

Also, there are a couple of goroutines launched in `common/network`:
https://github.com/cossacklabs/acra/blob/db23a52318029c7338a6c07ce2381d30eeadd56a/cmd/acra-translator/common/network.go#L60-L68
I guess those are going to linger for some time, unless we figure out a way to inject a waitgroup there as well.

vixentael (approved):
I agree w/ @ilammy

cmd/acra-translator/server/server.go
```diff
@@ -41,27 +42,27 @@ import (
...
 // NewReaderServer creates Reader server with provided params.
 func NewReaderServer(config *common.AcraTranslatorConfig, keystorage keystore.TranslationKeyStore, grpcServerFactory common.GRPCServerFactory, waitTimeout time.Duration) (server *ReaderServer, err error) {
+	var wg sync.WaitGroup
 	return &ReaderServer{
 		grpcServerFactory: grpcServerFactory,
 		waitTimeout:       waitTimeout,
 		config:            config,
 		keystorage:        keystorage,
 		connectionManager: network.NewConnectionManager(),
+		waitGroup:         wg,
```
Lagovas, line 65: actually if we declare field as value, not as pointer, than we don't need to initialize it. golang initialize all values by their default values.. we can skip `var wg sync.WaitGroup`.

cmd/acra-translator/server/server.go
Lagovas: +1

cmd/acra-translator/server/server.go
storojs72: Yes. But it is not very expressive naming for this function and I agree that it can be moved to our utility functions

cmd/acra-translator/server/server.go
storojs72: Done in https://github.com/cossacklabs/acra/pull/406/commits/55e177ef6709e55c8d819195c9989dda72a96d1f

cmd/acra-translator/server/server.go
storojs72: Done in https://github.com/cossacklabs/acra/pull/406/commits/55e177ef6709e55c8d819195c9989dda72a96d1f

cmd/acra-translator/server/server.go
storojs72: Done in https://github.com/cossacklabs/acra/pull/406/commits/55e177ef6709e55c8d819195c9989dda72a96d1f

PR #407, opened 2020-09-01
RegisterWithContext for signal handlers

storojs72:
This PR is a technical one. It is needed for EE.

Basically it moves `DefaultWaitGroupTimeoutDuration` constant (and makes it exportable in order to reuse in other packages) to common `utils` package. Additionally it adds new `RegisterWithContext` function to signal handler utility component that actually behaves like usual `Register` but avoids `os.Exit` and allows passing global context for better controlling of background goroutine that runs it.

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -29,6 +29,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"reflect"
+	"context"
 
 	"encoding/base64"
 	"github.com/cossacklabs/acra/keystore"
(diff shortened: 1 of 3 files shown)


Lagovas (approved):

vixentael (approved):

cmd/utils.go
```diff
@@ -108,6 +103,26 @@ func (handler *SignalHandler) RegisterWithoutExiting() {
...
+		case <-handler.ch:
+			for _, listener := range handler.listeners {
+				listener.Close()
+			}
+			for _, callback := range handler.callbacks {
+				callback()
+			}
+		case <-globalContext.Done():
+			// got signal for shutdown, so just return
+			return
```
ilammy, line 123: Do I understand it right that it is not necessary to close any listeners in this case because they will be closed by AcraServer when it sees that the context is Done()?

ilammy (approved):

cmd/utils.go
Lagovas: Not necessary, because some other component initialized exit from the program and here we just notify the background goroutine that it should stop the execution, to avoid any background actions.

For example, when acra-server receives SIGTERM, sigterm handler stops all components and should notify all goroutines that they should stop any processing to finalize log. It will raise SIGUSR1 handler too, but it should not do anything except finish goroutine.

PR #408, opened 2020-09-09
Generate keys with "acra-keys generate"

ilammy:
Add a new subcommand for key generation. This is a replacement for `acra-keymaker` and other tools like `acra-addzone`.

### Generating master keys

The first step is to generate a master key:

```
acra-keys generate --master_key_path=master.key --keystore=v2
```

The `--keystore` option is required since the format of master keys is different for keystore versions.

The resulting master key is placed into the specified file. It needs to be set in the environment:

```
ACRA_MASTER_KEY=$(cat master.key | base64)
```

### Generating initial keystore

After you have a master key, generate the initial keystore:

```
acra-keys generate --keystore=v2 --client_id=ACME
```

The `--keystore` option is required here as well since we need to know which format to use. When the keystore is already initialized, this option is not necessary as `acra-keys` can see the format for itself.

The `--client_id` option is now **required** as well. This is different from `acra-keymaker` which uses client ID `client` when it is not specified explicitly.

`acra-keys` will generate a default set of keys:

  - transport keys for AcraConnector, AcraServer, AcraTranslator
  - storage key for AcraWriter to use

It is also possible to initialize the keystore with only some of the keys. For example, if you need only the storage key without the transport keys, initialize the keystore like this:

```
acra-keys generate --keystore=v2 --client_id=ACME --client_storage_key
```

Note that key options are **different** from the ones used by `acra-keymaker`.

| Option | Key |
| -- | -- |
| `--acraconnector_transport_key` | transport keypair for AcraConnector |
| `--acraserver_transport_key` | transport keypair for AcraServer |
| `--acratranslator_transport_key` | transport keypair for AcraTranslator |
| `--client_storage_key` |  encryption keypair for AcraWriter |
| `--acrawebconfig_symmetric_key` | encryption key for AcraWebConfig |

The keystore will be initialized in the default directory: `.acrakeys` in the current directory. You can change this with `--keys_dir` (and `--keys_dir_public` for v1). Note that these option names are **different** from `acra-keymaker` which uses `--keys_output_dir`. This is to be consistent with other `acra-keys` subcommands.

### Rotating existing keys

`acra-keys generate` can be used to rotate keys in the existing keystore. Just run it again and specify the keys you need to rotate:

```
acra-keys generate --client_id=ACME --acraserver_transport_key
```

This will generate a new key for the `ACME` client to use on AcraConnector when connecting to AcraServer.

### New features for zones

`acra-keys generate` can also be used instead of `acra-addzone` to generate new zones:

```
$ acra-keys generate --zone
{"id":"DDDDDDDDtapktGNSfSnodqtv","public_key":"VUVDMgAAAC1UemR1A1qdHwVPp4iFTZKnHFNOMQqHmUc04SFGxwUk3wnmY0ut"}
INFO[0000] Generated new Acra zone
```

JSON output format is the same as `acra-addzone`.

You can also use it to rotate existing zone keys (previously this was impossible to do):

```
$ acra-keys generate --zone_id=DDDDDDDDtapktGNSfSnodqtv --zone_storage_key
{"id":"DDDDDDDDtapktGNSfSnodqtv","public_key":"VUVDMgAAAC0cIjpMAoXUNFamoZqCDaY9wAHAT/tXhfWOqTAYDqPp5PZAXtwB"}
INFO[0000] Generated zone storage key               
```

### Compatibility with `acra-keymaker` and its replacement

Note that the command-line is different and not compatible with `acra-keymaker`. The plan is to eventually deprecate `acra-keymaker` in favor of `acra-keys`. We could possible reimplement `acra-keymaker` and `acra-addzone` using the shared code from `acra-tools` to reduce maintenance burden. For now, both `acra-keymaker` and `acra-keys generate` will be supported, with the latter being the preferred way to generate keys.

### Pending tasks

- [x] Review by @Lagovas
- [x] Review by @shad
- [x] Review by @vixentael 
- [x] Adjust this PR for Acra EE
- [x] Rename `--new_zone` => `--zone`

### DIFF
--- a/cmd/acra-keys/keys/generate.go
+++ b/cmd/acra-keys/keys/generate.go
@@ -255,9 +255,12 @@ func (g *GenerateKeySubcommand) Execute() {
 		log.WithError(err).Fatal("Failed to open keystore")
 	}
 
-	err = GenerateAcraKeys(g, keystore)
+	generatedKeys, err := GenerateAcraKeys(g, keystore)
 	if err != nil {
 		log.WithError(err).Fatal("Failed to generate keys")
+	}
+	if !generatedKeys {
+		log.Info("No keys were updated")
 	}
 }
 
@@ -293,7 +296,8 @@ func GenerateMasterKey(params GenerateKeyParams) error {
 }
 
 // GenerateAcraKeys generates Acra CE keys as specified by the parameters.
-func GenerateAcraKeys(params GenerateKeyParams, keystore keystore.KeyMaking) error {
+// Returns true if some keys have been generated.
+func GenerateAcraKeys(params GenerateKeyParams, keystore keystore.KeyMaking) (bool, error) {
 	generateAcraConnector := params.GenerateAcraConnector()
 	generateAcraServer := params.GenerateAcraServer()
 	generateAcraTranslator := params.GenerateAcraTranslator()
@@ -312,37 +316,47 @@ func GenerateAcraKeys(params GenerateKeyParams, keystore keystore.KeyMaking) err
 		generateAcraWriter = true
 	}
 
+	// If the user runs just "acra-keys generate" with no arguments and the configuration file
+	// does not tell us the action either, we end up not doing anything useful.
+	// Return this state to the caller so that we can at least tell the user than nothing changed
+	// instead of keeping an ominous silence.
+	didSomething := false
+
 	if generateAcraConnector {
 		err := keystore.GenerateConnectorKeys(params.ClientID())
 		if err != nil {
 			log.WithError(err).Error("Failed to generate AcraConnector transport key")
-			return err
+			return didSomething, err
 		}
 		log.Info("Generated AcraConnector transport key")
+		didSomething = true
 	}
 	if generateAcraServer {
 		err := keystore.GenerateServerKeys(params.ClientID())
 		if err != nil {
 			log.WithError(err).Error("Failed to generate AcraServer transport key")
-			return err
+			return didSomething, err
 		}
 		log.Info("Generated AcraServer transport key")
+		didSomething = true
 	}
 	if generateAcraTranslator {
 		err := keystore.GenerateTranslatorKeys(params.ClientID())
 		if err != nil {
 			log.WithError(err).Error("Failed to generate AcraTranslator transport key")
-			return err
+			return didSomething, err
 		}
 		log.Info("Generated AcraTranslator transport key")
+		didSomething = true
 	}
 	if generateAcraWriter {
 		err := keystore.GenerateDataEncryptionKeys(params.ClientID())
 		if err != nil {
 			log.WithError(err).Error("Failed to generate client storage key")
-			return err
+			return didSomething, err
 		}
 		log.Info("Generated client storage key")
+		didSomething = true
 	}
 
 	if params.GenerateAcraWebConfig() {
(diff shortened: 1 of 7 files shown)


cmd/acra-keys/keys/generate.go
```diff
@@ -269,18 +269,18 @@ func GenerateMasterKey(params GenerateKeyParams) error {
...
 	switch version {
 	case "v1":
 		newKey, err = keystore.GenerateSymmetricKey()
 	case "v2":
 		newKey, err = keystoreV2.NewSerializedMasterKeys()
 	case "":
-		log.Errorf("Key store version is required: --keystore={v1|v2}")
-		return ErrMissingKeyStoreVersion
+		log.Errorf("Keystore version is required: --keystore={v1|v2}")
+		return ErrMissingKeystoreVersion
```
Lagovas, line 280: is it touch acra-ee too?

Lagovas (approved):

cmd/acra-keys/keys/generate.go
ilammy: Right now it does not affect Acra EE, since this is all new code. (I have forgot to rename it before submitting the PR.)

In the future Acra EE will get `acra-keys generate` support as well, reusing most of code from here and adding extra keys supported by Acra EE only.

shadinua:
Great!

> acra-keys generate --new_zone
Looks a little out of style. Probably, it worth to replace `--new_zone` with just `--zone`?

ilammy:
> Probably, it worth to replace --new_zone with just --zone?

Hmm... Won't it cause interference with `--zone_id`? Plus, zones are different in the sense that you can generate multiple distinct zones while generating other keys is merely *rotating* them (so it's still the same key, just a newer version).

shadinua:
You're right, this might be a bit confusing. Anyway, the `new` prefix adds nothing to the description of _type of entity_ it just clarifies the action or position of the created item across the entire timeline.

Today I can only suggest (maybe morning will bring more options?):
```
# Generates a new zone
acra-keys generate --zone
# Rotates a storage key 
acra-keys rotate --zone_storage_key --zone_id=DDDDDDDDtapktGNSfSnodqtv
```

ilammy:
> ```
> acra-keys generate --zone
> acra-keys rotate
> ```

This looks like a better interface, but I'm afraid that the current keystore interface does not really allow to differentiate between generation and rotation of the keys. (The difference is that you cannot rotate a key that does not exist.) We're just abusing the "generate a new one" function to rotate the keys, or generate an initial version if it does not exist.

Back to `--new_zone`, I'm not opposed to renaming it to `--zone`. Any second opinion? If there are no objections, I'm doing it.

ilammy:
I've pushed a minor tweak in `GenerateAcraKeys` interface to give Acra EE more control over the default key generation process. This plays a role in the keystore initialization sequence, when the user is not required to enumerate the keys to generate a default set of them.

shadinua (approved):

cmd/acra-keys/keys/generate.go
```diff
@@ -0,0 +1,435 @@
...
+		generateAcraServer = true
+		generateAcraTranslator = true
+		generateAcraWriter = true
+	}
+
+	// If the user runs just "acra-keys generate" with no arguments and the configuration file
+	// does not tell us the action either, we end up not doing anything useful.
+	// Return this state to the caller so that we can at least tell the user than nothing changed
+	// instead of keeping an ominous silence.
+	didSomething := false
```
vixentael, line 352: that's a good idea

vixentael (approved):

vixentael:
I suggest we copy the PR description into docs :D

PR #409, opened 2020-09-11
Improve Acra Go Report scores

iamnotacake:
* Fix `gofmt` issues (apply suggestions by `gofmt`)
* Fix misspells
* Fix golint issues (add comments to exported types and functions)
* Fix `ineffassign` issues (those are assignments to variables that are unused or rewritten afterwards)
* Fix two bugs (I guess) detected by `ineffassign`:
    * In `cmd/acra-server/client_commands_session.go` the function was
      returning instead of breaking `switch` and reporting error to client
    * In `decryptor/mysql/response_proxy.go` we had `:=` that overrides the
      scope so the new `err` was created instead of writing to already
      defined one

Feel free to remove the :beetle:`bug` label if those bugs are not actually bugs and I misunderstood the code.

It turns out that CircleCI scripts are a little bit broken and `golint` check was not performed automatically when PRs were created/updated. Gonna fix that in different PR as well as add `misspell` checker, and maybe `ineffassign` one.

### DIFF
--- a/decryptor/binary/binary_decryptor.go
+++ b/decryptor/binary/binary_decryptor.go
@@ -169,6 +169,8 @@ func (decryptor *Decryptor) ReadData(symmetricKey, zoneID []byte, reader io.Read
 
 	scell := cell.New(symmetricKey, cell.ModeSeal)
 	decrypted, err := scell.Unprotect(data, nil, zoneID)
+	utils.FillSlice(0, data)
+	data = nil
 
 	// fill zero symmetric_key
 	utils.FillSlice(byte(0), symmetricKey)
--- a/decryptor/binary/binary_decryptor.go
+++ b/decryptor/binary/binary_decryptor.go
@@ -169,7 +169,7 @@ func (decryptor *Decryptor) ReadData(symmetricKey, zoneID []byte, reader io.Read
 
 	scell := cell.New(symmetricKey, cell.ModeSeal)
 	decrypted, err := scell.Unprotect(data, nil, zoneID)
-	data = nil
+
 	// fill zero symmetric_key
 	utils.FillSlice(byte(0), symmetricKey)
 	if err != nil {
--- a/decryptor/postgresql/pg_hex_decryptor.go
+++ b/decryptor/postgresql/pg_hex_decryptor.go
@@ -252,7 +252,7 @@ func (decryptor *PgHexDecryptor) ReadData(symmetricKey, zoneID []byte, reader io
 	scell := cell.New(symmetricKey, cell.ModeSeal)
 
 	decrypted, err := scell.Unprotect(data, nil, zoneID)
-	data = nil
+
 	// fill zero symmetric_key
 	utils.FillSlice(byte(0), symmetricKey)
 	if err != nil {
@@ -262,7 +262,7 @@ func (decryptor *PgHexDecryptor) ReadData(symmetricKey, zoneID []byte, reader io
 	outputLength := hexEncodedLen(uint64(len(decrypted)))
 	decryptor.checkBuf(&decryptor.output, outputLength)
 	hex.Encode(decryptor.output[:outputLength], decrypted)
-	decrypted = nil
+
 	return decryptor.output[:outputLength], nil
 }
 
--- a/decryptor/postgresql/pg_hex_decryptor.go
+++ b/decryptor/postgresql/pg_hex_decryptor.go
@@ -252,6 +252,8 @@ func (decryptor *PgHexDecryptor) ReadData(symmetricKey, zoneID []byte, reader io
 	scell := cell.New(symmetricKey, cell.ModeSeal)
 
 	decrypted, err := scell.Unprotect(data, nil, zoneID)
+	utils.FillSlice(0, data)
+	data = nil
 
 	// fill zero symmetric_key
 	utils.FillSlice(byte(0), symmetricKey)
@@ -262,6 +264,8 @@ func (decryptor *PgHexDecryptor) ReadData(symmetricKey, zoneID []byte, reader io
 	outputLength := hexEncodedLen(uint64(len(decrypted)))
 	decryptor.checkBuf(&decryptor.output, outputLength)
 	hex.Encode(decryptor.output[:outputLength], decrypted)
+	utils.FillSlice(0, decrypted)
+	decrypted = nil
 
 	return decryptor.output[:outputLength], nil
 }
--- a/keystore/filesystem/server_keystore_test.go
+++ b/keystore/filesystem/server_keystore_test.go
@@ -495,17 +495,26 @@ func TestFilesystemKeyStoreExport(t *testing.T) {
 	}
 	// Since we cannot access all generated key pairs via AcraServer keystore,
 	// we generate them here and use Save... API
-	connectorKeyPair, _ := keys.New(keys.TypeEC)
+	connectorKeyPair, err := keys.New(keys.TypeEC)
+	if err != nil {
+		t.Fatalf("keys.New() failed: %v", err)
+	}
 	err = keyStore.SaveConnectorKeypair(clientID, connectorKeyPair)
 	if err != nil {
 		t.Fatalf("SaveConnectorKeypair() failed: %v", err)
 	}
-	serverKeyPair, _ := keys.New(keys.TypeEC)
+	serverKeyPair, err := keys.New(keys.TypeEC)
+	if err != nil {
+		t.Fatalf("keys.New() failed: %v", err)
+	}
 	err = keyStore.SaveServerKeypair(clientID, serverKeyPair)
 	if err != nil {
 		t.Fatalf("SaveServerKeypair() failed: %v", err)
 	}
-	translatorKeyPair, _ := keys.New(keys.TypeEC)
+	translatorKeyPair, err := keys.New(keys.TypeEC)
+	if err != nil {
+		t.Fatalf("keys.New() failed: %v", err)
+	}
 	err = keyStore.SaveTranslatorKeypair(clientID, translatorKeyPair)
 	if err != nil {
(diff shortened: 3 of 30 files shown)


keystore/filesystem/server_keystore_test.go
```diff
@@ -495,17 +495,17 @@ func TestFilesystemKeyStoreExport(t *testing.T) {
 	}
 	// Since we cannot access all generated key pairs via AcraServer keystore,
 	// we generate them here and use Save... API
-	connectorKeyPair, err := keys.New(keys.TypeEC)
+	connectorKeyPair, _ := keys.New(keys.TypeEC)
```
Lagovas, line 498: it's anti-pattern to avoid error check in golang. even if you know that this code always returns `nil` inside, you should check error value, because in a future somebody can change it and expect that it checked in other pieces of code. even if it tested above or below. In such cases you can create adapter which always will check error and encapsulate it for other callers:
```
func generateKeyPair(t *testing.T) *keys.KeyPair {
  keypair, err := keys.New(keys.TypeEC)
  if err != nil {t.Fatal(err)}
  return keypair
}

func TestSomething (t *testing.T){
  connectorKeyPair := generateKeyPair(t)
}
```

keystore/v2/keystore/importV1_test.go
```diff
@@ -100,17 +100,17 @@ func TestImportKeyStoreV1(t *testing.T) {
 	}
 	// Since we cannot access all generated key pairs via AcraServer keystore,
 	// we generate them here and use Save... API
-	connectorKeyPairV1, err := keys.New(keys.TypeEC)
+	connectorKeyPairV1, _ := keys.New(keys.TypeEC)
```
Lagovas, line 103: same

ilammy (approved):
I wholeheartedly support this change! Thanks for doing this tedious work of inspecting each and every warning :heart:

> Fix two bugs (I guess) detected by ineffassign

I believe Acra EE could benefit from similar inspection. Who how many bugs are lurking there in the dark.

> Feel free to remove the beetlebug label if those bugs are not actually bugs and I misunderstood the code.

We have discussed this in private with @iamnotacake, I also believe those are genuine bugs.

Lagovas (changes requested):

decryptor/binary/binary_decryptor.go
```diff
@@ -169,7 +169,7 @@ func (decryptor *Decryptor) ReadData(symmetricKey, zoneID []byte, reader io.Read
 
 	scell := cell.New(symmetricKey, cell.ModeSeal)
 	decrypted, err := scell.Unprotect(data, nil, zoneID)
-	data = nil
+
```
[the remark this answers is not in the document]
Lagovas: yes

decryptor/binary/binary_decryptor.go
```diff
@@ -169,7 +169,7 @@ func (decryptor *Decryptor) ReadData(symmetricKey, zoneID []byte, reader io.Read
 
 	scell := cell.New(symmetricKey, cell.ModeSeal)
 	decrypted, err := scell.Unprotect(data, nil, zoneID)
-	data = nil
```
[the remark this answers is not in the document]
vixentael: let's try to zeroize data, if not, put back `data = nil`

decryptor/postgresql/pg_hex_decryptor.go
```diff
@@ -252,7 +252,7 @@ func (decryptor *PgHexDecryptor) ReadData(symmetricKey, zoneID []byte, reader io
 	scell := cell.New(symmetricKey, cell.ModeSeal)
 
 	decrypted, err := scell.Unprotect(data, nil, zoneID)
-	data = nil
```
vixentael, line 255: same as above

decryptor/postgresql/pg_hex_decryptor.go
```diff
@@ -262,7 +262,7 @@ func (decryptor *PgHexDecryptor) ReadData(symmetricKey, zoneID []byte, reader io
 	outputLength := hexEncodedLen(uint64(len(decrypted)))
 	decryptor.checkBuf(&decryptor.output, outputLength)
 	hex.Encode(decryptor.output[:outputLength], decrypted)
-	decrypted = nil
+
```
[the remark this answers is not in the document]
vixentael: agree. zeroize & free.

keystore/filesystem/server_keystore_test.go
vixentael: agree

keystore/filesystem/server_keystore_test.go
```diff
@@ -495,17 +495,17 @@ func TestFilesystemKeyStoreExport(t *testing.T) {
...
 	err = keyStore.SaveConnectorKeypair(clientID, connectorKeyPair)
 	if err != nil {
 		t.Fatalf("SaveConnectorKeypair() failed: %v", err)
 	}
-	serverKeyPair, err := keys.New(keys.TypeEC)
```
vixentael, line 503: same

keystore/filesystem/server_keystore_test.go
```diff
@@ -495,17 +495,17 @@ func TestFilesystemKeyStoreExport(t *testing.T) {
...
+	serverKeyPair, _ := keys.New(keys.TypeEC)
 	err = keyStore.SaveServerKeypair(clientID, serverKeyPair)
 	if err != nil {
 		t.Fatalf("SaveServerKeypair() failed: %v", err)
 	}
-	translatorKeyPair, err := keys.New(keys.TypeEC)
```
vixentael, line 508: same

vixentael (commented):
Thank you, that's a lot of work done!

However, please pay attention of zeroizing data.

iamnotacake:
The best solution I've found so far for zeroing byte slices looks like this:
```go
for i := range b {
   b[i] = 0
}
```
This loop seem to be recognized by compiler ([disasm](https://godbolt.org/z/YW57ns), [codegen test](https://github.com/golang/go/blob/7615b20d06500fe5c95c061f6ff32e0c97639a60/test/codegen/slices.go#L18-L24)) and calls `runtime.memclrNoHeapPointers` to make it zero, it's quite fast on common architectures (including x86_64).
What do y'all think about using this to zero plaintext buffers (byte slices actually) and using `b = nil` as before for making ciphertext buffers deallocate faster? Even though I'm not sure `= nil` it really makes a difference.

Lagovas:
we already have functions for that - https://github.com/cossacklabs/acra/blob/master/utils/utils.go#L143

iamnotacake:
> we already have functions for that - https://github.com/cossacklabs/acra/blob/master/utils/utils.go#L143

Wow, thanks for info. However... seems like it cannot be optimized properly. See [disassemble](https://godbolt.org/z/YKq3oz), no call to `memclrNoHeapPointers`, replace `= value` with `= 0` and woah, it's optimized. Go compiler is quite ~dumb~ simple. So I guess it's better to inline the loop. Changing that function signature isn't a good idea. Please correct me if I'm wrong.

Lagovas:
it's okay to optimize our zeroize function. Probably will be good to check do we use FillSlice for any other purposes except zeroizing. If not then we can add new one like ZeroizeBuffer and replace FillSlice with it. If we use it for some filling with specific value except 0 then leave it, add new and replace only when we do zeroizing.
@ilammy interesting ur opinion too)

vixentael:
re: zeroizing optimizations. As zeroizing is part of security code practices, we should be careful and shouldn't bury security-sensitive changes in large "improve Acra Go report" PR.

Please keep this PR as `refactoring` (aka changing code without changing its functionality), not as `rewriting functions that code-base relies on`.

1. Use existing `ZeroizeSymmetricKey` function and `data=nil` to clean up the data in this PR.
2. If you think that optimizations are necessary, please create separate issue, and separate PR. In a perfect world, we should make sure that changes in zeroing won't create new security risks (how? unit tests? memory footprint tests?).

ilammy:
I just looked through the code and I don't see `FillSlice` being used with anything but zero as the first argument, neither in CE nor in EE. So I'd support optimizing `ZeroizeThing` functions, replacing any remaining calls to `FillSlice` with optimizing zeroization, and removing `FillSlice` altogether. Though, I think this is a topic for a separate PR and should be tracked in a separate task. Here, in this PR, I think the changes should be limited to some new zeroization calls.

> Ah, I'm thinking too much about optimizations.

That's fine, but any talk about optimizations should be accompanied not only by disassemblies, but benchmarks as well. And it assumes that no one chimes in with _economic_ benchmarks which might say that it's not worth arguing for several hours about the best approach to optimize the 0,001% run time of Acra to take 0,0009% instead :wink:

> Even though I'm not sure = nil it really makes a difference.

I'm personally not sure either. I have no clue how Go's GC actually works, but given the other things in the language design, I would not expect it to be too sophisticated either. HotSpot's GC, for example, tracks liveness of variables so if a local variable is not used in a function after a certain point then GC treats is as dead, regardless of whether you set it to `null`. Go's GC is probably not that smart, but it's not really that likely for collection to happen before the function returns, and after it returns the stack variable is dead (but the parent frames might still reference the slice).

So to sum it up: it's all mumbo-jumbo again, and I'd rather avoid cargo-culting with

```
data = nil // free the memory
```

unless there's supporting evidence. That notwithstanding, Go's GC is weird and [*there are* cases](https://blog.twitch.tv/en/2019/04/10/go-memory-ballast-how-i-learnt-to-stop-worrying-and-love-the-heap-26c2462549a2/) where seemingly unreasonable behavior improves performance. The difference is that those people have benchmarks and numbers and we have only "it seems so" to back up the claims.

Lagovas (approved):

ilammy (approved):

vixentael (approved):

PR #410, opened 2020-09-15
Add new linters to Acra CircleCI configuration

iamnotacake:
* Fix `golint` not being installed in `prepare.sh` and thus not reporting issues
* Add `misspell` linter to detect common misspells
* Add `ineffassign` linter to detect ineffectual assignments (assigned but not used *or* unintentionally assigned to wrong var because of how scopes and `:=` work in Go)
* Make all these linters write output to console, so we'll see what the issues are by looking at CI log

I'm still unsure whether it's good idea to allow **10** `ineffassign` issues. There are currently 3, and those are intentional `= nil` assignments.
Like, we won't see them until we reach 10, but then we won't be able to merge PR because CircleCI tells us there're too many issues


iamnotacake:
Also, we may need to update Go versions in `config.yml` since they're kinda old.
@Lagovas please confirm if so: which versions to remove, which to add.
I mean `VERSIONS: 1.11.5 1.12` in `general_validation` job.

ilammy (approved):

vixentael (approved):

.circleci/check_ineffassign.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# ignore protobuf-generated code, as well as yacc (.y) parser files
+# duplicate output to stderr so we can see it in console while `wc` is counting the lines
+result=$($GOPATH/bin/ineffassign . | grep -v "\.pb\.go\|\.y" | tee /dev/stderr | wc -l)
+
+if [[ $result -gt 10 ]]; then
```
vixentael, line 7: put 3

vixentael:
> I'm still unsure whether it's good idea to allow 10 ineffassign issues. There are currently 3, and those are intentional = nil assignments.

I suggest to put 3. If someone will miss the legitimate issue, we'll see warning. If someone will introduce intentional issue, they will need to change 3->4.

ilammy:
Nice idea with `tee`! That's gonna be useful to see in the logs.

> I'm still unsure whether it's good idea to allow 10 ineffassign issues. There are currently 3, and those are intentional = nil assignments.

I think that if there are 3 places with intentional assignments then the limit should be set to 3. I believe we tolerate those three but don't really want new code to be adding them. It's better for CI runs to fail immediately rather than someone adding yet another ineffective assignment in their PR, discover that 5 previous PRs added more assignments, and now all of them are a problem.

> Also, we may need to update Go versions in config.yml since they're kinda old.

I believe that if old versions still work then we can keep them (unless they cause other issues, like adding +10 minutes to each build per version). I'd guess that we should be testing at least with the latest 1.15.2, running 1.14.9 may be good as well. Acra does not really use all the latest tech so it's not that pressing to check _all_ the versions, but the latest stable is a must.

Lagovas:
> Also, we may need to update Go versions in `config.yml` since they're kinda old.
> @Lagovas please confirm if so: which versions to remove, which to add.
> I mean `VERSIONS: 1.11.5 1.12` in `general_validation` job.

last 2 versions from https://golang.org/dl/ : 1.15.2, 1.14.9

iamnotacake:
> > Also, we may need to update Go versions in `config.yml` since they're kinda old.
> > @Lagovas please confirm if so: which versions to remove, which to add.
> > I mean `VERSIONS: 1.11.5 1.12` in `general_validation` job.
> 
> last 2 versions from https://golang.org/dl/ : 1.15.2, 1.14.9

OK, but should we leave any of two old versions?

Also, and more importantly, tests are failing on my machine with 1.15.2 (`circleci local execute --job general_validation`).
I put only `1.15.2` in `VERSIONS` and `/home/user/go_root_1.15.2/go` in `GOROOT` respectively.
With `1.14.9` in both places, tests are passing. The errors are related to KeyStore:
```
=== RUN   TestKeyStoreInMemory/TestKeyStoreCleanImport
    keyStore.go:264: incorrect imported list: [keyring/public keyring/symmetric keyring/keypair]
=== RUN   TestKeyStoreInMemory/TestKeyStoreDuplicateImport
    keyStore.go:330: cannot open key ring with key pair: key path does not exist
=== RUN   TestKeyStoreInMemory/TestKeyStoreDuplicateImportSkip
    keyStore.go:382: incorrect imported list: [keyring/public keyring/symmetric keyring/keypair]
=== RUN   TestKeyStoreInMemory/TestKeyStoreDuplicateImportOverwrite
    keyStore.go:517: incorrect imported list: [keyring/symmetric keyring/keypair keyring/public]
=== RUN   TestKeyStoreInMemory/TestKeyStoreListing
=== RUN   TestKeyStoreInMemory/TestKeyStoreExportPublicOnly
time="2020-09-15T15:00:00Z" level=info msg="Key not exported: keyring/symmetric (key has no public data)" service=keystore
    keyStore.go:588: incorrect imported list: [keyring/public keyring/keypair]
--- FAIL: TestKeyStoreInMemory (0.00s)
    --- FAIL: TestKeyStoreInMemory/TestKeyStoreCleanImport (0.00s)
    --- FAIL: TestKeyStoreInMemory/TestKeyStoreDuplicateImport (0.00s)
    --- FAIL: TestKeyStoreInMemory/TestKeyStoreDuplicateImportSkip (0.00s)
    --- FAIL: TestKeyStoreInMemory/TestKeyStoreDuplicateImportOverwrite (0.00s)
    --- PASS: TestKeyStoreInMemory/TestKeyStoreListing (0.00s)
    --- FAIL: TestKeyStoreInMemory/TestKeyStoreExportPublicOnly (0.00s)
=== RUN   TestKeyStoreFilesystem
=== RUN   TestKeyStoreFilesystem/TestKeyStoreCleanImport
    keyStore.go:264: incorrect imported list: [keyring/public keyring/symmetric keyring/keypair]
=== RUN   TestKeyStoreFilesystem/TestKeyStoreDuplicateImport
    keyStore.go:330: cannot open key ring with key pair: key path does not exist
=== RUN   TestKeyStoreFilesystem/TestKeyStoreDuplicateImportSkip
    keyStore.go:382: incorrect imported list: [keyring/public keyring/symmetric keyring/keypair]
=== RUN   TestKeyStoreFilesystem/TestKeyStoreDuplicateImportOverwrite
    keyStore.go:517: incorrect imported list: [keyring/symmetric keyring/keypair keyring/public]
=== RUN   TestKeyStoreFilesystem/TestKeyStoreListing
=== RUN   TestKeyStoreFilesystem/TestKeyStoreExportPublicOnly
time="2020-09-15T15:00:00Z" level=info msg="Key not exported: keyring/symmetric (key has no public data)" service=keystore
    keyStore.go:588: incorrect imported list: [keyring/public keyring/keypair]
```
don't look at `0.00s`, I replaced timings with them for more simple side-by-side diff

.circleci/check_ineffassign.sh
```diff
@@ -0,0 +1,13 @@
...
+  echo "Too many ineffassign issues: $result"
+  exit 1;
```
shadinua, line 9: The semicolon `;` is unnecessary here and below. Usually it should not be placed when we use code formatting with line breaks.

shadinua (approved):

Lagovas:
> incorrect imported list

@ilammy , did you see such errors before?

iamnotacake:
> > incorrect imported list
> 
> @ilammy , did you see such errors before?

No, I didn't. Most of the time I was only testing linters so the code doing `check_gotest.sh` and `.../tests.py` was commented. But when it wasn't, tests did not fail. I checked thrice that it succeeds with `1.14` and fails with `1.15`, so I don't think there's something obvious I'm missing. Tests with Go `1.11.5` and `1.12` were always successful as far as I remember.

BTW, it these errors are because of me missing something *or* if there will be a different PR to fix them, what do you think about putting `VERSIONS: 1.11.5 1.14.9` in `config.yml`? I mean `1.14` instead of `1.12` and leave `1.11` as is to see how far can we go being compatible with `1.11`.

ilammy:
> > incorrect imported list
>
> @ilammy , did you see such errors before?

Nope, me neither. It seems like some regression with 1.15. Or rather, a sleeper bug, most probably. Right now I can reproduce the bug with 1.15, but not 1.14.

ilammy:
@iamnotacake,

> it these errors are because of me missing something or if there will be a different PR to fix them

I believe the best way would be to add 1.14 in this PR, and the PR which fixes the issue can bump that to 1.15, once it's fixed.

ilammy:
PR i#412 has been merged. @iamnotacake, please pull in updates from master. That should resolve the test issue with Go 1.15.

iamnotacake:
> PR #412 has been merged. @iamnotacake, please pull in updates from master. That should resolve the test issue with Go 1.15.

Indeed, your fix solves the problem :+1: 
Now, I'm gonna update this PR, including addition of Go `1.15.2` to `VERSIONS`

vixentael (approved):

ilammy (approved):

PR #412, opened 2020-09-15
Fix Go 1.15 regression in keystore tests

ilammy:
[Go 1.15 has changed serialization of "SET OF" elements in ASN.1][1] to be sorted. This means that now the serialized form of exported key rings does not preserve the order in which they have been given to the `Export()` method. Therefore, we cannot depend on `Import()` to return their names after import in the same order. Replace the slice comparison with unordered comparison.

Some tests also depended on keys being exported and imported sequentially in the given order. This is no longer the case with Go 1.15 so tweak the tests to avoid such dependency.

[1]: https://golang.org/doc/go1.15#encoding/asn1

This issue has been found by @iamnotacake in #410 and is actual only for Go 1.15+. ~~This PR should wait until that one is merged, and bump the Go version to 1.15 for CircleCI to run the tests with up to date Go.~~

This affects user-observable behavior only in pathological cases: invalid imports will fail in a different way. However, the happy case is not affected by Go version used to build Acra.

### DIFF
--- a/keystore/v2/keystore/api/tests/keyStore.go
+++ b/keystore/v2/keystore/api/tests/keyStore.go
@@ -20,6 +20,7 @@ package tests
 import (
 	"crypto/subtle"
 	"errors"
+	"sort"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ilammy:
Or you know... Maybe we can merge this PR first to support Go 1.15, and then #410 will bump Go to that version with linters.

Lagovas (approved):

vixentael:
That's a funny one. Thank you @iamnotacake for pointing out and @ilammy for fixing!

PR #414, opened 2020-09-18
AcraTranslator refactoring: service start up and shutdown, signals listener

storojs72:
This PR introduces several updates for our AcraTranslator service. 

In general I refactored main component (`ReaderServer`) and added support of SIGHUP handling to the service. What is refactored: 
- Added new `StartFromFileDescriptor` function that allows to run HTTP and gRPC servers
having listener obtained from correspondent descriptor, instead of creating a new ones in order to preserve
current listeners in new process;
- Splitted monster `Start` function on lower components `startHTTP`, `startGRPC` in order
to improve readability;
- Wrapped `Start` and `StartFromFileDescriptor` with timeout exit (single point) and avoiding
hidden `os.Exit` calls. This is extremely important for EE version that relies on CE ReaderServer
component;
- Moved from 'multiply listeners - multiply connections' to 'single listener - multiply connection'
with transparent management of stopping/closing it. This is also important for better controlling of
code execution;
- Added some getters which is needed by EE version.

There are two questions about previous implementation that were not obvious for me (most likely they are addressed to @Lagovas), and probably we can discuss it while review process: 1) why previously we had 'multiply listeners - multiply connection' model (when we called `HandleClientConnection`, it was a `AcceptConnection` function inside that created internally new listener and returned accepted connection via channel, their closing was managed via child contexts cancelling in separate goroutine) if `listener` itself allows calling `Accept` function safely from multiply goroutines; and 2) why does switching to new process works, if file descriptors created while SIGHUP processing and on service starting are different? (I asked @Lagovas about it and we came to the point that we know order number of Acra descriptor - it's 3, but it seems that we need not the order number but correct actual value of `fd` to start)

How did I test this? Manually. I created a following methodology for testing it (locally run AcraTranslator, AcraConnector and 2 toy applications: one for sending HTTP and second - for gRPC requests):

1) Test HTTP component without/with active connections;
2) Test gRPC component without/with active connections;
3) Test HTTP + gRPC components that run simultaneously without/with active connections;

For sending HTTP requests I used the following piece of code (thanks to our documentation - but some examples are outdated there):
```
for run in {1..100}; 
do curl -X POST --data-binary @client.acrastruct --header "Content-Type: application/octet-stream" http://127.0.0.1:8000/v1/decrypt; 
done
```

For sending gRPC requests I used following go code:
```
conn, err := grpc.Dial("127.0.0.1:8001", grpc.WithInsecure())
	if err != nil {
		t.Fatal(err)
	}
	defer conn.Close()


	client := grpc1.NewReaderClient(conn)

	for start := time.Now(); time.Since(start) < time.Second * 10; {
		_, _ = client.Decrypt(context.Background(),
			&grpc1.DecryptRequest{
				ClientId:   []byte("testclientid"),
				ZoneId:     []byte("test_zone"),
				Acrastruct: []byte("blabla"),
			},
			grpc.EmptyCallOption{})
	}
```
While those applications were running, I sent (multiply times) SIGHUP signal to the AcraTranslator and observed that service restarted and there were no cut connections. Testing without connections are rather straightforward.


shadinua:
> For sending HTTP requests I used the following piece of code (thanks to our documentation - but some examples are outdated there):
> 
> ```
> for run in {1..100}; 
> do curl -X POST --data-binary @client.acrastruct --header "Content-Type: application/octet-stream" http://127.0.0.1:8000/v1/decrypt; 
> done
>```

In the example above we send requiests sequentially. I'd suggest to make a simple, but effective impromevent:
```bash
TEST_THREADS=20
TEST_CYCLES=100
for i in $(seq 1 $TEST_THREADS); do
    for j in $(seq 1 $TEST_CYCLES); do
        curl -X POST \
            --data-binary @client.acrastruct \
            --header "Content-Type: application/octet-stream" \
            http://127.0.0.1:8000/v1/decrypt
    done &
done
```

cmd/acra-translator/acra-translator.go
```diff
@@ -162,6 +172,56 @@ func main() {
...
+	go sigHandlerSIGHUP.RegisterWithContext(mainContext)
+	sigHandlerSIGHUP.AddCallback(func() {
+		log.Infof("Received incoming SIGHUP signal")
+		log.Debugf("Stop accepting new connections, waiting until current connections close")
+		readerServer.StopListeners()
+
+		// Set env flag for forked process
+		if err := os.Setenv(GracefulEnv, "true"); err != nil {
+			log.WithError(err).Errorln("Unexpected error on os.Setenv, graceful restart won't work. Please check env variables, especially gracefulEnv")
+		}
```
ilammy, line 184: Do I understand it correctly that if this fails then we still go on with restarting AcraTranslator? It will not reuse the listening sockets and will try to reopen them. It will probably not work since this process is still active, but I guess we give it a chance? Why not just fail immediately like we do with all errors that follow?

I guess we can't bail out at this point since `readerServer` has been already stopped, so our best bet is to die with an error and rely on service manager to restart AcraTranslator.

(Also, I believe the log message should use `GracefulEnv` constant value, not just `gracefulEnv` string.)

cmd/acra-translator/acra-translator.go
```diff
@@ -162,6 +172,56 @@ func main() {
...
+
+		var fdHTTP, fdGRPC uintptr
+		if listener := readerServer.GetHTTPListener(); listener != nil {
+			fdHTTP, err = network.ListenerFileDescriptor(readerServer.GetHTTPListener())
```
ilammy, line 188: You can use `listener` variable inside the if clause instead of getting it again.

cmd/acra-translator/acra-translator.go
```diff
@@ -191,7 +255,20 @@ func main() {
...
+		if *incomingConnectionGRPCString != "" && *incomingConnectionHTTPString != "" {
+			readerServer.StartFromFileDescriptor(mainContext, DescriptorHTTP, DescriptorGRPC)
+		} else {
+			if *incomingConnectionHTTPString != "" {
+				readerServer.StartFromFileDescriptor(mainContext, DescriptorHTTP, 0)
+			}
+			if *incomingConnectionGRPCString != "" {
+				readerServer.StartFromFileDescriptor(mainContext, 0, DescriptorGRPC)
+			}
+		}
```
ilammy, line 268: I don't quite understand what happens here. Could you please elaborate?

What I find strange is that zero is a valid file descriptor value (usually stdin), I don't think it's okay to use it like this.

network/utils.go
```diff
@@ -103,8 +104,11 @@ func (listener *safeCloseListener) Close() error {
 
 // Accept proxy call to wrapped listener and wrapp accepted connection with safeCloseConnection
 func (listener *safeCloseListener) Accept() (net.Conn, error) {
-	conn, err := listener.Listener.Accept()
-	return newSafeCloseConnection(conn), err
+	if listener.Listener != nil {
+		conn, err := listener.Listener.Accept()
+		return newSafeCloseConnection(conn), err
+	}
+	return nil, errors.New("nil listener")
```
ilammy, line 111: I'd suggest to extract the error into

```go
var ErrNilListener = errors.New("nil listener")
```

cmd/acra-translator/common/network.go
```diff
@@ -18,55 +18,47 @@ package common
...
-		return nil, err
-	}
 
-	// run goroutine that just accept connections and return them and stop on error. you can stop it by closing listener
+	// Run goroutine that just accept connections and return them and stop on error.
+	// We need to have a control under this goroutine while shutdown process but we can't rely on
+	// parentContext.Done since this event comes later that closing listener that accepts (we use listener
+	// provided by caller, and caller is responsible for closing it).
+	// For this reason we assume that *net.OpError == "accept" in case of accepting error
+	// is normal case when shutdown is invoked
```
ilammy, line 36: So... If understand correctly, previously closing the listener resulted in an `accept` error being sent into `errCh`, and now we avoid that. However, the context is basically useless now. Well... This is a weird flex. I don't see any immediate way to improve it, but I don't like it being counterintuitive: usually methods which accept a context can be expected to be cancelable via that context. This is not the case here.

cmd/acra-translator/server/server.go
```diff
@@ -68,15 +70,13 @@ func NewReaderServer(config *common.AcraTranslatorConfig, keystorage keystore.Tr
...
-	}
+	server.StopListeners()
+
 	// non block stop
 	if server.grpcServer != nil {
-		server.waitGroup.Add(1)
+		server.backgroundWorkersSync.Add(1)
 		go func() {
-			defer server.waitGroup.Done()
+			defer server.backgroundWorkersSync.Done()
```
ilammy, line 79: I don't see how this renaming improves readability very much, but okay...

ilammy (commented):
I have yet to read `cmd/acra-translator/server/server.go` properly. Here are some drive-by comments.

cmd/acra-translator/acra-translator.go
vixentael: Agree, I'd suggest to use existing `listener`.

cmd/acra-translator/acra-translator.go
```diff
@@ -162,6 +172,56 @@ func main() {
...
+				log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCantGetFileDescriptor).
+					Fatalln("System error: failed to get acra-socket file descriptor for HTTP:", err)
+
+			}
+		}
+		if listener := readerServer.GetGRPCListener(); listener != nil {
+			fdGRPC, err = network.ListenerFileDescriptor(readerServer.GetGRPCListener())
+			if err != nil {
+				log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCantGetFileDescriptor).
+					Fatalln("System error: failed to get acra-socket file descriptor for GRPC:", err)
```
vixentael, line 199: What is "acra-socket"?
I mean from user perspective. What users should think about when they read "failed to get acra-socket file descriptor ..". Maybe use just "failed to get Acra file descriptor .."?

cmd/acra-translator/acra-translator.go
```diff
@@ -162,6 +172,56 @@ func main() {
...
+		if err != nil {
+			log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCantForkProcess).
+				Fatalln("System error: failed to fork new process", err)
+		}
+
+		log.Infof("%s process forked to PID: %v", ServiceName, fork)
+
+		// Wait a maximum of N seconds for existing connections to finish
+		if utils.WaitWithTimeout(readerServer.GetConnectionManager().WaitGroup, time.Duration(*closeConnectionTimeout)*time.Second) {
+			log.Warningf("Server shutdown Timeout: %d active connections will be cut", readerServer.GetConnectionManager().Counter)
```
vixentael, line 219:
```suggestion
			log.Warningf("Server shutdown timeout: %d active connections will be cut", readerServer.GetConnectionManager().Counter)
```

cmd/acra-translator/acra-translator.go
```diff
@@ -170,10 +230,14 @@ func main() {
 			os.Exit(1)
 		}
 		log.Infof("Configured to send metrics and stats to `incoming_connection_prometheus_metrics_string`")
-		sigHandlerSIGTERM.AddCallback(func() {
+		prometheusClose := func() {
```
vixentael, line 233: 👍

cmd/acra-translator/common/network.go
```diff
@@ -18,55 +18,47 @@ package common
...
 		for {
 			conn, err := listener.Accept()
 			if err != nil {
-				logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCantAcceptNewConnections).WithError(err).Errorln("Error on accept connection")
-				errCh <- err
-				cancel()
+				// https://stackoverflow.com/questions/13417095/how-do-i-stop-a-listening-server-in-go
+				if operationalError, ok := err.(*net.OpError); ok && operationalError.Op == "accept" {
+					// This is a normal case, when listener is closed while it accepts, so just return
+					logger.Infoln("Listener has been stopped by caller while accepting")
```
vixentael, line 45: Make this log Debug level?
I don't think that end users should see this log.

cmd/acra-translator/server/server.go
```diff
@@ -193,129 +187,216 @@ func (server *ReaderServer) HandleConnectionString(parentContext context.Context
...
 
-func withHandlerName(ctx context.Context, name string) context.Context {
-	return context.WithValue(ctx, handlerName{}, name)
-}
+	// provide way to register new services and custom server
+	if server.config.IncomingConnectionGRPCString() != "" {
+		listener, err := network.Listen(server.config.IncomingConnectionGRPCString())
+		if err != nil {
+			log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorTranslatorCantAcceptNewHTTPConnection).
+				Errorln("Can't create HTTP listener from specified connection string")
```
vixentael, line 224: copy paste detected :)
```suggestion
				Errorln("Can't create gRPC listener from specified connection string")
```

cmd/acra-translator/server/server.go
```diff
@@ -193,129 +187,216 @@ func (server *ReaderServer) HandleConnectionString(parentContext context.Context
...
+	}
 
-func withHandlerName(ctx context.Context, name string) context.Context {
-	return context.WithValue(ctx, handlerName{}, name)
-}
+	// provide way to register new services and custom server
+	if server.config.IncomingConnectionGRPCString() != "" {
+		listener, err := network.Listen(server.config.IncomingConnectionGRPCString())
+		if err != nil {
+			log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorTranslatorCantAcceptNewHTTPConnection).
```
vixentael, line 223:
```suggestion
			log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorTranslatorCantAcceptNewGRPCConnection).
```

cmd/acra-translator/server/server.go
```diff
@@ -193,129 +187,216 @@ func (server *ReaderServer) HandleConnectionString(parentContext context.Context
...
+					Errorln("Took error on handling HTTP requests")
+				server.Stop()
+				errCh <- err
+			}
+			return
+		}
+	}()
+}
+
+func (server *ReaderServer) startGrpc(logger *log.Entry, decryptorData *common.TranslatorData, errCh chan<- error, listener net.Listener) {
```
vixentael, line 271:
```suggestion
func (server *ReaderServer) startGRPC(logger *log.Entry, decryptorData *common.TranslatorData, errCh chan<- error, listener net.Listener) {
```

I see that functions and variables in code use HTTP and GRPC as ALL CAPS format, let's continue doing so.

cmd/acra-translator/server/server.go
```diff
@@ -193,129 +187,216 @@ func (server *ReaderServer) HandleConnectionString(parentContext context.Context
...
+				return
+			}
+			opts = append(opts, grpc.Creds(wrapper))
+		}
+
+		server.listenerGRPC = listener
+		grpcListener := common.WrapListenerWithMetrics(listener)
+		grpcServer, err := server.grpcServerFactory.New(decryptorData, opts...)
+		if err != nil {
+			logger.WithError(err).Errorln("Can't create new grpc server")
```
vixentael, line 295:
```suggestion
			logger.WithError(err).Errorln("Can't create new gRPC server")
```

cmd/acra-translator/server/server.go
```diff
@@ -193,129 +187,216 @@ func (server *ReaderServer) HandleConnectionString(parentContext context.Context
...
-				log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorTranslatorCantHandleHTTPConnection).
-					Errorln("Took error on handling HTTP requests")
-				server.Stop()
-				os.Exit(1)
-			}
-		}()
+		// create HTTP listener from correspondent file descriptor
+		file := os.NewFile(fdHTTP, "/tmp/acra-translator_http")
+		if file == nil {
+			logger.Errorln("Can't create new file from descriptor for acra listener")
```
vixentael, line 348:
```suggestion
			logger.Errorln("Can't create new file from descriptor for Acra HTTP listener")
```

cmd/acra-translator/server/server.go
```diff
@@ -193,129 +187,216 @@ func (server *ReaderServer) HandleConnectionString(parentContext context.Context
...
-				opts = append(opts, grpc.Creds(wrapper))
-			}
-
-			grpcListener := common.WrapListenerWithMetrics(listener)
-
-			grpcServer, err := server.grpcServerFactory.New(decryptorData, opts...)
+		// load gRPC listener from correspondent file descriptor
+		file := os.NewFile(fdGRPC, "/tmp/acra-translator_grpc")
+		if file == nil {
+			logger.Errorln("Can't create new file from descriptor for acra listener")
```
vixentael, line 373:
```suggestion
			logger.Errorln("Can't create new file from descriptor for Acra gRPC listener")
```

cmd/acra-translator/server/server.go
```diff
@@ -351,3 +432,67 @@ func (server *ReaderServer) processHTTPConnection(parentContext context.Context,
...
+	if server.listenerHTTP != nil {
+		err := stopListener(server.listenerHTTP)
+		if err != nil {
+			log.WithError(err).Warningln("Error occured while stopping HTTP listener")
+		}
+	}
+	if server.listenerGRPC != nil {
+		err := stopListener(server.listenerGRPC)
+		if err != nil {
+			log.WithError(err).Warningln("Error occured while stopping GRPC listener")
```
vixentael, line 468:
```suggestion
			log.WithError(err).Warningln("Error occured while stopping gRPC listener")
```

cmd/acra-translator/acra-translator.go
```diff
@@ -162,6 +172,56 @@ func main() {
...
 
+	go sigHandlerSIGHUP.RegisterWithContext(mainContext)
+	sigHandlerSIGHUP.AddCallback(func() {
+		log.Infof("Received incoming SIGHUP signal")
+		log.Debugf("Stop accepting new connections, waiting until current connections close")
+		readerServer.StopListeners()
+
+		// Set env flag for forked process
+		if err := os.Setenv(GracefulEnv, "true"); err != nil {
+			log.WithError(err).Errorln("Unexpected error on os.Setenv, graceful restart won't work. Please check env variables, especially gracefulEnv")
```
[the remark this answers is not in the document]
vixentael: or, as @ilammy mentioned, log value of `GracefulRestartEnv` constant direcly

cmd/acra-translator/server/server.go
```diff
@@ -115,30 +108,31 @@ func (server *ReaderServer) HandleConnectionString(parentContext context.Context
...
 		for {
 			var connection net.Conn
 			select {
 			case connection = <-connectionChannel:
 				break
 			case <-parentContext.Done():
-				logger.WithError(parentContext.Err()).Debugln("Stop wrapping new connections")
+				if !errors.As(parentContext.Err(), &context.Canceled) {
+					logger.WithError(parentContext.Err()).Debugln("Stop wrapping new connections")
+				}
```
ilammy, line 130: Doesn't `errors.As` assign to its second argument if the cast is successful? `context.Canceled` is global "constant variable" from the `context` package. Maybe you meant to use `errors.Is` here instead?

There are other places where `errors.As` is used similarly and may need attention.

cmd/acra-translator/server/server.go
```diff
@@ -193,129 +187,216 @@ func (server *ReaderServer) HandleConnectionString(parentContext context.Context
...
-func getHandlerName(ctx context.Context) string {
-	if s, ok := ctx.Value(handlerName{}).(string); ok {
-		return s
+	select {
+	case <-parentContext.Done():
+		break
+	case <-errCh:
+		// error that occurred in background goroutines will be logged, so just break and exit with timeout,
+		// according to defer func
+		break
```
ilammy, line 237: I'd actually log the error here again with some high-level information, indicating that this error has caused the connection to close. The goroutines worked so hard to put this error into the channel, it would be a waste to not use it.

Ditto in `StartFromFileDescriptor`.

cmd/acra-translator/server/server.go
```diff
@@ -193,129 +187,216 @@ func (server *ReaderServer) HandleConnectionString(parentContext context.Context
...
 		// must be last
 		if server.config.StopOnPoison() {
 			log.Infoln("Add poison record callback with AcraTranslator termination")
-			poisonCallbacks.AddCallback(&base.StopCallback{})
+			poisonCallbackStorage.AddCallback(&base.StopCallback{})
 		}
 	}
+}
+
+func (server *ReaderServer) exitWithTimeout() {
```
ilammy, line 324: I'd call it something like `waitForExitTimeout` to avoid making an impression that calling this method *causes* the server to exit.

cmd/acra-translator/server/server.go
```diff
@@ -193,129 +187,216 @@ func (server *ReaderServer) HandleConnectionString(parentContext context.Context
...
-					return
-				}
-				opts = append(opts, grpc.Creds(wrapper))
-			}
-
-			grpcListener := common.WrapListenerWithMetrics(listener)
-
-			grpcServer, err := server.grpcServerFactory.New(decryptorData, opts...)
+		// load gRPC listener from correspondent file descriptor
+		file := os.NewFile(fdGRPC, "/tmp/acra-translator_grpc")
```
ilammy, line 371: While we're here, maybe the `"/tmp/acra-translator_grpc"` should be extracted into some constant. This value is actually irrelevant, it's only a placeholder. But would be nice to indicate to the reader that it's not some random value. `grpcFilenamePlaceholder`? Does anyone have better ideas?

cmd/acra-translator/acra-translator.go
```diff
@@ -46,6 +46,9 @@ import (
 const (
 	ServiceName        = "acra-translator"
 	defaultWaitTimeout = 10
+	GracefulEnv        = "GRACEFUL_RESTART"
+	DescriptorHTTP     = 3
```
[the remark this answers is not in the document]
storojs72: Definitely. But can you please explain once again, why it works if when we fork process (it seems to me) we need _not an order number of descriptor_ but _descriptor value_ to pass, isn't it?

cmd/acra-translator/server/server.go
```diff
@@ -115,30 +108,31 @@ func (server *ReaderServer) HandleConnectionString(parentContext context.Context
...
 			case <-parentContext.Done():
-				logger.WithError(parentContext.Err()).Debugln("Stop wrapping new connections")
+				if !errors.As(parentContext.Err(), &context.Canceled) {
+					logger.WithError(parentContext.Err()).Debugln("Stop wrapping new connections")
+				}
 				return
 			}
 
-			wrappedConnection, clientID, err := server.config.ConnectionWrapper.WrapServer(context.TODO(), connection)
+			connectionContext := context.TODO()
```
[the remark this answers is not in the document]
storojs72: Do you mean that we need an ability to cancel actual connection in such a way? Does it relate to the previous usage of multiply listeners that create connections instead of one listener as currently?

cmd/acra-translator/acra-translator.go
storojs72: Yes. It will use `Start` function without using descriptors. And actually I don't know. I ported it from our AcraServer (https://github.com/cossacklabs/acra/blob/master/cmd/acra-server/acra-server.go#L391). cc @Lagovas

cmd/acra-translator/acra-translator.go
storojs72: That's true. Thank you

network/utils.go
storojs72: Definitely. Thank you

cmd/acra-translator/common/network.go
storojs72: This is an afford to handle a case of `"use of closed network connection"` error. Yes, context is almost useless, except that we extract logger from it (which is also not obvious for me - why)

cmd/acra-translator/acra-translator.go
```diff
@@ -46,6 +46,9 @@ import (
 const (
 	ServiceName        = "acra-translator"
 	defaultWaitTimeout = 10
+	GracefulEnv        = "GRACEFUL_RESTART"
```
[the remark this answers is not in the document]
storojs72: Ok

cmd/acra-translator/acra-translator.go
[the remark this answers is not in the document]
storojs72: Ok. Probably you meant `log.WithError(err).Errorf("Unexpected error on os.Setenv, graceful restart won't work. Please check env variables, especially %s", GracefulRestartEnv)`

cmd/acra-translator/acra-translator.go
storojs72: legacy from `AcraServer`. Yes `failed to get Acra gRPC file descriptor` is really better

cmd/acra-translator/server/server.go
storojs72: ok

cmd/acra-translator/server/server.go
storojs72: ok, thank you

cmd/acra-translator/server/server.go
storojs72: ok

cmd/acra-translator/server/server.go
storojs72: ok, thank you

cmd/acra-translator/server/server.go
storojs72: ok

cmd/acra-translator/server/server.go
storojs72: TBH, I didn't think too much on it. Now after reading descriptions of `Is` and `As` it seems to me that `As` may panic if it can't cast and `Is` will return false in such case, so `Is` is really more appropriate here.

cmd/acra-translator/server/server.go
storojs72: ok

cmd/acra-translator/server/server.go
storojs72: ok, thank you

cmd/acra-translator/server/server.go
storojs72: Yes. It would be better. Thank you

storojs72:
@shadinua thank you for extension of the HTTP app. I will use your version for further testing

cmd/acra-translator/acra-translator.go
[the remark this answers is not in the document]
vixentael: yes, looks good

cmd/acra-translator/acra-translator.go
Lagovas: we should check it in function [StartFromFileDescriptor](https://github.com/cossacklabs/acra/blob/f8ab52028858926fd2ed645dd092a371cea62db5/cmd/acra-translator/server/server.go#L344). Add some value which will represent null as integer value

cmd/acra-translator/acra-translator.go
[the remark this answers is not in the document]
Lagovas: Okay. When we do fork, child process inherit parent's descriptors. And probably all of them will be closed in parent process on exit if they haven't specific flag to not close on exit. Golang has wrapper around fork. It has function which do some preparation before return control, and they duplicate all descriptors which you can specify directly to be sure that they will not be closed.
So, our goal to fork new process and pass descriptors of our current listeners. For example server listener has fd=23 and api listeners has fd=152. So we should tell our child process that it should save and reuse these descriptors and use them as listeners. But every new process has own descriptor table and we will not know which of them responsible for these listeners. So we pass them directly, golang duplicate them and it use first free descriptor values. 3 and 4 (probably previous author of these code checked it and verified that it always use these values:) and we mark them as responsible for server and api listeners, and reuse them.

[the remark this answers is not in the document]
storojs72: Thank you

cmd/acra-translator/acra-translator.go
storojs72: ok. Will change on `Fatalf`

cmd/acra-translator/acra-translator.go
storojs72: Yes. That is weird piece of code that I wrote while debugging. Now I think we can do just `readerServer.StartFromFileDescriptor(mainContext, DescriptorHTTP, DescriptorGRPC)` in any case. Inside `StartFromFileDescriptor` we have `http`, `grpc` conditions that will use appropriate descriptors

cmd/acra-translator/common/network.go
storojs72: @ilammy, now I can't reproduce the `use of closed network connection` error with the EE translator. Probably it was due to some intermediate version of code while debugging, so now I think we can avoid it and use just:
```
conn, err := listener.Accept()
if err != nil {
	logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCantAcceptNewConnections).WithError(err).Errorln("Error on accept connection")
	errCh <- err
	return
}
```
which is much more intuitive

storojs72:
@ilammy, @Lagovas, @vixentael after cleaning the source code from weird leftovers after debugging, it seems to be good to merge. Only SSL issue is left to be resolved

ilammy (approved):
Well, my comments have all been addressed so here's my green thingie :)

There are still a couple of spelling nits from @vixentael though.

> Only SSL issue is left to be resolved for go 1.15.2 in order to pass CircleCI checks

I'm working on it. The issue has been isolated and fixed. I'm doing some final tests and preparing the PR. 

@storojs72, if you wish to merge this PR, you can go ahead and drop 1.15 from tested versions. Or you can wait for that PR to be merged.

Lagovas (approved):
lgtm, but let's merge after fixing @vixentael comments)

storojs72:
@Lagovas , @ilammy can You please point, what should I actually fix, because it seems that I added those changes in https://github.com/cossacklabs/acra/pull/414/commits/3673fcf368ce054660393f5b01922ce0757c999d isn't it?

ilammy:
@storojs72, I've applied the easy ones, changing log lines. It seems that only this one remains: https://github.com/cossacklabs/acra/pull/414#discussion_r491192768
I'm not sure that `EventCodeErrorTranslatorCantAcceptNewGRPCConnection` exists (don't see it it `logging/event_codes.go`), you may need to add it first before updating the log.

ilammy:
Welp, it *does* conflict with recently merged fix for TLS. Okay, let me resolve the merge conflict, and I'll apply the remaining fix suggested by @vixentael along the way.

ilammy:
This should do it. @vixentael, please confirm.

@storojs72, I'm sorry for barging into your branch. Please do `git pull` first if you plan to add something else.

vixentael (approved):
great work, thank you everybody!

storojs72:
Nothing more. Let's merge it

PR #415, opened 2020-09-18
Move AcraServer common files into "common" package

ilammy:
This PR improves code reuse in AcraServer by moving shared code into a separate package which can be imported from Acra EE. (It is not possible to import packages providing binaries.)

The shared code now lives in the `common` module, similar to AcraTranslator. It has been cleaned up and refactored a bit to allow Acra EE to inject its own configuration where necessary.

Here we make no effort to split the ginormous `main` function into more reusable bits. This is an adventure for another day.

There should be no user-visible changes due to this refactoring. More technical details are available in commit messages.

### DIFF
--- a/cmd/acra-server/common/listener.go
+++ b/cmd/acra-server/common/listener.go
@@ -39,8 +39,6 @@ type SServer struct {
 	config                *Config
 	listenerACRA          net.Listener
 	listenerAPI           net.Listener
-	fddACRA               uintptr
-	fdAPI                 uintptr
 	connectionManager     *network.ConnectionManager
 	listeners             []net.Listener
 	errorSignalChannel    chan os.Signal
--- a/cmd/acra-server/common/listener.go
+++ b/cmd/acra-server/common/listener.go
@@ -208,6 +208,16 @@ func (server *SServer) start(listener net.Listener, callback *callbackData, logg
 	}
 }
 
+// ListenerAcra returns listener for AcraServer database connections.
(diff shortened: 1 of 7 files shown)


cmd/acra-server/common/listener.go
```diff
@@ -38,15 +39,16 @@ type SServer struct {
 	config                *Config
 	listenerACRA          net.Listener
 	listenerAPI           net.Listener
-	fddACRA               uintptr
```
Lagovas, line 41: why remove it? it used for sighup handler and fork

ilammy: > why remove it? it used for sighup handler and fork

This ones are unused. See [this commit](https://github.com/cossacklabs/acra/pull/415/commits/ef26077b7ce868a541e16fa4d347638ad97ab48a) which removes all places where `fddACRA` and `fdAPI` are used: they are set to constant values of fds and never used again.

Actual handling of these fds [happens here](https://github.com/cossacklabs/acra/blob/b2cb4f34aa18edb0eb95d18aa641bf7daedbd546/cmd/acra-server/acra-server.go#L374-L396): the fd values are extracted from `net.Listener` objects and passed to the new process after standard fds, as fds 3 and 4, [aka `descriptorAcra` and `descriptorAPI`](https://github.com/cossacklabs/acra/blob/b2cb4f34aa18edb0eb95d18aa641bf7daedbd546/cmd/acra-server/acra-server.go#L68-L69).

vixentael (approved):

PR #418, opened 2020-09-23
Create new Docker image for faster Acra CI tests

iamnotacake:
Currently, all CircleCI jobs (described in `.circleci/config.yml`) are based on fat `cossacklabs/android-build` Docker image.
CircleCI is doing useless job downloading it because it's mostly uneeded (the Android SDK, NDK and VM).
Also there's a script called `prepare.sh` which downloads and installs some Debian packages, as well as Python ones and at least one Go compiler.
So, making one more Docker image, which does not contain Android stuff but contain all the things `prepare.sh` installs would give us nice boost on CI tests speed. It will also be useful for local tests (using CircleCI CLI) since one would be able to run the tests without waiting for `prepare.sh` doing it's job yet another time.

#### Here's some comparison:
`android-build` is the image used as base for all jobs before.
`ci-py-go-themis` is what this PR brings.

| based on Docker image | android-build | ci-py-go-themis |
| --- | --- | --- |
| image size | 7.73GB | 1.95GB |
| has bundled Go | no`[1]` | yes |
| has bundled Go linters | no`[2]` | yes |
| Go versions | configured in `config.yml` | hardcoded in Dockerfile`[3]` |
| has bundled Python | no`[4]` | yes |

`[1]` downloads and unpacks Go tarballs on every job
`[2]` downloads and builds Go linters on every job
`[3]` integration tests find installed Go roots and test with all of them; there was an idea to take Go versions from specific file and use that file in both Docker image build and jobs from `config.yml`
`[4]` runs `apt install python python-setuptools python3 python3-setuptools python3-pip` and installs some additional Python packages on every job
Speaking about time, using `ci-py-go-themis` is faster, but not so much, only because we don't run `prepare.sh` before every job anymore.

Better review this PR commit-by-commit. You'll find some more details in commit messages. If needed, any updates to `ci-py-go-themis.dockerfile` can be easily checked since I can build-n-push `cossacklabs/ci-py-go-themis:latest` from this branch.


iamnotacake:
That's interesting. I've never seen
```
#!/bin/bash -eo pipefail
pg_isready -U${POSTGRES_USER} -d${POSTGRES_DB} -h127.0.0.1

127.0.0.1:5432 - no response
```
while doing local tests. So it's either my changes or CircleCI is not working properly.
`postgresql` passed, but `postgresql-ssl` did not. Both had the same steps, only one different env var.

ilammy (commented):
I have to say that this PR is really well-commented and deserves some kudos! :+1:

Makefile
```diff
@@ -111,9 +120,10 @@ endif
 
 #===== Functions ===============================================================
 
+# set BUILD_CACHE to anything to enable caching
```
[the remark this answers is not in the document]
iamnotacake: Old comment, forgot to remove

.circleci/check_configs.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+export PATH=$GOROOT/bin:$PATH
+
```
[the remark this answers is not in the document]
iamnotacake: `GOROOT` and `GOPATH` are different things. The first contains path to Go toolchain of some version. The second one contains path to executables build by `go get` (currently, few linters). We add `$GOROOT/bin` to `PATH` if we need Go compiler. But the tools from `$GOPATH/bin` are always available.

> Also, what if GOROOT is not set?

`GOROOT` is set in docker image itself. You can redefine it if needed (some scripts do so when iterate over `/usr/lib/go`)

docker/ci-py-go-themis.dockerfile
```diff
@@ -1,3 +1,86 @@
...
+
+RUN GO_PREFIX_DIR=/usr/lib/go/1.15.2 \
+    GO_VERSION=1.15.2 \
+    GO_TARBALL_DIGEST=b49fda1ca29a1946d6bb2a5a6982cf07ccd2aba849289508ee0f9918f6bb4552 \
+    GO_TARBALL_CLEAN=1 \
+    /image.scripts/install_go.sh
+
+ENV GOROOT="/usr/lib/go/1.15.2/go" GOPATH="/home/user/gopath" GO111MODULE="auto"
+# if you need Go, put `export PATH=$GOROOT/bin:$PATH` in you scripts
+# if you need a different Go version, set GOROOT to another /usr/lib/go/*/go before setting PATH
```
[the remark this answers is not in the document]
iamnotacake: > Why not add default Go version to PATH as well?

Because I though it's easier to do `PATH=$GOROOT/bin:$PATH`
than `PATH=$(echo $PATH | sed "s|/usr/lib/go/\d+\.\d+\.\d+/go/bin|$GOROOT/bin|")`
in case you need a custom Go version and don't wanna prepend many different `GOROOT/bin`s in `PATH`

.circleci/config.yml
```diff
@@ -128,8 +128,8 @@ jobs:
     environment:
       FILEPATH_ERROR_FLAG: /tmp/test_fail
       VERSIONS: 1.13.15 1.14.9 1.15.2
-      TEST_MYSQL: true
-      TEST_DB_PORT: 3306
+      TEST_MYSQL: "true"
+      TEST_DB_PORT: "3306"
```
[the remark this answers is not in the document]
iamnotacake: Actually, that was my editor complaining that it expected string as env var value. I guess, it's just implicitly converted when read

.circleci/integration.sh
```diff
@@ -16,25 +16,34 @@ export TEST_OUTPUT_FOLDER="${HOME}/tests_output"
...
 # set correct permissions for ssl keys here because git by default recognize changing only executable bit
 # http://git.661346.n2.nabble.com/file-mode-td6467904.html#a6469081
 # https://stackoverflow.com/questions/11230171/git-is-changing-my-files-permissions-when-i-push-to-server/11231682#11231682
 find tests/ssl -name "*.key" -type f -exec chmod 0600 {} \;
-for version in $VERSIONS; do
-    echo "-------------------- Testing Go version $version"
 
+for GOROOT in $(find /usr/lib/go -maxdepth 2 -path '*.*.*/go'); do
+    version="$(echo $GOROOT | sed -E 's|^/usr/lib/go/([0-9]+\.[0-9]+\.[0-9]+)/go$|\1|')"
+    echo "-------------------- Testing Go version $version at $GOROOT"
```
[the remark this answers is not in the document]
iamnotacake: Good notice about regex, thanks! I guess we'll talk more about going from `for GOROOT in /usr/lib/go` to `VERSIONS`-like as it was before

Lagovas:
> That's interesting. I've never seen
> 
> ```
> #!/bin/bash -eo pipefail
> pg_isready -U${POSTGRES_USER} -d${POSTGRES_DB} -h127.0.0.1
> 
> 127.0.0.1:5432 - no response
> ```
> 
> while doing local tests. So it's either my changes or CircleCI is not working properly.
> `postgresql` passed, but `postgresql-ssl` did not. Both had the same steps, only one different env var.

I missed this message. At first if you do pg_isready to postgresql which require tls then you should pass paths to tls keys and certs. Our postgresql-ssl configured with mutual authentication and expects client's certificate. Plus check do you run container with `-p 5432:5432` which make 5432 port available on host machine.

.circleci/check_configs.sh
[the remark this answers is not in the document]
ilammy: > > Also, what if GOROOT is not set?
> 
> GOROOT is set in docker image itself.

I meant the case when someone wants to execute the script directly, outside of the container context.

docker/ci-py-go-themis.dockerfile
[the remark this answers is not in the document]
ilammy: > Because I though it's easier to do PATH=$GOROOT/bin:$PATH

I mean, it's easier to *not have to do that at all* for plain `go` to just work, launching the default version of Go for the image. When a CI script wants to iterate through multiple versions, it can call them by absolute paths, or prepend `$GOROOT/bin` to PATH as necessary.

iamnotacake:
Spesking about Go versions and controlling them from `config.yml` like before. Here's an idea:
* we install some fresh Go toolchains into `/ust/lib/go` (this is what the `dockerfile` does now)
* we then specify which versions we wanna test against (but let's call it `GO_VERSIONS` instead of `VERSIONS`)
* there gonna be a script (replaces old `prepare.sh`) that checks whether all that versions present in `/usr/lib/go`, if no — downloads them
* in scripts like `check_gotest.sh` and `integration.sh` we iterate over `GO_VERSIONS` instead of dirs in `/usr/lib/go`

However, @shadinua had a different idea:
* create a file where we specify which Go versions are supported (and thus we test them)
* take this file into account while building docker image and while running CircleCI jobs

there's still a possibility that we may wanna test versions that do not exist in current `ci-py-go-themis:latest`, so I guess I have to do step 3 from list above anyway

.circleci/check_gotest.sh
```diff
@@ -2,14 +2,19 @@
 
 # Run test in each go environment and log errors to $FILEPATH_ERROR_FLAG.
 # If all tests pass successfully then the file will not be created at all.
-export GOPATH=$HOME/$GOPATH_FOLDER;
+# export GOPATH=$HOME/$GOPATH_FOLDER
```
vixentael, line 5: why it is commented?

docker/ci-py-go-themis.dockerfile
```diff
@@ -1,3 +1,86 @@
...
+RUN pip3 install --user -r /home/user/python_tests_requirements.txt 
+# run as separate command due to same dependency 'sqlalchemy' to avoid duplicated requirement and error
+RUN pip3 install --user -r $HOME/python_acrawriter_tests_requirements.txt 
+# install from sources because pip install git+https://github.com/mysql/mysql-connector-python not support recursive submodules
+RUN git clone https://github.com/Lagovas/mysql-connector-python && \
+    cd mysql-connector-python && \
+    python3 setup.py clean build_py && \
+    sudo python3 setup.py install_lib && \
+    cd - && \
+    sudo rm -rf mysql-connector-python
```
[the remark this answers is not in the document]
iamnotacake: Doing so would require leaving two `COPY` commands in dockerfile. So you look into the script (suppose it's called `docker/_scripts/ci/install_python_deps.sh`) and don't understand what are those txt files and where do we get them. Then you look into the dockerfile and understand what's happening. That's kinda complex. Or not? Maybe there's a better solution.

.circleci/check_gotest.sh
iamnotacake: Had to remove it, commented instead :c

.circleci/check_gotest.sh
```diff
@@ -2,14 +2,29 @@
 
 # Run test in each go environment and log errors to $FILEPATH_ERROR_FLAG.
 # If all tests pass successfully then the file will not be created at all.
-export GOPATH=$HOME/$GOPATH_FOLDER;
 # cd to code with go.mod file outside of GOPATH to work with module-based behaviour
 # https://github.com/golang/go/wiki/Modules#when-do-i-get-old-behavior-vs-new-module-based-behavior
 cd $HOME/project
```
ilammy, line 7: BTW, don't we already start in this directory? I thought that CircleCI launches the scripts from the source tree directory.

.circleci/check_gotest.sh
```diff
@@ -2,14 +2,29 @@
...
 
-for version in $VERSIONS; do
-    export GOROOT=$HOME/go_root_$version/go;
-    export PATH=$GOROOT/bin/:$PATH;
+OLD_PATH=$PATH
+
+if [ -z "$GO_VERSIONS" ]; then
+    GO_VERSIONS="$(echo $GOROOT | sed -E 's|^/usr/lib/go/([0-9]+\.[0-9]+(\.[0-9]+)?)/go$|\1|')"
+fi
```
ilammy, line 13: It would be nice to add a comment, like "test with all installed Go versions if no specific versions are requested" so that people won't need to decipher the regexp every time.

.circleci/check_gotest.sh
```diff
@@ -2,14 +2,29 @@
...
+
+for go_version in $GO_VERSIONS; do
+    export GOROOT=/usr/lib/go/$go_version/go
+
+    if [ ! -d $GOROOT ]; then
+        echo "Error: Go $go_version is not installed, $GOROOT does not exist"
+        exit 1
+    fi
+
+    export PATH=$GOROOT/bin:$OLD_PATH
```
ilammy, line 23: Please quote the strings, just in case they have spaces in the shell. It's better to overquote in shell.

.circleci/check_gotest.sh
```diff
@@ -2,14 +2,29 @@
...
+
+    echo "Using Go $go_version at $GOROOT"
+    echo PATH=$PATH
```
ilammy, line 26: Also, it might be good idea to dump `$(go version)` output to make sure that Go version is correct.

.circleci/config.yml
```diff
@@ -31,26 +26,29 @@ jobs:
...
-      - run: sudo apt-get install -y postgresql-client
-      - run: pg_isready -U${POSTGRES_USER} -d${POSTGRES_DB} -h127.0.0.1
+      # database in neighbor container need some time to start
+      # wait for the DB to be reachable, up to 15 pings with 1 second interval
+      - run: |
+          set +eo pipefail;
+          for try in {0..15}; do
+            pg_isready -U${POSTGRES_USER} -d${POSTGRES_DB} -h127.0.0.1; [ $? -eq 0 ] && exit 0 || sleep 1;
+          done;
+          exit 1;
```
ilammy, line 51: You don't really need semicolons here, it's a truly multiline script if you use `|` to start the block in YAML.

Also, the `[ $? -eq 0 ]` is redundant, you can use `condition && consequent || alternate` just fine. (The `$?` value is preserved in the clauses, if you need it. Also, using this form removes the need to `set +e`.)

ilammy (commented):
Oh, I see the CI builds started to pass now.

.circleci/check_gotest.sh
iamnotacake: Yeah, okay. Though, it's not "test with all installed Go versions", it's "test with Go version that is set by default in the container"

.circleci/config.yml
iamnotacake: Do you think it's good idea to include one little script in docker container and then call it here and in other jobs?
The script would execute passed command passed amount of times. Just to make `config.yml` more readable, like this:
`try_few_times --count 15 --sleep 1 pg_isready -U${POSTGRES_USER} -d${POSTGRES_DB} -h127.0.0.1`

ilammy: > Do you think it's good idea to include one little script in docker container and then call it here and in other jobs?

In that case I'd rather have it as

```yaml
- run: .circleci/check_postgresql.sh
```

which is much more readable.

.circleci/check_gotest.sh
iamnotacake: Just checked, CircleCI does `cd` into that dir. And it is also specified in env `CIRCLE_WORKING_DIRECTORY=~/project` (that's a default value, see [docs](https://circleci.com/docs/2.0/configuration-reference/#job_name)).
So should we just remove `cd $HOME/project` or put `cd $(eval echo $CIRCLE_WORKING_DIRECTORY)` instead to be sure? We would need that `eval` since it contains `~` which needs to be resolved. I guess just removing `cd` is simpler and won't break anything.

ilammy: > So should we just remove cd $HOME/project

I think this will be the most correct approach. If we don't change the working directory, we can expect it to be the directory with checked out source tree, hence this `cd` is redundant. Removing it will make the script more readable, IMO.

> since it contains ~ which needs to be resolved.

(My Bash works just fine with tildes in variables :thinking:)

.circleci/check_configs.sh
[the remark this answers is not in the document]
iamnotacake: Making all these scripts usable from both CircleCI environment (using docker image `ci-py-go-themis`) and host (using whatever is installed on the system) will be a different PR if it'll be needed

docker/ci-py-go-themis.dockerfile
```diff
@@ -1,3 +1,86 @@
...
+    build-essential \
+    ca-certificates \
+    curl \
+    git \
+    gnupg \
+    libpq-dev \
+    libssl-dev \
+    openssl \
+    psmisc \
+    python python-setuptools \
```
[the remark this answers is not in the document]
iamnotacake: Indeed it's not needed

.circleci/config.yml
```diff
@@ -33,6 +33,8 @@ jobs:
           POSTGRES_USER: test
           POSTGRES_DB: test
     environment:
+      POSTGRES_USER: test
+      POSTGRES_DB: test
```
ilammy, line 37: I'm kinda sad that these variables are not passed to the containers, but nothing we can about it.

.circleci/config.yml
```diff
@@ -44,11 +46,11 @@ jobs:
       # database in neighbor container need some time to start
       # wait for the DB to be reachable, up to 15 pings with 1 second interval
       - run: |
-          set +eo pipefail;
+          set +o pipefail
```
ilammy: I don't see many pipes in this (or any other) snippet... I believe this can be removed as well.

ilammy (approved):

.circleci/check_gotest.sh
ilammy: Oh... Yeah, you're right! Stupid me. I tested only `DIR=~` :facepalm:

ilammy:
I'd wait for @shadinua to give a look, make sure he's satisfied with the CI updates.

.circleci/check_mysql_connection.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+[ -n "$MYSQL_USER" ] || MYSQL_USER=test
+[ -n "$MYSQL_PASSWORD" ] || MYSQL_PASSWORD=test
```
ilammy, line 4: Bash has a cute trick in its sleeve:

```bash
MYSQL_USER="${MYSQL_USER:-test}"
```

which is, in my experience, more conventional for default parameter values.

.circleci/check_mysql_connection.sh
```diff
@@ -0,0 +1,14 @@
...
+
+[ -n "$NUM_PINGS" ] || NUM_PINGS=15
+[ -n "$DELAY" ] || DELAY=1
+
+for try in $(seq $NUM_PINGS); do
+    mysqladmin ping -h127.0.0.1 -u"${MYSQL_USER}" -p"${MYSQL_PASSWORD}" --protocol=TCP && exit 0 || sleep $DELAY
+done
```
ilammy, line 11: Uh... I'd maybe make the database address a parameter as well. Wouldn't cost much here, in a separate file, but who knows when it might come useful. Otherwise this 127.0.0.1 sticks out like a sore thumb. (And I guess it can be `localhost`, to allow IPv6 where available.)

.circleci/config.yml
```diff
@@ -45,12 +45,7 @@ jobs:
       - checkout
       # database in neighbor container need some time to start
       # wait for the DB to be reachable, up to 15 pings with 1 second interval
```
ilammy: I believe this comment should be moved to the scripts. Here the reader does not need know about such intricacies.

.circleci/check_mysql_connection.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
```
ilammy, line 2: I believe

```bash
set -eu
```

at the start of the script won't hurt, and is a good practice.

.circleci/config.yml
Lagovas: will be better if these variables used to configure tests - https://github.com/cossacklabs/acra/blob/master/.circleci/integration.sh#L7
I thought that tests rely on them, but not. To sync them we need pass same values to test

iamnotacake: > will be better if these variables used to configure tests

I guess we can't just take exported variables from script executed in one container and apply them to another container. Also, database containers start before our ones (be it `android-build` or `ci-py-go-themis`). So I guess we shall duplicate credentials for every job: one user+password+... in DB container env and one (with generic names such as `TEST_DB_USER`) in our container env. What do ya think?

.circleci/check_postgresql_connection.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+[ -n "$POSTGRES_USER" ] || POSTGRES_USER=test
```
vixentael, line 3: similar as above

.circleci/check_postgresql_connection.sh
```diff
@@ -0,0 +1,14 @@
...
+[ -n "$POSTGRES_DB" ] || POSTGRES_DB=test
+
+[ -n "$NUM_PINGS" ] || NUM_PINGS=15
+[ -n "$DELAY" ] || DELAY=1
+
+for try in $(seq $NUM_PINGS); do
+    pg_isready -U"${POSTGRES_USER}" -d"${POSTGRES_DB}" -h127.0.0.1 && exit 0 || sleep $DELAY
```
vixentael, line 10: similar as above

docker/ci-py-go-themis.dockerfile
```diff
@@ -0,0 +1,91 @@
...
+    GO_TARBALL_CLEAN=1 \
+    /image.scripts/install_go.sh
+
+RUN GO_PREFIX_DIR=/usr/lib/go/1.15.2 \
+    GO_VERSION=1.15.2 \
+    GO_TARBALL_DIGEST=b49fda1ca29a1946d6bb2a5a6982cf07ccd2aba849289508ee0f9918f6bb4552 \
+    GO_TARBALL_CLEAN=1 \
+    /image.scripts/install_go.sh
+
+ENV GOROOT="/usr/lib/go/1.15.2/go" GOPATH="/home/user/gopath" GO111MODULE="auto"
```
vixentael, line 55: is it ok that we duplicate and hardcode this?

vixentael (commented):
good progress! 

i still see comments from @ilammy that should be fixed

.circleci/check_mysql_connection.sh
```diff
@@ -1,14 +1,21 @@
 #!/usr/bin/env bash
+set -e
 
-[ -n "$MYSQL_USER" ] || MYSQL_USER=test
-[ -n "$MYSQL_PASSWORD" ] || MYSQL_PASSWORD=test
+TEST_DB_HOST="${TEST_DB_HOST:-127.0.0.1}"
```
ilammy, line 4: I guess a variable for port can be added as well. Tests use `TEST_DB_PORT` for that.

I know, we use default ports for PostgreSQL and MySQL, but for the sake of completeness...

ilammy (approved):

docker/ci-py-go-themis.dockerfile
iamnotacake: This line (along with modified `PATH` few lines below) adds some default Go, so it's available out-of-the-box in the container. Only running `install_go.sh` won't affect env vars. We then may redefine `GOROOT` (and modify `PATH`) if some exact version is needed during tests. Not sure whether this is what you were asking about.

.circleci/check_mysql_connection.sh
shadinua: Yes, I agree, it's worth to add an argument for the DB port.

Makefile
```diff
@@ -24,6 +24,15 @@ BUILD_DATE := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
...
 BUILD_DIR_ABS := $(abspath $(BUILD_DIR))
+ifeq ($(DOCKER_BUILD_CACHE),true)
+DOCKER_BUILD_FLAGS += --no-cache=false
+else ifeq ($(DOCKER_BUILD_CACHE),false)
+DOCKER_BUILD_FLAGS += --no-cache=true
+else ifneq ($(DOCKER_BUILD_CACHE),)
+$(error DOCKER_BUILD_CACHE should be either true or false)
+else
+DOCKER_BUILD_FLAGS += --no-cache=true
+endif
```
shadinua, line 35: Probably, we can make this block a bit simpler? Here we also add a comment for adjustable variable (which is expected since we started creating self-documenting Makefiles).
The comment explicitly describes accepted variables and minimizes the risk of mistake.
```suggestion
## Enable Docker build cache (true|false)
DOCKER_BUILD_CACHE ?= false
ifeq ($(DOCKER_BUILD_CACHE),true)
DOCKER_BUILD_FLAGS += --no-cache=false
else
DOCKER_BUILD_FLAGS += --no-cache=true
endif
```

docker/ci-py-go-themis.dockerfile
```diff
@@ -0,0 +1,91 @@
...
+
+# Include helpful scripts
+RUN mkdir /image.scripts
+COPY docker/_scripts/acra-build/install_go.sh /image.scripts/
+RUN chmod +x /image.scripts/*.sh
+
+# Install Go
+RUN GO_PREFIX_DIR=/usr/lib/go/1.13.15 \
+    GO_VERSION=1.13.15 \
+    GO_TARBALL_DIGEST=01cc3ddf6273900eba3e2bf311238828b7168b822bb57a9ccab4d7aa2acd6028 \
```
shadinua, line 39: I'd suggest:
* create a file `docker/_scripts/acra-build/install_go.csums` with checksums in format `<semver2> <checksum_type> <checksum>`
* in the dockerfile iterate over the versions in single `RUN` clause

It make it reusable `install_go.sh` script in other parts without hardcoding checksums and changing them over the whole code each time we change even patch versions.

docker/_scripts/acra-build/install_go.sh
```diff
@@ -4,34 +4,70 @@
...
 
-export PATH="$PATH:${GO_PREFIX_DIR}/go/bin"
-go version
+GO_VERSIONS="${GO_VERSIONS:-1.15.2}"
+
+for GO_VERSION in $GO_VERSIONS; do
+  # gotta replace `.` -> `\.` for AWK regex so it will match exactly what we're asking for
+  go_awk_version="${GO_VERSION//./\\.}"
+
+  GO_TARBALL_DIGEST="$(awk "/^$go_awk_version / { print \$2 }" "$CSUMS_FILE")"
```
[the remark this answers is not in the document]
shadinua: I do not mind this option, but, I must confess, I do not see much benefit from this. In the existing version, the data file looks cleaner, in addition, a hash type field can potentially be added, which in the proposed version will complicate the construction.

shadinua (approved):
Great job, thank you!

vixentael (approved):

PR #419, opened 2020-09-26
Separate TLS config for AcraConnector and DB

ilammy:
Currently AcraServer uses only one TLS configuration for everything. This means, in particular, that both AcraConnector and DB are expected to use the same private CA, if root certificates are not available in the system certificate store. This is not always the case, and it would be nice to allow separate TLS configurations.

This PR adds several new options:

| New option | Old option | Purpose |
| -- | -- | -- |
| `tls_client_ca` | `tls_ca` | path to CAfile with trusted root certificates for AcraConnector verification |
| `tls_client_cert` | `tls_cert` | path to server certificate presented to AcraConnectors |
| `tls_client_key` | `tls_key` | path to private key to the certificate above |
| `tls_client_auth` | `tls_auth` | TLS authentication level when accepting AcraConnector connections |

| New option | Old option | Purpose |
| -- | -- | -- |
| `tls_database_ca` | `tls_ca` | path to CAfile with trusted root certificates for database verification |
| `tls_database_cert` | `tls_cert` | path to client certificate used when connecting to database |
| `tls_database_key` | `tls_key` | path to private key to the certificate above |
| `tls_database_auth` | `tls_auth` | TLS authentication level when connecting to database |
| `tls_database_sni` | `tls_db_sni` | server name expected when connecting to database (renamed option) |

New options override old options. The old options still remains not deprecated, they might be useful as a shortcut.

This required a refactoring to split the TLS configuration, ~~so in the future it will be possible to have more specific configuration (think, different allowed authentication modes, cryptosuites, TLS versions, etc.)~~ which allowed to easily split the configuration.

There are no integration tests for this option currently. (Do we need them?) I have tested it manually, with recently updated certificates. The timing could not have been better...


vixentael (approved):
Looks inspiring

shadinua:
Great, thanks!

Will it be too complex to add a separate configuration for `tls_auth` as well? This would cover most of the inconvenience with the TLS configuration.

ilammy:
> Will it be too complex to add a separate configuration for tls_auth as well?

Will do, refactoring should make that trivial.

ilammy:
@shadinua, here you go. There are now new options `tls_auth_client` and `tls_auth_database` which override the basic level set by `tls_auth` (by default, the highest one).

For example, you might leave the default value of `RequireAndVerifyClientCert` for AcraConnector, but allow the database to just present *any* client certificate with `--tls_auth_database=2`. Or validate client certificates if given, but don't care about AcraConnector ones: `--tls_auth=2 --tls_auth_client=0`.

The common option is still there in case we'd need to add more TLS configs later, like with CA. Right now it can make more sense to configure client *and* database explicitly, rather than configuring this, say, for the client and for everyone else.

cmd/acra-server/acra-server.go
```diff
@@ -108,9 +110,13 @@ func main() {
 	useTLS := flag.Bool("acraconnector_tls_transport_enable", false, "Use tls to encrypt transport between AcraServer and AcraConnector/client")
 	tlsKey := flag.String("tls_key", "", "Path to private key that will be used in AcraServer's TLS handshake with AcraConnector as server's key and database as client's key")
 	tlsCert := flag.String("tls_cert", "", "Path to tls certificate")
-	tlsCA := flag.String("tls_ca", "", "Path to root certificate which will be used with system root certificates to validate Postgresql's and AcraConnector's certificate")
+	tlsCA := flag.String("tls_ca", "", "Path to additional CA certificate for AcraConnector and database certificate validation")
+	tlsClientCA := flag.String("tls_ca_client", "", "Path to additional CA certificate for AcraConnector certificate validation (if different from database CA)")
```
vixentael, line 114:
```suggestion
	tlsClientCA := flag.String("tls_ca_client", "", "Path to additional CA certificate for AcraConnector certificate validation (setup if AcraConnector certificate CA is different from database certificate CA)")
```

cmd/acra-server/acra-server.go
```diff
@@ -108,9 +110,13 @@ func main() {
...
+	tlsDbCA := flag.String("tls_ca_database", "", "Path to additional CA certificate for database certificate validation (if different from AcraConnector CA)")
```
vixentael, line 115:
```suggestion
	tlsDbCA := flag.String("tls_ca_database", "", "Path to additional CA certificate for database certificate validation (setup if database certificate CA is different from AcraConnector certificate CA)")
```

vixentael (approved):

decryptor/mysql/response_proxy.go
```diff
@@ -176,17 +171,27 @@ func NewMysqlProxy(ctx context.Context, decryptor base.Decryptor, dbConnection,
...
+		clientTLSConfig:        tweakTLSConfigForMySQL(setting.ClientTLSConfig()),
+		dbTLSConfig:            tweakTLSConfigForMySQL(setting.DatabaseTLSConfig()),
 		ctx:                    ctx,
 		logger:                 logging.GetLoggerFromContext(ctx),
 		queryObserverManager:   observerManager,
 		decryptionObserver:     base.NewColumnDecryptionObserver(),
 	}, nil
 }
 
+func tweakTLSConfigForMySQL(config *tls.Config) *tls.Config {
```
vixentael, line 186: good naming :)

Lagovas (approved):

ilammy:
After some discussion we have decided to split the remaining options as well. That is, it is possible to set the keys and certificates independently as well. That way we can have different AcraConnector and database TLS configurations if necessary (and there are still common options in case the configuration can be shared).

Given that, the original names for new options were suboptimal. They have been updated for better namespacing:

  - `tls_client_ca`, `tls_client_cert`, `tls_client_key`, `tls_client_auth`
  - `tls_database_ca`, `tls_database_cert`, `tls_database_key`, `tls_database_auth`, `tls_database_sni`

The `tls_db_sni` option has been renamed into `tls_database_sni` for consistency. (The old one is still accepted.)

ilammy:
I don't quite like that `tls_client_cert` option specifies *server* certificate to be used by AcraServer (mandatory parameter), but it's consistent with the original naming idea.

CHANGELOG_DEV.md
```diff
@@ -1,3 +1,14 @@
+## 0.85.0 - 2020-09-28
+
+### Added
+
+- More specific TLS configuration options:
```
vixentael, line 5:
```suggestion
- More specific TLS configuration options that allow to configure separate TLS settings between client app and AcraServer, and between AcraServer and database:
```

vixentael (approved):

shadinua (approved):
Thanks a lot from infrastructure guys! :)

vixentael:
Looks amazing, let's merge and use!

decryptor/base/proxy.go
```diff
@@ -53,8 +54,13 @@ func (p *proxySetting) Censor() acracensor.AcraCensorInterface {
...
 }
 
-// TLSConfig return tls.Config
-func (p *proxySetting) TLSConfig() *tls.Config {
+// ClientTLSConfig return tls.Config to use for database connection, if any.
+func (p *proxySetting) ClientTLSConfig() *tls.Config {
+	return p.tlsConfig
+}
+
+// DatabaseTLSConfig return tls.Config to use for database connection, if any.
```
iamnotacake: (minor) Consider making these comments more specific, such as
> ClientTLSConfig return tls.Config to use for client<>Acra connection, if any.

> DatabaseTLSConfig return tls.Config to use for Acra<>Database connection, if any.

Lagovas (approved):

PR #420, opened 2020-09-30
Extract ClientSession interface

ilammy:
This is a preparatory refactoring for prepared statement implementation.

Prepared statements are inherently associated with a *session* – a connection from the client to the database which is effectively proxied by AcraServer. There is a concept of session in AcraServer's command code (`cmd/acra-server` module), but the decryption proxy code is oblivious to that. It works with just a pair of connections: the one from the client to the AcraServer, and other one from AcraServer to the database.

Here we introduce a more specific concept of client session, defined by the `ClientSession` interface which the proxy code is taught to use. In order to make that happen we need to refactor quite a bunch of code so there are a lot of things happening in this PR.

### How to read this changset

The recommended way to review is by commit. Most of the comments with rational are there.

* **Remove unused fields from ClientSession**
* **Refactor ClientCommandSession construction**

First some unneeded code is removed to make further things cleaner.

* **Move HandleClientConnection to SServer**

Then we move business logic of connection proxying into AcraServer's code (`SServer`) rather than keeping it as an attribute of the client session itself. Now the session is just a passive vessel for two connections and associated context.

* **Export ClientSession's Close() method**
* **Use ClientSession's logger via accessor** 
* **Introduce other ClientSession accessors**

Miscellaneous minor changes to avoid accessing client session data directly. Instead, new accessor methods are introduces (which will later be used by proxies outside of `cmd/acra-server` module).

* **Unique session IDs on ClientSession's logs**

Make sure that each session gets a unique `session_id` field in logs. This makes is really trivial to filter logs for activity of a particular session.

Don't be too smart with GUIDs or something, keep the field as short as possible. We only need it to be unique for a relatively short timeframe.

* **Use ClientSession in session proxy constructors**

Finally, instead of passing just two connections and an unrelated context, pass a `ClientSession` instance over to the proxy. Introduce an interface which allows access to interesting fields of AcraServer's `ClientSession`.

### DIFF
--- a/cmd/acra-server/common/client_commands_session.go
+++ b/cmd/acra-server/common/client_commands_session.go
@@ -45,18 +45,16 @@ const (
 
 // ClientCommandsSession handles Secure Session for client commands API
 type ClientCommandsSession struct {
-	ClientSession
-	Server   *SServer
-	keystore keystore.ServerKeyStore
+	ctx        context.Context
+	server     *SServer
+	config     *Config
+	keystore   keystore.ServerKeyStore
+	connection net.Conn
 }
 
 // NewClientCommandsSession returns new ClientCommandsSession
-func NewClientCommandsSession(keystorage keystore.ServerKeyStore, config *Config, connection net.Conn) (*ClientCommandsSession, error) {
-	clientSession, err := NewClientSession(context.Background(), config, connection)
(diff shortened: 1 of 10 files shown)


Lagovas (approved):
cool

PR #421, opened 2020-10-01
AcraServer refactoring for better shutdown control

storojs72:
This PR introduces refactoring on main component level (`SServer`) of AcraServer service.

I used approach, described in T1661 and already implemented in https://github.com/cossacklabs/acra/pull/414 for AcraTranslator. Just short reminder: we need better control of shutdown process because it is very important for EE version in context of audit logging (where we need to finish execution of service with special service log messages). So here on main component level I added synchronization of spawned goroutines (with a help of `waitGroup` mechanism) and signalling of shutdown invocation with a help of `parentContext` (passed to `SServer` from main function of AcraServer).

For testing I used two versions of AcraServer (CE and EE), according to the procedures described in https://github.com/cossacklabs/acra/pull/414 but with little modification, since we AcraServer works with databases. Test application is the following (basic Go's unit test):

```
func TestDebug(t *testing.T) {
	connectionsNum := 10
	queriesNum := 50000

	var wg sync.WaitGroup
	for i := 0; i < connectionsNum; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			db, err := sql.Open("postgres", "port=9393 user=test dbname=test password=test sslmode=disable")
			if err != nil {
				t.Fatal(err)
			}
			defer db.Close()
			var col string
			for i := 0; i < queriesNum; i++ {
				rows, err := db.Query("SELECT 2 + 2")
				if err != nil {
					t.Fatal(err)
				}
				defer rows.Close()
				for rows.Next() {
					err = rows.Scan(&col)
					if err != nil {
						t.Fatal(err)
					}
				}
			}
		}()
	}
	wg.Wait()
}
```

Here I create 10 simultaneous connections to AcraServer and execute 50000 toy queries (`SELECT 2+2`) inside each of them.
While execution, I send SIGHUP, SIGTERM signals and manually check the behavior of the service. While SIGHUP testing I encountered some errors when connections are cut:
```
time="2020-09-30T20:39:37+03:00" level=error msg="Network error" client_id=client code=1100 error="read tcp 127.0.0.1:9393->127.0.0.1:45902: use of closed network connection" interrupt_side="Client/Connector<->Database"

time="2020-09-30T20:39:37+03:00" level=warning msg="Can't flush writer" client_id=client code=1301 decrypt_mode=wholecell error="write tcp 127.0.0.1:9393->127.0.0.1:45908: use of closed network connection" proxy=server
```

As discussed with @Lagovas this seems to be expected errors, because connections are long-term and when they are cut, proxying functions (`ProxyClientConnection`, `ProxyDatabaseConnection`) that handle encryption/decryption are still try to use connection that is already closed (and proxies don't have information if this connection closing is caused by SIGHUP/SIGTERM, e.g. "legitimate" case or it is some unexpected network error). The rest of system signals handling works as expected. For SIGTERM it gracefully stops service (and we guarantee that all goroutines are finished before os.Exit), and SIGHUP forks child process and exits from parent. 

Changes in CE version of AcraServer also present in this PR (they are minimal), while changes in EE version will be introduced in separate PR to EE repository.

### DIFF
--- a/cmd/acra-server/acra-server.go
+++ b/cmd/acra-server/acra-server.go
@@ -478,7 +478,7 @@ func main() {
 		logging.SetLogLevel(logging.LogDiscard)
 	}
 
-	ctx := context.TODO()
+	ctx := context.Background()
 	if os.Getenv(gracefulEnv) == "true" {
 		if *withZone || *enableHTTPAPI {
 			go server.StartCommandsFromFileDescriptor(ctx, descriptorAPI)
--- a/cmd/acra-server/acra-server.go
+++ b/cmd/acra-server/acra-server.go
@@ -31,6 +31,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"crypto/tls"
 	"flag"
 	"net/http"
@@ -477,16 +478,17 @@ func main() {
 		logging.SetLogLevel(logging.LogDiscard)
 	}
 
+	ctx := context.TODO()
 	if os.Getenv(gracefulEnv) == "true" {
 		if *withZone || *enableHTTPAPI {
-			go server.StartCommandsFromFileDescriptor(descriptorAPI)
+			go server.StartCommandsFromFileDescriptor(ctx, descriptorAPI)
 		}
-		go server.StartFromFileDescriptor(descriptorAcra)
+		go server.StartFromFileDescriptor(ctx, descriptorAcra)
 	} else {
 		if *withZone || *enableHTTPAPI {
-			go server.StartCommands()
+			go server.StartCommands(ctx)
 		}
-		go server.Start()
+		go server.Start(ctx)
 	}
 
 	// on sighup we run callback that stop all listeners (that stop background goroutine of server.Start())
--- a/cmd/acra-server/common/listener.go
+++ b/cmd/acra-server/common/listener.go
@@ -19,10 +19,12 @@ package common
 import (
 	"context"
 	"errors"
+	"github.com/cossacklabs/acra/utils"
 	"io"
 	"net"
 	url_ "net/url"
 	"os"
+	"sync"
 	"syscall"
 	"time"
 
@@ -45,6 +47,8 @@ type SServer struct {
 	errorSignalChannel    chan os.Signal
 	restartSignalsChannel chan os.Signal
 	proxyFactory          base.ProxyFactory
+	backgroundWorkersSync sync.WaitGroup
+	stopListenersSignal   chan bool
 }
 
 // ErrWaitTimeout error indicates that server was shutdown and waited N seconds while shutting down all connections.
@@ -52,12 +56,14 @@ var ErrWaitTimeout = errors.New("timeout")
 
 // NewServer creates new SServer.
 func NewServer(config *Config, proxyFactory base.ProxyFactory, errorChan chan os.Signal, restarChan chan os.Signal) (server *SServer, err error) {
+	stopListenersSignal := make(chan bool)
 	return &SServer{
 		config:                config,
 		connectionManager:     network.NewConnectionManager(),
 		errorSignalChannel:    errorChan,
 		restartSignalsChannel: restarChan,
 		proxyFactory:          proxyFactory,
+		stopListenersSignal:   stopListenersSignal,
 	}, nil
 }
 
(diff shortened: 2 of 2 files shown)


cmd/acra-server/acra-server.go
```diff
@@ -477,16 +478,17 @@ func main() {
 		logging.SetLogLevel(logging.LogDiscard)
 	}
 
+	ctx := context.TODO()
```
ilammy, line 481: Typically, `context.Background()` is used for top-level contexts established in **main** (or in other top-level contexts). `context.TODO()` is intended for method calls deeper in the stack, as a temporary placeholder until a usable context is threaded into the method.

cmd/acra-server/common/listener.go
```diff
@@ -253,31 +264,47 @@ func (server *SServer) processConnection(connection net.Conn, callback *callback
...
-					Errorln("Stop accepting new connections due net.Timeout")
-				return
-			}
-			logger.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCantAcceptNewConnections).
-				Errorln("Can't accept new connection")
+		if err := parentContext.Err(); err != nil {
+			// this means that parentContext is Done.
+			// And here we are not interesting in particular reasons of it, so just return
 			return
-		}
```
ilammy, line 269: Since the positive branch of this `if` ends with a `return`, you don't really need explicit `else`. That makes the code easier to read, and it will probably result in a better diff in this PR.

Also, I'd log the error since it may be interesting why the context has been closed.

cmd/acra-server/common/listener.go
```diff
@@ -291,8 +318,19 @@ func (server *SServer) ListenerAPI() net.Listener {
...
+	// background goroutines spawned by SServer will finish their execution or force their closing.
+	if utils.WaitWithTimeout(&server.backgroundWorkersSync, utils.DefaultWaitGroupTimeoutDuration) {
+		log.Errorf("Couldn't stop all background goroutines spawned by readerServer. Exited by timeout")
+	}
+}
+
 // Start listening connections from proxy
-func (server *SServer) Start() {
+func (server *SServer) Start(parentContext context.Context) {
+	defer server.waitForExitTimeout()
```
ilammy, line 332: Is this in the right place? Do we need to wait even if, say, `network.Listen(...)` fails? Shouldn't this be just a regular call after `server.run(...)`? Or can it be just a part of `run()`?

Ditto in `StartFromFileDescriptor`, `StartCommands`, `StartCommandsFromFileDescriptor` with their respective failures.

cmd/acra-server/common/listener.go
```diff
@@ -443,16 +488,36 @@ func (server *SServer) StartCommandsFromFileDescriptor(fd uintptr) {
...
-		server.errorSignalChannel <- syscall.SIGTERM
 		return
 	}
 	listenerWithFileDescriptor, ok := listenerFile.(network.ListenerWithFileDescriptor)
 	if !ok {
 		logger.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorFileDescriptionIsNotValid).
 			Errorf("System error: file descriptor %d is not a valid socket", fd)
+		server.errorSignalChannel <- syscall.SIGTERM
 		return
 	}
```
ilammy, line 499: Hm... Why signal TERM on line 497 and 484, but not for `net.FileListener` failure?

ilammy (approved):
Overall, LGTM! :+1:

cmd/acra-server/common/listener.go
```diff
@@ -45,19 +47,23 @@ type SServer struct {
...
 // NewServer creates new SServer.
 func NewServer(config *Config, proxyFactory base.ProxyFactory, errorChan chan os.Signal, restarChan chan os.Signal) (server *SServer, err error) {
+	stopListenersSignal := make(chan bool)
 	return &SServer{
 		config:                config,
 		connectionManager:     network.NewConnectionManager(),
 		errorSignalChannel:    errorChan,
 		restartSignalsChannel: restarChan,
 		proxyFactory:          proxyFactory,
+		stopListenersSignal:   stopListenersSignal,
```
Lagovas, line 66: why not just `stopListenersSignal: make(chan bool),`?

cmd/acra-server/common/listener.go
Lagovas: actually to check is context done or not better to use .Done() channel:
```
select {
case <-parentContext.Done():
  return
default:
  break
}
```
if context not finished then default branch will allow to continue execution

cmd/acra-server/common/listener.go
Lagovas: nice catch. better to call it after an initialized listener.

cmd/acra-server/acra-server.go
storojs72: Yes. Thanks

cmd/acra-server/common/listener.go
storojs72: Actually, yes. You are both right

cmd/acra-server/common/listener.go
storojs72: Got it. Thanks

cmd/acra-server/common/listener.go
storojs72: Yeah. While refactoring, I had no clear understanding why do we need this, but later I came up that we have to gracefully shutdown our service in case of errors during listener creation, since we invoke `Start` (and its variations) in separate goroutine in AcraServer (in contrast to AcraTranslator where we don't and consequently are able to just return). Thank you for attentiveness

cmd/acra-server/common/listener.go
storojs72: Yes. That's better. -1 redundant line

Lagovas (approved):

ilammy (approved):
I guess
```diff
-ctx := context.TODO()
+ctx := context.Background()
```
and we're ready to merge?

storojs72:
I will merge it, since PR in EE will depend on master branch of CE. If it will be some more discussions/questions, I can address them in separate PR

PR #422, opened 2020-10-01
Prepared statement and cursor registries

ilammy:
In order to handle prepared statements we need to keep track of them. This is what registries are for.

Implement a couple of simple objects to keep track of name → entity mappings. Both prepared statements and cursors are identified by name. (In PostgreSQL they may be anonymous as well, with empty names.) There are simple accessors to their core properties: the query which is prepared, and the statement which is iterated by a cursor.

These interfaces are pretty much preliminary as I'm not sure what we'll need for the implementation. They may also be unnecessarily extensible. For one, we might not actually need PostgreSQL and MySQL specific registries. For now, keep it like that in case we'd need to add something PostgreSQL-specific. However, if that won't happen then all of this is likely to move to the `base` module as shared utilities.

Expected approvals:
- [X] @Lagovas 
- [x] @iamnotacake

### DIFF
--- a/decryptor/base/proxy.go
+++ b/decryptor/base/proxy.go
@@ -96,8 +96,6 @@ type ClientSession interface {
 
 	PreparedStatementRegistry() PreparedStatementRegistry
 	SetPreparedStatementRegistry(registry PreparedStatementRegistry)
-	CursorRegistry() CursorRegistry
-	SetCursorRegistry(registry CursorRegistry)
 }
 
 // ProxyFactory create new Proxy for specific database
@@ -105,10 +103,13 @@ type ProxyFactory interface {
 	New(clientID []byte, clientSession ClientSession) (Proxy, error)
 }
 
-// PreparedStatementRegistry keeps track of active prepared statements within a ClientSession.
+// PreparedStatementRegistry keeps track of active prepared statements and cursors within a ClientSession.
 type PreparedStatementRegistry interface {
-	Add(statement PreparedStatement) error
+	AddStatement(statement PreparedStatement) error
 	StatementByName(name string) (PreparedStatement, error)
+
+	AddCursor(cursor Cursor) error
+	CursorByName(name string) (Cursor, error)
 }
 
 // PreparedStatement is a prepared statement, ready to be executed.
@@ -117,12 +118,6 @@ type PreparedStatement interface {
 	Name() string
 	Query() sqlparser.Statement
 	QueryText() string
-}
-
-// CursorRegistry keeps track of active cursors within a ClientSession.
-type CursorRegistry interface {
-	Add(cursor Cursor) error
-	CursorByName(name string) (Cursor, error)
 }
 
 // Cursor is used to iterate over a prepared statement.
--- a/decryptor/base/proxy.go
+++ b/decryptor/base/proxy.go
@@ -24,6 +24,7 @@ import (
 	acracensor "github.com/cossacklabs/acra/acra-censor"
 	"github.com/cossacklabs/acra/encryptor/config"
 	"github.com/cossacklabs/acra/keystore"
+	"github.com/cossacklabs/acra/sqlparser"
 )
 
 // ProxySetting provide data access methods for proxy factories
@@ -92,9 +93,26 @@ type ClientSession interface {
 	Context() context.Context
 	ClientConnection() net.Conn
 	DatabaseConnection() net.Conn
+
+	PreparedStatementRegistry() PreparedStatementRegistry
+	SetPreparedStatementRegistry(registry PreparedStatementRegistry)
 }
 
 // ProxyFactory create new Proxy for specific database
 type ProxyFactory interface {
 	New(clientID []byte, clientSession ClientSession) (Proxy, error)
+}
+
+// PreparedStatementRegistry keeps track of active prepared statements within a ClientSession.
+type PreparedStatementRegistry interface {
+	Add(statement PreparedStatement) (bool, error)
+	StatementByName(name string) (PreparedStatement, error)
+}
+
+// PreparedStatement is a prepared statement, ready to be executed.
+// It can be either a textual SQL statement from "PREPARE", or a database protocol equivalent.
+type PreparedStatement interface {
+	Name() string
+	Query() sqlparser.Statement
+	QueryText() string
 }
--- a/decryptor/base/proxy.go
+++ b/decryptor/base/proxy.go
@@ -106,9 +106,11 @@ type ProxyFactory interface {
 // PreparedStatementRegistry keeps track of active prepared statements and cursors within a ClientSession.
 type PreparedStatementRegistry interface {
 	AddStatement(statement PreparedStatement) error
+	DeleteStatement(name string) error
 	StatementByName(name string) (PreparedStatement, error)
 
 	AddCursor(cursor Cursor) error
+	DeleteCursor(name string) error
 	CursorByName(name string) (Cursor, error)
 }
 
--- a/decryptor/base/proxy.go
+++ b/decryptor/base/proxy.go
@@ -107,7 +107,7 @@ type ProxyFactory interface {
 
 // PreparedStatementRegistry keeps track of active prepared statements within a ClientSession.
 type PreparedStatementRegistry interface {
-	Add(statement PreparedStatement) (bool, error)
+	Add(statement PreparedStatement) error
 	StatementByName(name string) (PreparedStatement, error)
 }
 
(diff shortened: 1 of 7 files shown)


decryptor/base/proxy.go
```diff
@@ -92,9 +93,41 @@ type ClientSession interface {
...
 
 // ProxyFactory create new Proxy for specific database
 type ProxyFactory interface {
 	New(clientID []byte, clientSession ClientSession) (Proxy, error)
 }
+
+// PreparedStatementRegistry keeps track of active prepared statements within a ClientSession.
+type PreparedStatementRegistry interface {
+	Add(statement PreparedStatement) (bool, error)
+	StatementByName(name string) (PreparedStatement, error)
```
Lagovas, line 111: actually registry should allow to remove statements too because user can drop existing statements and re-assign queries by statement name. So `Add` operation should update statement if it exists or we should add one more method for updates.

decryptor/postgresql/prepared_statements.go
```diff
@@ -0,0 +1,136 @@
...
+	_, exists := r.registry[name]
+	r.registry[name] = statement
+	return exists, nil
+}
+
+// PgPreparedStatement is a PostgreSQL PreparedStatement.
+type PgPreparedStatement struct {
+	name string
+	text string
+	sql  sqlparser.Statement
```
Lagovas, line 64: just to think about it. probably we will need to add some context storage for statement. for example bound values or our encryption settings related to this prepared statement (but probably not if we will do it in another way and search encryption settings by prepared statement name in other registry)

decryptor/postgresql/prepared_statements.go
```diff
@@ -0,0 +1,136 @@
...
+	name := statement.Name()
+	_, exists := r.registry[name]
+	r.registry[name] = statement
+	return exists, nil
+}
+
+// PgPreparedStatement is a PostgreSQL PreparedStatement.
+type PgPreparedStatement struct {
+	name string
+	text string
```
Lagovas, line 63: what about to use [OnQueryObject](https://github.com/cossacklabs/acra/blob/master/decryptor/base/observers.go#L27) which cache statement (to avoid one more parsing) and re-generate query on demand

decryptor/base/proxy.go
ilammy: > user can drop existing statements and re-assign queries by statement name

Yeah, its possible. However, since it's not strictly necessary for correctness of Acra's processing (database will fail any attempt to use removed statements), removal tracking won't be implemented right away. Though, I do feel that we'd better have it to avoid accumulating named prepared statements, despite the expectation that their number is likely to be limited.

> So Add operation should update statement if it exists or we should add one more method for updates.

In fact, it currently acts more like an `Update`: if there is no statement then it's added to the registry, and if there is then then old one is replaced with the new one. Do you feel that we'd better name this method `Update` instead?

(My idea for the future was to have separate `Add` and `Delete` methods, with `Add` *not* allowing updates. That's because PostgreSQL protocol does not allow to update a statement in place, except for the unnamed statements. Named ones should be first removed, and then re-added under the same name.)

decryptor/postgresql/prepared_statements.go
ilammy: Hm... Does it result in much memory saved? I mean, when we're going to created prepared statement objects we will have the SQL text on hand (from the user data), and we'd need to parse it anyway. They don't seem like massive pieces of data so that we can't keep them both around.

Though, I'm not really sure that we will need both text and the parsed query. I guess, for processing we will be interested in the parsed form, and the text might get useful only for debug logging or something. In that case it kinda makes sense to keep only the parsed form and restore the text from it if necessary.

OnQueryObjects will be used later for processing. I intend them to just refer to existing prepared statement from the registry, not keep a personal copy of the parsed query.

decryptor/base/proxy.go
Lagovas: > removal tracking won't be implemented right away

but why? we can track `DEALLOCATE` commands and remove prepared statement from registry on each one.
But yes, if user will deallocate and try to use deallocated prepared statement, database should return error.
But on other side each such call of prepared statement may cause some actions and calculations from acra-server. And there may be case when client's app has mistake and they use incorectly dropped prepared statements. And if it did without acra-server, they just took errors in logs and work in this state some time.
But with acra-server it may cause performance degradation if app incorrectly use and it cause extra-calculations which may slow other correct queries (for example fetching prepared statement state from distributed storage, fetching keys, censor computations, etc)

> (My idea for the future was to have separate Add and Delete methods, with Add not allowing updates. That's because PostgreSQL protocol does not allow to update a statement in place, except for the unnamed statements. Named ones should be first removed, and then re-added under the same name.)

Probably will be enough to leave Add with update logic (because of unnamed statements) but in implementation of registry we will add check that update allowable only for unnamed statement. If it is not unnamend then return error on try to add statement without `delete`.

decryptor/postgresql/prepared_statements.go
Lagovas: Main goal of adding OnQueryObject was to avoid extra cpu-computations on parsing string every time. And first time it was introduced for censor + encryption processors. Some censors rules work over parsed query, some of them over sql string.
Then it passed to encryption processors which may change query itself and do it over parsed query. And string representation used only on final stage when statement converted to string to place it into output packet to database. Second goal was to avoid synchronization between statement struct and sql query. If each processor accept string + statement, it doesn't know what will used of these in next processor. So if it updated statement, it should update query string and pass it forward.

Probably we should fill our registry by the same flow of processing query. Create some subscriber for `OnQuery` event (which used by other encryption processors) and be sure that it last processor which will accept OnQueryObject and save in registry after all manipulations by previous processors (instead placing it as separate block of code before generating final presentation of packet to database with updated query)

decryptor/base/proxy.go
ilammy: > but why? we can track DEALLOCATE commands and remove prepared statement from registry on each one.

By "later" I mean "probably not in the same PR if it turns out to be nontrivial", not "next year". We'll track removals. The arguments you lay out are the reason. It might not affect correctness (i.e., data observed by the application) but it has the potential to affect performance by consuming compute resources on requests which are bound to fail in the database, by consuming memory resources for effectively dead statements, etc.

I'l add removal methods here in this PR, and will keep the `Add` method with its current logic (always update), but add a TODO to refuse updates for existing named statements when we're sure that `Delete` is going to be executed as required.

Lagovas (approved):

ilammy:
Gee, it turned out to be *a bit* more involved than I expected. While I've been adding removal of cursors and statements I have remembered that cursors of a remove statement should be automatically removed too. This makes cursor and statement registry tightly coupled. So... I have merged them into a single entity which keeps track of both. (Well, as it should have been from the start since cursors don't make much sense without their associated statements.)

For correct removal we need to track which cursors are associated with which prepared statements. One alternative is to keep track of this in `PreparedStatement` interface directly and add cursors there, but this will means that we won't be able to create a cursor beforehand, without entering it into the registry immediately (to wait for database response). So I have opted for a bit more weird implementation where this information is kept separately.

All of this has become quite complex, not your simple map, so I have thrown in some unit tests there as well.

ilammy:
@iamnotacake, it works! :+1:
```
/home/user/project/decryptor/postgresql/prepared_statements.go:121:11: ineffectual assignment to ok
Too many ineffassign issues: 4
```

ilammy:
Thanks to feedback from @iamnotacake, I have realized that we *can* keep cursors in prepared statements directly, which is more natural and may come useful later. It does require a bunch of random downcasts, but it should be okay.

decryptor/postgresql/prepared_statements_test.go
```diff
@@ -0,0 +1,307 @@
...
+	// The cursor should be updated.
+	foundCursor, err := registry.CursorByName("cursor")
+	if err != nil {
+		t.Fatal("cannot look up cursor after update", err)
+	}
+	if foundCursor != cursor2 {
+		t.Error("did not find the same cursor")
+	}
+
+	// Now, the same should work just find if a cursor for different statement reuses the same name,
```
iamnotacake, line 151:
```suggestion
	// Now, the same should work just fine if a cursor for different statement reuses the same name,
```

decryptor/postgresql/prepared_statements_test.go
```diff
@@ -0,0 +1,307 @@
...
+		t.Fatal("cannot add initial statement", err)
+	}
+
+	// Quickly remove it...
+	err = registry.DeleteStatement("statement")
+	if err != nil {
+		t.Fatal("cannot remove statement", err)
+	}
+
+	// There should no statement now.
```
iamnotacake, line 189:
```suggestion
	// There should be no statement now.
```

iamnotacake (approved):

Lagovas (approved):

PR #424, opened 2020-10-07
PgProtocolState handler

ilammy:
Introduce a separate object to take care of PostgreSQL protocol packets. With extended protocol we will need to track both requests and responses before we are able to process a query, so we need a place to store this information temporarily.

The protocol state is fundamentally associated with the session so it makes sense to keep one there. However, we need to keep it opaque as MySQL and PostgreSQL protocols are very different. For now, let's use just `interface{}` as a placeholder. Maybe later we will be able to extract some more specific interface out of the implementations.

`PgProtocolState` keeps user query in memory until the database issues a response. The query text is sensitive information, especially when transparent encryption is in use. We should remove it from memory as soon as it possible. The database sends `ReadyForQuery` when it has completed processing of the previous query, this is the point when we know that the query is no longer needed and can be freed.

Expected review:

- [x] @Lagovas
- [ ] @iamnotacake

### DIFF
--- a/decryptor/postgresql/pg_decryptor.go
+++ b/decryptor/postgresql/pg_decryptor.go
@@ -80,6 +80,9 @@ var (
 	ErrInvalidProtocolState             = errors.New("ClientSession contains invalid ProtocolState")
 )
 
+// ErrBlockedByCensor is returned when a query has been denied by AcraCensor.
+var ErrBlockedByCensor = NewPgError("AcraCensor blocked this query")
+
 // PgSQL constant sizes and types.
 const (
 	// DataRowLengthBufSize each postgresql packet contain 4 byte that store length of message contents in bytes, including self
@@ -321,7 +324,7 @@ func (proxy *PgProxy) handleQueryPacket(packet *PacketHandler, logger *log.Entry
 }
 
 func (proxy *PgProxy) sendClientAcraCensorError(logger *log.Entry) error {
-	errorMessage, err := NewPgError("AcraCensor blocked this query")
+	errorMessage, err := ErrBlockedByCensor
 	if err != nil {
 		logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCodingPostgresqlCantGenerateErrorPacket).
 			WithError(err).Errorln("Can't create PostgreSQL error message")
--- a/decryptor/postgresql/pg_decryptor.go
+++ b/decryptor/postgresql/pg_decryptor.go
@@ -228,44 +228,38 @@ func (proxy *PgProxy) ProxyClientConnection(errCh chan<- error) {
 			return
 		}
 		proxy.dbConnection.SetWriteDeadline(time.Now().Add(network.DefaultNetworkTimeout))
-		// we are interested only in requests that contains sql queries
-		if !(packet.IsSimpleQuery() || packet.IsParse()) {
-			if err = packet.sendPacket(); err != nil {
+
+		_, censorSpan := trace.StartSpan(packetSpanCtx, "censor")
+
+		// Let the protocol observer take a look at the packet, keeping note of it.
+		action, err := proxy.protocolState.HandleClientPacket(packet)
+		if err != nil {
+			errCh <- err
+			return
+		}
+		switch action {
+		case QueryPacket:
+			// If that's some sort of a packet with query data inside it, go on...
+
+		default:
+			// Pass all other uninteresting packets through to the database without processing.
+			if err := packet.sendPacket(); err != nil {
 				logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorNetworkWrite).WithError(err).Errorln("Can't forward packet to db")
 				errCh <- err
 				return
 			}
-			if packet.terminatePacket {
+			// If this should be the last packet, signal about it and stop listening.
+			if action == TerminationPacket {
 				errCh <- io.EOF
 				return
 			}
 			continue
 		}
-		_, censorSpan := trace.StartSpan(packetSpanCtx, "censor")
-		var query string
-		if packet.IsSimpleQuery() {
-			query, err = packet.GetSimpleQuery()
-			if err != nil {
-				logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCodingPostgresqlCantExtractQueryString).WithError(err).Errorln("Can't fetch query string from Query packet")
-				errCh <- err
-				return
-			}
-		} else if packet.IsParse() {
-			query, err = packet.GetParseQuery()
-			if err != nil {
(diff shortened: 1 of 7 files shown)


ilammy:
After more work on extended protocol processing (outside of this PR) I've got fed up with all this duplication of packet forwarding, timer stopping, error logging and handling. I've extracted more code into utility methods so that packet forwarding after processing happens only in one place. That makes the code more clear and readable, in my opinion.

decryptor/postgresql/protocol.go
```diff
@@ -0,0 +1,99 @@
...
+		return QueryPacket, nil
+	}
+
+	// We are not interested in other packets, just pass them through.
+	return OtherPacket, nil
+}
+
+// HandleDatabasePacket observes a packet with database response,
+// extracts useful information from it, and confirms client requests.
+func (p *PgProtocolState) HandleDatabasePacket(packet *PacketHandler) (PacketType, error) {
```
Lagovas, line 84: imho, will be more predictable if `Handle*` will do something and return just `error`, and then callee can get info by `packet.GetPacketType()`.

decryptor/postgresql/pg_decryptor.go
```diff
@@ -214,89 +229,108 @@ func (proxy *PgProxy) ProxyClientConnection(errCh chan<- error) {
...
 
-		if err := packet.sendPacket(); err != nil {
-			logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorNetworkWrite).WithError(err).Errorln("Can't send packet")
-			errCh <- err
-			return
+	default:
+		// Forward all other uninteresting packets to the database without processing.
+		return false, nil
+	}
+	query := proxy.protocolState.PendingQuery()
```
Lagovas, line 284: imho better to move it to separate function `handleQueryPacket` and call it in `case QueryPacket`. Because in a future we will handle `execute` and other packets and probably will be handled in a separate case.

decryptor/postgresql/pg_decryptor.go
```diff
@@ -523,54 +557,66 @@ func (proxy *PgProxy) ProxyDatabaseConnection(errCh chan<- error) {
...
+	case DataPacket:
+		// If that's some sort of a packet with response data inside it, go on...
+
+	default:
+		// Forward all other uninteresting packets to the client without processing.
+		return nil
+	}
+
+	logger.Debugln("Matched data row packet")
+	if err = packet.parseColumns(); err != nil {
```
Lagovas, line 594: same here with data packet. better to encapsulate to separate function and call it in correct case branch.

decryptor/postgresql/protocol.go
ilammy: Hm... Yeah, okay. I wanted to return the type so that it's more closely associated with the packet. However, `PgProtocolState` is inherently stateful so something  like `LastPacketType()` should be okay and consistent with how the other information is accessed.

decryptor/postgresql/pg_decryptor.go
ilammy: Mmm... Yeah, makes sense. I intended to keep the handling of other query-like packet similar, but it very much makes sense to have processing as

```go
switch type {
case This:
        handleThis()
case That:
        handleThat()
default:
        handleElse()
}
```
rather than the current
```go
switch type {
case This:
        // break out of switch
default:
        handleElse()
}
handleThis()
```

decryptor/postgresql/pg_decryptor.go
```diff
@@ -305,6 +302,7 @@ func (proxy *PgProxy) ProxyClientConnection(errCh chan<- error) {
 
 		censorSpan.End()
 
+		// Forwared the (modified) query to the database.
```
iamnotacake:
```suggestion
		// Forward the (modified) query to the database.
```

decryptor/postgresql/pg_decryptor.go
```diff
@@ -232,84 +232,105 @@ func (proxy *PgProxy) ProxyClientConnection(errCh chan<- error) {
 
 		_, censorSpan := trace.StartSpan(packetSpanCtx, "censor")
 
-		// Let the protocol observer take a look at the packet, keeping note of it.
-		action, err := proxy.protocolState.HandleClientPacket(packet)
+		// Massage the packet. This should not normally fail. If it does, the database will not receive the packet.
```
iamnotacake, line 235: > Massage the packet

are you sure? :thinking:

decryptor/postgresql/pg_decryptor.go
```diff
@@ -232,84 +232,105 @@ func (proxy *PgProxy) ProxyClientConnection(errCh chan<- error) {
...
+			Errorln("Error occurred on query handler")
+	}
+	if changed {
+		packet.ReplaceQuery(newQuery.Query())
+	}
+	return false, nil
+}
+
+func (proxy *PgProxy) sendClientAcraCensorError(logger *log.Entry) error {
+	errorMessage, err := NewPgError("AcraCensor blocked this query")
```
iamnotacake, line 324: Sure, this error won't happen too often. But consider making it constant to avoid generation of the same packet every time it happens. Though I'm not sure whether Go allows calling such complex functions when creating constants (as opposed to C++ `constexpr` of Rust `const fn`).

decryptor/postgresql/pg_decryptor.go
```diff
@@ -557,65 +557,66 @@ func (proxy *PgProxy) ProxyDatabaseConnection(errCh chan<- error) {
 		}
 		proxy.clientConnection.SetWriteDeadline(time.Now().Add(network.DefaultNetworkTimeout))
 
-		// Let the protocol observer take a look at the packet, keeping note of it.
-		action, err := proxy.protocolState.HandleDatabasePacket(packetHandler)
+		// Massage the packet. This should not normally fail. If it does, the client will not receive the packet.
```
iamnotacake, line 565: One more massage :massage:

decryptor/postgresql/pg_decryptor.go
ilammy: > Though I'm not sure whether Go allows calling such complex functions when creating constants

Don't worry, it does not allow that :) Usually the errors are defined as global *variables*, precisely for that reason. The error objects have identities, so they can't be expressed in terms of simple constants (which work like substitution in Go), and the compiler won't do smart things like initializing them once and replacing uses with references.

ilammy: Woops. It seems that we both got bamboozled by this "error" word.

For some reason I thought that `NewPgError` is just a fancy constructor for Go's `error` when in fact it creates a PostgreSQL packet with a given error. In this case I don't think it's common to extract those things into reusable constants as they are not really useful anywhere else outside of the package. Plus, `NewPgError` can fail and there's a question of how this failure [during program startup] should be handled. It's not nice to just ignore it because we know it won't fail.

So... I've reverted the change (which I quickly applied without even trying to compile), and I think we can tolerate this 'inefficiency' since this condition should not happen often.

Lagovas (approved):

PR #425, opened 2020-10-07
Test handling of integer data fields

iamnotacake:
Add tests where we have integer column along with some encrypted columns.
Make sure this number stays "as is" and at the same time does not affect other columns.
* Test using acrawriter (create `AcraStruct` to `INSERT INTO` database)
* Test using `Zone` in transparent mode


tests/test.py
```diff
@@ -4785,6 +4785,128 @@ def testEmptyValues(self):
         self.assertEqual(row['binary'], b'')
 
 
+class TestEncryptionWithIntFields(BaseTestCase):
```
Lagovas, line 4788: hm, if we just check that field left as is, we can just add it to `TestTransparentEncryption`/`TestTransparentEncryptionWithZone` tests and extend table. There we check that all is correct when result contains null values, empty strings and we can add integer too as special case

@ilammy what you think about it? do we need separate test case which do the same and take a time for execution. or we can all such cases merge into one testcase?

ilammy: > do we need separate test case which do the same and take a time for execution. or we can all such cases merge into one testcase?

Mmm... Part of me thinks that it's a good thing to have separate tests where each of them tests their own thing, as that makes it easier to pinpoint the cause if one of them breaks. However, considering that we already have a test for transparent encryption which basically tests all item types, I guess it also makes sense to just extend it instead of adding a new one.

So I'm for the merge. @iamnotacake, please do :pray:

tests/test.py
```diff
@@ -4785,6 +4785,128 @@ def testEmptyValues(self):
...
+            self.temp_table.insert(),
+            {'id': value_id, 'data': data_encrypted, 'number': number, 'binary': binary})
+
+        # check values (select all)
+        result = self.engine1.execute(sa.select([self.temp_table]).where(self.temp_table.c.id == value_id))
+        row = result.fetchone()
+        self.assertEqual(row['id'], value_id)
+        self.assertEqual(row['data'], data)
+        self.assertEqual(row['number'], number)
+        self.assertEquals(row['binary'], binary)
```
ilammy, line 4818: This `Equal` vs `Equals` irks me a bit... Is there any difference? :thinking:

tests/test.py
```diff
@@ -4785,6 +4785,128 @@ def testEmptyValues(self):
...
+
+        # check values (select numbers only)
+        result = self.engine1.execute(
+            sa
+                .select([self.temp_table.columns.id, self.temp_table.columns.number])
+                .where(self.temp_table.c.id == value_id)
```
ilammy, line 4824: Paint me surprised, I didn't know that Python allowed to format method chaining like this. I wonder what PEP 8 has to say about it, though. (I'm not opposed to it. It's actually pretty readable:+1:)

ilammy (commented):
Good stuff! :+1:

I get a feeling that you've got a feeling of what the integration testing system of Acra is.

tests/test.py
iamnotacake: One of them seems to be deprecated, will fix that

tests/test.py
```diff
@@ -4553,6 +4554,7 @@ def get_context_data(self):
         context = {
             'id': get_random_id(),
             'default_client_id': get_pregenerated_random_data().encode('ascii'),
+            'number': 123456,
```
ilammy, line 4557: I'd suggest to use `get_random_id` or something to get a random integer here, just to avoid breaking the pattern of using random inputs, not fixed one. However, I don't insist.

ilammy (approved):

tests/test.py
iamnotacake: Done. Had to add another line to preserve the value between calls to `get_context_data()`

ilammy (approved):

PR #426, opened 2020-10-08
Observe extended query protocol

ilammy:
Add (or update) parsers for new extended protocol packets that we're interested in:

  - **Parse** and **ParseComplete** which prepare statements for execution
  - **Bind** and **BindComplete** which carry parameters for prepared statements
  - **Execute** which initiate prepared statement execution

This gives us all the context information we will need for later processing of extended queries. Use it to maintain prepared statement and cursor registries.

For now, we only observe the protocol flow between the client and the database and don't do much extra about it, only logging when we catch this or that packet. The current integration test suite already has some tests which use the extended protocol: `TestPgPlaceholders`, `CensorBlacklistTest`, `CensorWhitelistTest`, `TestPostgresqlBinaryPreparedStatement`. They all use unnamed cursors and sometimes named prepared statements. I've been looking at these tests to ensure that the parsers don't crash Acra. The coverage is not perfect but that's at least something. In the future I'd like to extend the `BasePrepareStatementMixin` test to include more diverse prepared statements.

Prerequisites:
- [X] #424 needs to be merged
- [X] Zeroize packet content after we're done with the query

Expected review:
- [x] @Lagovas
- [x] @iamnotacake

### DIFF
--- a/decryptor/postgresql/packet_handler.go
+++ b/decryptor/postgresql/packet_handler.go
@@ -288,6 +288,11 @@ func (packet *PacketHandler) IsBindComplete() bool {
 	return packet.messageType[0] == BindCompleteMessageType
 }
 
+// IsExecute return true if packet has Execute type
+func (packet *PacketHandler) IsExecute() bool {
+	return packet.messageType[0] == ExecuteMessageType
+}
+
 // GetParseData returns parsed Parse packet data.
 // Use this only if IsParse() is true.
 func (packet *PacketHandler) GetParseData() (*ParsePacket, error) {
@@ -310,6 +315,18 @@ func (packet *PacketHandler) GetBindData() (*BindPacket, error) {
 		return nil, err
 	}
 	return bind, nil
+}
+
+// GetExecuteData returns parsed Execute packet data.
+// Use this only if IsExecute() is true.
+func (packet *PacketHandler) GetExecuteData() (*ExecutePacket, error) {
+	packet.logger.Debugln("GetExecuteData")
+	execute, err := NewExecutePacket(packet.descriptionBufferCopy())
+	if err != nil {
+		packet.logger.Debugln("Failed to parse Bind packet")
+		return nil, err
+	}
+	return execute, nil
 }
 
 // ReplaceQuery query in packet with new query and update packet length
--- a/decryptor/postgresql/packet_handler.go
+++ b/decryptor/postgresql/packet_handler.go
@@ -272,6 +272,16 @@ func (packet *PacketHandler) IsParseComplete() bool {
 	return packet.messageType[0] == ParseCompleteMessageType
 }
 
+// IsBind return true if packet has Bind type
+func (packet *PacketHandler) IsBind() bool {
+	return packet.messageType[0] == BindMessageType
+}
+
+// IsBindComplete return true if packet has BindComplete type
+func (packet *PacketHandler) IsBindComplete() bool {
+	return packet.messageType[0] == BindCompleteMessageType
+}
+
 // GetParseData returns parsed Parse packet data.
 // Use this only if IsParse() is true.
 func (packet *PacketHandler) GetParseData() (*ParsePacket, error) {
@@ -282,6 +292,18 @@ func (packet *PacketHandler) GetParseData() (*ParsePacket, error) {
 		return nil, err
 	}
 	return parse, nil
+}
+
+// GetBindData returns parsed Bind packet data.
+// Use this only if IsBind() is true.
+func (packet *PacketHandler) GetBindData() (*BindPacket, error) {
+	packet.logger.Debugln("GetBindData")
+	bind, err := NewBindPacket(packet.descriptionBuf.Bytes())
+	if err != nil {
+		packet.logger.Debugln("Failed to parse Bind packet")
+		return nil, err
+	}
+	return bind, nil
 }
 
 // ReplaceQuery query in packet with new query and update packet length
--- a/decryptor/postgresql/packet_handler.go
+++ b/decryptor/postgresql/packet_handler.go
@@ -267,6 +267,11 @@ func (packet *PacketHandler) IsParse() bool {
 	return packet.messageType[0] == ParseMessageType
 }
 
+// IsParseComplete return true if packet has ParseComplete type
+func (packet *PacketHandler) IsParseComplete() bool {
+	return packet.messageType[0] == ParseCompleteMessageType
+}
+
 // GetParseData returns parsed Parse packet data.
 // Use this only if IsParse() is true.
 func (packet *PacketHandler) GetParseData() (*ParsePacket, error) {
--- a/decryptor/postgresql/packet_handler.go
+++ b/decryptor/postgresql/packet_handler.go
@@ -242,6 +242,12 @@ func (packet *PacketHandler) Reset() {
 	packet.messageType[0] = 0
 }
 
+func (packet *PacketHandler) descriptionBufferCopy() []byte {
+	buffer := make([]byte, packet.descriptionBuf.Len())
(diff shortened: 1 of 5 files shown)


decryptor/postgresql/packet_handler.go
```diff
@@ -262,15 +273,60 @@ func (packet *PacketHandler) IsParse() bool {
...
+	return packet.messageType[0] == BindCompleteMessageType
+}
+
+// IsExecute return true if packet has Execute type
+func (packet *PacketHandler) IsExecute() bool {
+	return packet.messageType[0] == ExecuteMessageType
+}
+
+// GetParseData returns parsed Parse packet data.
+// Use this only if IsParse() is true.
```
Lagovas, line 297: imho better to not make it as contract for callee and check one more time here with `IsParse`

decryptor/postgresql/pg_decryptor.go
```diff
@@ -214,89 +234,134 @@ func (proxy *PgProxy) ProxyClientConnection(errCh chan<- error) {
...
+				preparedStatement := proxy.protocolState.PendingParse()
+				log = log.WithField("prepared_name", preparedStatement.Name())
+			}
+			log.Debugln("New query")
 		}
 	}
+
+	// Let AcraCensor take a look at the query text.
+	// If it's not okay (and we're still alive), don't let the database see the query.
+	if censorErr := proxy.censor.HandleQuery(query.Query()); censorErr != nil {
```
Lagovas, line 320: what about to make censor as handler for `OnQuery` subscriber for unifying? It will allow us to list all handlers which we should turn on/off for acra. And provide way to change order of handlers, their amount, etc. Or even add them as plugins by config file. It will be easier if they will be common.

decryptor/postgresql/packet_handler.go
ilammy: > imho better to not make it as contract for callee and check one more time here with IsParse

Well, it's not like a strict contract... If this is no Parse packet, `GetParseData` will return `nil` and an error. It can be used to check for packet type but I feel that having separate `IsParse` is cleaner, and it's a cheap check. That way you can tell a difference between a malformed Parse packet and not a Parse packet at all.

decryptor/postgresql/pg_decryptor.go
ilammy: > what about to make censor as handler for OnQuery subscriber for unifying?

Hm... Sounds reasonable, actually, but we'd need to find a way to signal this special condition of AcraCensor emitting an error, but handling this error in a special way: aborting the client request and continuing service, as opposed to closing the session on other errors. I'm not sure what would be the best approach here. Downcasting to AcraCensor interface type to check for AcraCensor error specifically?.. Anyway, I'll keep a note of this idea.

> It will allow us to list all handlers which we should turn on/off for acra.

But isn't that already done via the configuration file, albeit implicitly? Or do you mean some global blade switch to enable/disable, say, AcraCensor, or tokenization, or anything else, *regardless* of the encryptor configuration file?

Lagovas: > But isn't that already done via the configuration file, albeit implicitly? Or do you mean some global blade switch to enable/disable, say, AcraCensor, or tokenization, or anything else, regardless of the encryptor configuration file?

Our encryption/decryption logic works as chain of handlers which pre-defined at start and their order have some sense. For example we can detokenize at first and then decrypt. Or decrypt first and then detokenize. Or we can mask and tokenize and vice versa. And we can setup it from configuration file and allow users to configure it. And we can add censor here too. And turn-off some of these handlers. It higher level than encryption config. Because encryptor's config tells what and how to encrypt. But we can provide the order of global handlers at start. Remove them, repeat, etc.

ilammy: > For example we can detokenize at first and then decrypt. Or decrypt first and then detokenize. Or we can mask and tokenize and vice versa. And we can setup it from configuration file and allow users to configure it. And we can add censor here too.

Oh, right :+1: *That* is not possible with the current configuration file, it's all hardwired in the proxy setup code and the implementations of query observers.

ilammy:
I expect CI to be successful this time. PR is *really* ready for review now.

Lagovas (approved):
lgtm

decryptor/postgresql/utils.go
```diff
@@ -193,6 +193,30 @@ func NewBindPacket(data []byte) (*BindPacket, error) {
...
+	return p.portal
+}
+
+// NewExecutePacket parses Executre packet from data.
+func NewExecutePacket(data []byte) (*ExecutePacket, error) {
+	portal, data, err := readString(data)
+	if err != nil {
+		return nil, err
+	}
+	maxRows := binary.BigEndian.Uint32(data)
```
iamnotacake, line 234: Meybe we should check whether `len(data) == 4` and `return nil, ErrPacketTruncated` if it's not?
Otherwise we can get a panic, see [encoding.binary.BigEndian.Uint32() sources](https://golang.org/src/encoding/binary/binary.go#L111)

utils/utils.go
```diff
@@ -139,15 +139,20 @@ func FillSlice(value byte, data []byte) {
 	}
 }
 
+// ZeroizeBytes wipes a byte slice from memory, filling it with zeros.
+func ZeroizeBytes(data []byte) {
+	FillSlice(0, data)
```
iamnotacake, line 144: This should be faster, as mentioned [here](https://github.com/cossacklabs/acra/pull/409#issuecomment-691951455):
```suggestion
	for i := range data {
		data[i] = 0
	}
```
Go compiler may be too dumb to inline call to `FillSlice()` and use `runtime.memclrNoHeapPointers` on `data`

decryptor/postgresql/protocol.go
```diff
@@ -174,3 +169,25 @@ func (p *PgProtocolState) HandleDatabasePacket(packet *PacketHandler) error {
...
+		p.pendingBind.Zeroize()
+	}
+	p.pendingBind = nil
+
+	if p.pendingExecute != nil {
+		p.pendingExecute.Zeroize()
+	}
+	p.pendingExecute = nil
+
+	// OnQuery uses "string" values and those can't be safely zeroized :(
```
iamnotacake, line 191: Maybe add `TODO` here? Also, we don't have to zeroize everything with `\0` byte, it can actually be any ASCII char code. This way, "zeroized" strings will remain valid even if they contained multi-byte chars. Of course, if there is a safe way of inplace string update.

decryptor/postgresql/utils.go
ilammy: > Meybe we should check whether len(data) == 4 and return nil, ErrPacketTruncated if it's not?

Great point! Thanks :heart:

decryptor/postgresql/protocol.go
ilammy: I'm afraid it's technically impossible to perform safely. In Go strings are immutable (from user standpoint). While I guess it's possible to hack into [`reflect.StringHeader`](https://golang.org/pkg/reflect/#StringHeader), I don't this it's a portable, safe, and future-proof approach.

Since zeroization is a best-effort technique, I believe it is acceptable to omit it here and rely on GC to launder the memory eventually.

utils/utils.go
ilammy: Okay, you win. I feel that this is mostly pointless optimization, but for the sake of resolving this argument once and for all, I need to do something.

I'll replace new uses of `FillSlice` with inlined loop in this PR. Then, after this PR is merged, there will be a separate PR replacing all remaining uses of `FillSlice(0, ...)` with appropriate `Zeroize...()` call. And then I'll remove `FillSlice` entirely from the code base since that's the only use it has.

iamnotacake (approved):

PR #429, opened 2020-10-14
Transparent encryption of prepared statement parameters

ilammy:
This PR adds transparent encryption of prepared statement parameters for PostgreSQL binary protocol. That is, when a **libpq** driver uses prepared statements, AcraServer will be encrypting not only the values given directly in the prepared SQL query text, but also the parameters passed during prepared statement execution.

Many drivers use the prepared statement protocol—technically called *extended query protocol*—for all queries, not just prepared ones. The extended protocol provides some *extended* flexibility. In this case the **Bind** messages are empty and don't carry any parameters. However, if the parameters *are* used, typically to implement actual prepared statements, AcraServer did not process them. Now it does.

Note that this **does not** add support for *textual* prepared queries (i.e., `PREPARE` and `EXECUTE` in SQL). This will come later as prepared statements are more often used by database drivers implicitly rather than explicitly spelled in SQL. There's also "text" format of parameter values in PostgreSQL protocol, but I was not able to trigger it in tests. This will be implemented later too.

<details>

### Accessing Bind message data

Add `OnBind()` method to the `QueryObserver` interface used for transparent encryption processing. This will allow to implement transparent encryption for prepared statement parameters. The parameters refer to the query but the query cannot be changed, only the parameters may be.

Parameter values are encapsulated in the `BoundValue` interface, similar to `OnQueryObject` interface for the `OnQuery()` method of `QueryObserver`.

`BindPacket` gets `GetParameters()` and `SetParameters()` accessors operating on the parameter values. The values are stored in the packet in a particular manner which must be respected. See [PostgreSQL docs](https://www.postgresql.org/docs/current/protocol-message-formats.html) for more details on the layout.

### Handling Bind packets 

Now we're ready to handle the Bind packets. This is performed similar to Parse packets, but now we need to retrieve an extra context: the prepared statement for which the bind is performed. After we have it, we can inspect the prepared statement values.

Note that in case of client error *we* have to deal with an invalid Bind packets, such as referring to a nonexistent prepared statement. In such cases we simply let the packets pass to the database unchanged. Then the database is likely to return an error response and this won't go anywhere.

Also note that the parameter count after processing might be different from the original input provided by the client. This is why OnBind() has the interface is has: instead of in-place modifications of values, it returns a new list of possibly modified values. Right now we won't be using this capability, but in the future this should allow us to handle queries like this:

    INSERT INTO table(col1, col2) VALUES ($1, $1);

by encrypting `col1` and `col2` values separately, into two different values. For that we may need to rewrite the query into

    INSERT INTO table(col1, col2) VALUES ($1, $2);

and when the client sends a Bind with one parameter, rewrite that too into a Bind with two parameters.

### Test transparent encryption of bound values

Add a new test to verify transparent encryption of bound values in prepared statements. Inherit from existing transparent encryption test to keep the coverage. However, now make sure that prepared statements are used to insert and update values in the database instead of using SQLAlchemy for that.

Unfortunately, I could not find a way to get SQLAlchemy to compile queries into proper prepared statement syntax, therefore the tests 'compile' the necessary SQL manually.

### Transparent encryption of INSERT and UPDATE parameters

Implement transparent encryption of INSERT parameters. For example, if the user prepares the following query:

    INSERT INTO table(col1, col2, col3) VALUES ($1, $2, 'user');

then we should encrypt the values of `$1` and `$2` parameters. However, we should not and cannot encrypt the 'user' string when processing bound parameters. That value should have been encrypted when the statement had been prepared.

In order to perform transparent encryption we need to map the parameter placeholders (`$1`, `$2`...) onto the table columns. That way we can consult the database schema from AcraServer configuration file and see whether a column needs to be encrypted and in what way.

We gather this information by analyzing the query. Note that not all queries are supported for transparent encryption. Only those that are simple enough for Acra to analyze will be encrypted. Other queries will be passed to the database unchanged. Or rather, the parameters of such queries will not be encrypted.

Just like with INSERT, AcraServer supports only a simple enough subset of queries which are direct updates of the table:

    UPDATE table SET col1 = $1, col2 = $2, col3 = 'fixed value';

We don't support more complex ones. This might come later, or might not be possible at all (e.g., queries updating table from another table).

</details>

Remaining tasks:

- [x] I'd like to reduce code duplication in INSERT and UPDATE handling a bit. It's too much copypasta there.

Expected review:

- [x] @Lagovas
- [x] @iamnotacake

### DIFF
--- a/decryptor/base/observers.go
+++ b/decryptor/base/observers.go
@@ -18,6 +18,7 @@ package base
 
 import (
 	"context"
+
 	"github.com/cossacklabs/acra/logging"
 	"github.com/cossacklabs/acra/sqlparser"
 	"github.com/sirupsen/logrus"
@@ -61,11 +62,50 @@ func NewOnQueryObjectFromQuery(query string) OnQueryObject {
 	return &onQueryObject{query: query}
 }
 
-// QueryObserver will be used to notify about coming new query
+// BoundValue is a value provided for prepared statement execution.
+// Its exact type and meaning depends on the corresponding query.
+type BoundValue interface {
+	Data() []byte
+	Encoding() BoundValueEncoding
+}
+
+// BoundValueEncoding specifies how to interpret the bound data.
+type BoundValueEncoding int
+
+// Supported values of BoundValueEncoding.
+const (
+	BindText BoundValueEncoding = iota
+	BindBinary
+)
+
+type boundValue struct {
+	data     []byte
+	encoding BoundValueEncoding
+}
+
+// Data of the bound value.
+func (v *boundValue) Data() []byte {
+	return v.data
+}
+
+// Encoding of the bound value data.
+func (v *boundValue) Encoding() BoundValueEncoding {
+	return v.encoding
+}
+
+// NewBoundValue makes a standard BoundValue from value data.
+func NewBoundValue(data []byte, encoding BoundValueEncoding) BoundValue {
+	return &boundValue{data, encoding}
+}
+
+// QueryObserver observes database queries and is able to modify them.
+// Methods should return "true" as their second bool result if the data has been modified.
 type QueryObserver interface {
 	ID() string
-	// OnQuery return true if output query was changed otherwise false
+	// Simple queries and prepared statements during preparation stage. SQL is modifiable.
 	OnQuery(data OnQueryObject) (OnQueryObject, bool, error)
+	// Prepared statement parameters during execution stage. Parameter values are modifiable.
+	OnBind(statement sqlparser.Statement, values []BoundValue) ([]BoundValue, bool, error)
 }
 
 // QueryObservable used to handle subscribers for new incoming queries
@@ -122,4 +162,21 @@ func (manager *ArrayQueryObserverableManager) OnQuery(query OnQueryObject) (OnQu
 		}
 	}
 	return currentQuery, changedQuery, nil
+}
+
+// OnBind would be called for each added observer to manager.
+func (manager *ArrayQueryObserverableManager) OnBind(statement sqlparser.Statement, values []BoundValue) ([]BoundValue, bool, error) {
+	currentValues := values
+	changedValues := false
+	for _, observer := range manager.subscribers {
+		newValues, changedNow, err := observer.OnBind(statement, currentValues)
+		if err != nil {
+			return values, false, err
+		}
+		if changedNow {
+			currentValues = newValues
+			changedValues = true
+		}
+	}
+	return currentValues, changedValues, nil
 }
--- a/decryptor/base/observers.go
+++ b/decryptor/base/observers.go
@@ -66,21 +66,21 @@ func NewOnQueryObjectFromQuery(query string) OnQueryObject {
 // Its exact type and meaning depends on the corresponding query.
 type BoundValue interface {
 	Data() []byte
-	Encoding() BoundValueEncoding
+	Format() BoundValueFormat
 }
 
-// BoundValueEncoding specifies how to interpret the bound data.
-type BoundValueEncoding int
+// BoundValueFormat specifies how to interpret the bound data.
+type BoundValueFormat int
 
-// Supported values of BoundValueEncoding.
+// Supported values of BoundValueFormat.
 const (
-	BindText BoundValueEncoding = iota
-	BindBinary
+	TextFormat BoundValueFormat = iota
+	BinaryFormat
 )
 
 type boundValue struct {
-	data     []byte
-	encoding BoundValueEncoding
+	data   []byte
+	format BoundValueFormat
 }
 
 // Data of the bound value.
@@ -88,14 +88,14 @@ func (v *boundValue) Data() []byte {
 	return v.data
 }
 
-// Encoding of the bound value data.
-func (v *boundValue) Encoding() BoundValueEncoding {
-	return v.encoding
+// Format of the bound value data.
+func (v *boundValue) Format() BoundValueFormat {
+	return v.format
 }
 
 // NewBoundValue makes a standard BoundValue from value data.
-func NewBoundValue(data []byte, encoding BoundValueEncoding) BoundValue {
-	return &boundValue{data, encoding}
+func NewBoundValue(data []byte, format BoundValueFormat) BoundValue {
+	return &boundValue{data, format}
 }
 
 // QueryObserver observes database queries and is able to modify them.
--- a/decryptor/postgresql/utils.go
+++ b/decryptor/postgresql/utils.go
@@ -229,11 +229,6 @@ func (p *BindPacket) parameterEncodingByIndex(i int) (base.BoundValueEncoding, e
 
 // SetParameters updates statement parameters from Bind packet.
 func (p *BindPacket) SetParameters(values []base.BoundValue) {
-	p.updateParameterFormats(values)
-	p.updateParameterValues(values)
-}
-
-func (p *BindPacket) updateParameterFormats(values []base.BoundValue) {
 	// See "Bind" description in https://www.postgresql.org/docs/current/protocol-message-formats.html
 	// If there are no parameters then don't bother.
 	if len(values) == 0 {
@@ -273,10 +268,10 @@ func (p *BindPacket) updateParameterFormats(values []base.BoundValue) {
 			}
 		}
 	}
-}
-
-func (p *BindPacket) updateParameterValues(values []base.BoundValue) {
-	p.paramValues = make([][]byte, len(values))
+	// Finally, replace parameter values. Reuse the top-level array if we can.
+	if len(values) != len(p.paramValues) {
+		p.paramValues = make([][]byte, len(values))
+	}
 	for i := range p.paramValues {
 		p.paramValues[i] = values[i].Data()
 	}
--- a/decryptor/postgresql/utils.go
+++ b/decryptor/postgresql/utils.go
@@ -166,8 +166,8 @@ type BindPacket struct {
 // ErrUnknownFormat is returned when Bind packet contains a value format that we don't recognize.
 var ErrUnknownFormat = errors.New("unknown Bind packet format")
 
-// ErrNotEnougFormats is returned when Bind packet is malformed and does not contain enough formats for values.
-var ErrNotEnougFormats = errors.New("format index out of range")
+// ErrNotEnoughFormats is returned when Bind packet is malformed and does not contain enough formats for values.
+var ErrNotEnoughFormats = errors.New("format index out of range")
 
 const (
 	bindFormatText   = 0
@@ -220,7 +220,7 @@ func (p *BindPacket) parameterEncodingByIndex(i int) (base.BoundValueEncoding, e
 		format = p.paramFormats[i]
 	} else {
 		log.WithField("index", i).WithField("max", len(p.paramFormats)).Debug("Bind format array too short")
-		return base.BindText, ErrNotEnougFormats
+		return base.BindText, ErrNotEnoughFormats
 	}
 	// Options currently include text and binary formats.
 	switch format {
--- a/decryptor/postgresql/utils.go
+++ b/decryptor/postgresql/utils.go
@@ -21,7 +21,9 @@ import (
 	"encoding/binary"
 	"errors"
 
+	"github.com/cossacklabs/acra/decryptor/base"
 	"github.com/cossacklabs/acra/utils"
+	log "github.com/sirupsen/logrus"
 )
 
 var terminator = []byte{0}
@@ -157,6 +159,14 @@ type BindPacket struct {
 	resultFormats []uint16
 }
 
+// ErrUnknownFormat is returned when Bind packet contains a value format that we don't recognize.
+var ErrUnknownFormat = errors.New("unknown Bind packet format")
+
+const (
+	bindFormatText   = 0
+	bindFormatBinary = 1
+)
+
 // PortalName returns the name of the portal created by this request.
 // An empty name means unnamed portal.
 func (p *BindPacket) PortalName() string {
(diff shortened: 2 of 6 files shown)


decryptor/postgresql/utils.go
```diff
@@ -176,6 +189,142 @@ func (p *BindPacket) Zeroize() {
 	}
 }
 
+// GetParameters extracts statement parameters from Bind packet.
+func (p *BindPacket) GetParameters() ([]base.BoundValue, error) {
+	values := make([]base.BoundValue, len(p.paramValues))
+	for i := range values {
+		encoding, err := p.parameterEncoding(i)
```
Lagovas, line 196: can't get used to golang's forloops (if one argument and iterate array then it is index, two args is index and value. but with maps one arg is key)... 
I thought that `i` is element of empty array, not index and try to understand why we pass empty element to `parameterEncoding`. Then I tried to understand why we pass just index to method named like field and expect result *(

can we rename `parameterEncoding` to `getParameterEncodingByIndex` or smth like that?

decryptor/postgresql/utils.go
```diff
@@ -176,6 +189,142 @@ func (p *BindPacket) Zeroize() {
...
+	// If there are no formats then all values use the default: text.
+	if len(p.paramFormats) == 0 {
+		return base.BindText, nil
+	}
+	// If there is only one format then it is used for all values.
+	var format uint16
+	if len(p.paramFormats) == 1 {
+		format = p.paramFormats[0]
+	} else {
+		format = p.paramFormats[i]
```
Lagovas, line 216: is it okay that we don't check that `i` not bigger than `len(p.paramFormats)`?

decryptor/postgresql/utils.go
```diff
@@ -176,6 +189,142 @@ func (p *BindPacket) Zeroize() {
...
+		return base.BindBinary, nil
+	default:
+		log.WithField("index", i).WithField("format", format).Debug("Unknown Bind format")
+		return base.BindText, ErrUnknownFormat
+	}
+}
+
+// SetParameters updates statement parameters from Bind packet.
+func (p *BindPacket) SetParameters(values []base.BoundValue) {
+	p.updateParameterFormats(values)
```
Lagovas, line 232: can we add one more function like `updateParameters(values)` which will do both functions? as I understand, formats and values must be synchronized with each other and it should be done together, yes?

decryptor/postgresql/utils.go
```diff
@@ -238,6 +387,9 @@ func NewExecutePacket(data []byte) (*ExecutePacket, error) {
 	return &ExecutePacket{portal, maxRows}, nil
 }
 
+const maxUint16 = 1<<16 - 1
```
Lagovas, line 390: why not to use `math.MaxUint16`?

encryptor/queryDataEncryptor.go
```diff
@@ -313,6 +316,187 @@ func (encryptor *QueryDataEncryptor) OnQuery(query base.OnQueryObject) (base.OnQ
...
+		columns = make([]string, len(insert.Columns))
+		for i, column := range insert.Columns {
+			columns[i] = column.String()
+		}
+	} else if cols := schema.Columns(); len(cols) > 0 {
+		columns = cols
+	}
+
+	changed := false
+	newValues := make([]base.BoundValue, len(values))
```
Lagovas, line 360: can we move it inside `if` statement to not allocate and copy values when we don't need it? plus we can move `if changed` check with `return` too.

encryptor/queryDataEncryptor.go
```diff
@@ -313,6 +316,187 @@ func (encryptor *QueryDataEncryptor) OnQuery(query base.OnQueryObject) (base.OnQ
...
+		//
+		// That is, where placeholders uniquely identify the column and used directly
+		// as inserted values of a single row. We don't support functions, casts,
+		// inserting multiple values and query results, etc.
+		//
+		// Walk through the query to find out which placeholders stand for which columns.
+		placeholders := make(map[int]string)
+		switch rows := insert.Rows.(type) {
+		case sqlparser.Values:
+			if len(rows) == 1 {
```
Lagovas, line 377: why we can't handle case with several values? queries like `insert into table(...) values (), (), ();`?

encryptor/queryDataEncryptor.go
```diff
@@ -313,6 +316,187 @@ func (encryptor *QueryDataEncryptor) OnQuery(query base.OnQueryObject) (base.OnQ
...
+			switch values[i].Encoding() {
+			case base.BindBinary:
+				changed = true
+				encryptedData, err := encryptor.encryptWithColumnSettings(settings, values[i].Data())
+				if err == ErrUpdateLeaveDataUnchanged {
+					return values, false, nil
+				}
+				newValues[i] = base.NewBoundValue(encryptedData, base.BindBinary)
+			default:
+				// Not supported at the moment.
```
Lagovas, line 411: for a future, as I understand, we can call `sqlparser.Parse(values[i].Data())` and expect something like `SQLVal` element from parser. Then `switch val.Type` and work with values. If it's hex value or acceptable string, we can encrypt them or tokenize. If it's integer, we can parse it with `strconv.Atoi`, generate new and encode with `strconv.Itoa` to not break format.

decryptor/postgresql/utils.go
ilammy: Mmm... Good point. I was too focused on the special cases: zero and one format.

Invalid format indices should not happen with correct packets, but I guess it makes sense to add a check and return an error in this case.

decryptor/postgresql/utils.go
ilammy: > can we add one more function like updateParameters(values) which will do both functions?

Uh... `SetParameters` is this function.

> formats and values must be synchronized with each other and it should be done together

Yeah, that's true.

I've split it into two functions early on in attempt to improve readability. However, this does not seem to improve it that much, so I guess it's better to just put everything into `SetParameters`.

decryptor/postgresql/utils.go
ilammy: > why not to use `math.MaxUint16`?

Thanks. I missed these constants. Will change.

encryptor/queryDataEncryptor.go
ilammy: > why we can't handle case with several values?

Uh... Now that you ask this... There does not seem to be a reason why not. The only thing that we care about is the placeholders being uniquely mapped onto columns. Which the code does not seems to check, btw. I'll fix this, and will add support for multiple value clauses.

encryptor/queryDataEncryptor.go
ilammy: > can we move it inside if statement to not allocate and copy values when we don't need it?

I'll see what I can do. At worst, we can check the `changed` value and allocate on the first change.

encryptor/queryDataEncryptor.go
ilammy: Mmm... Yeah, that was kinda the plan :+1:

I suspect that "text" will contain literal SQL which can be parsed. It *might* require additional type information though. On the positive side, the encryptor also has a suitable decoder around which might help with byte strings.

However, I did not find a way to trigger this format from tests so I'm not sure what exactly comes here.

decryptor/postgresql/utils.go
```diff
@@ -165,6 +165,9 @@ type BindPacket struct {
 // ErrUnknownFormat is returned when Bind packet contains a value format that we don't recognize.
 var ErrUnknownFormat = errors.New("unknown Bind packet format")
 
+// ErrNotEnougFormats is returned when Bind packet is malformed and does not contain enough formats for values.
+var ErrNotEnougFormats = errors.New("format index out of range")
```
Lagovas, line 170: ErrNotEnoug**h**Formats

Lagovas (approved):

encryptor/queryDataEncryptor.go
ilammy: I've cleaned up the code, extracted it into a utility method shared by INSERT and UPDATE implementations. Now it won't allocate a new slice if none of the values need to be encrypted. I've also fixed an issue in the original code while at it.

encryptor/queryDataEncryptor.go
ilammy: I've added a TODO into the code to handle this case later. I've also added a warning so that we see it in logs if the text format is being used and we don't handle it.

Lagovas (approved):
lgtm

encryptor/queryDataEncryptor.go
ilammy: I've extracted placeholder processing into a utility method shared between INSERT and UPDATE statements. I've also added additional safeguards there which should allow multiple VALUES in INSERT queries. Provided that those clauses are consistent. That is, we won't be able to handle
```
INSERT ... VALUES ($1, $2),  ($2, $1)
```
where we can't figure out what columns the values are for. However, something like
```
INSERT ... VALUES ($1, $2),  ($3, $4),  ($5, $6)
```
or
```
INSERT ... VALUES ($1, $2, 'foo'),  ($1, $2, 'bar')
```
should work.

encryptor/queryDataEncryptor.go
```diff
@@ -316,6 +316,12 @@ func (encryptor *QueryDataEncryptor) OnQuery(query base.OnQueryObject) (base.OnQ
 	return query, false, nil
 }
 
+// ErrInvalidPlacholder is returned when Acra cannot parse SQL placeholder expression.
+var ErrInvalidPlacholder = errors.New("invalid placeholder value")
```
Lagovas, line 320: ErrInvalidPlac**e**holder

encryptor/queryDataEncryptor.go
```diff
@@ -316,6 +316,12 @@ func (encryptor *QueryDataEncryptor) OnQuery(query base.OnQueryObject) (base.OnQ
...
+
+// ErrInconsistentPlacholder is returned when a placeholder refers to multiple different columns.
+var ErrInconsistentPlacholder = errors.New("inconsistent placeholder usage")
```
Lagovas, line 323: same

encryptor/queryDataEncryptor.go
```diff
@@ -451,6 +447,43 @@ func (encryptor *QueryDataEncryptor) encryptUpdateValues(update *sqlparser.Updat
...
+		// PostgreSQL placeholders look like "$1". Parse the number out of them.
+		text := string(placeholder.Val)
+		index, err := strconv.Atoi(strings.TrimPrefix(text, "$"))
+		if err != nil {
+			logrus.WithField("placeholder", text).WithError(err).Warning("Cannot parse placeholder")
+			return err
+		}
+		// Placeholders use 1-based indexing and "values" (Go slice) are 0-based.
+		index--
+		if index >= len(values) {
```
Lagovas, line 468: hm, it is a questionable check. We should be sure that postgresql doesn't allow something like `insert into table1 values ($1+$2, $3, $4)`, where $1 and $2 may be integers (which should be summed) or strings (which should be concatenated). But such cases we can't support, but it is another error. We should warn that it is not an invalid query but unsupported feature.

ilammy: It does not check the columns, it checks the values. Effectively, that
```
prepare something as insert into table1 values ($1+$2, $3, $4)
```
is followed by
```
execute something(1, 2, 3, 4)
```
where there are enough values for all placeholders. If there are two values, for example, we just bail out with an error and skip this **Bind** packet. The database is likely to reply with an error too.

I'm not sure this is a right decision – to let the database see column values which should have been encrypted – but I don't feel that we should police the queries too hard and deny their processing if Acra cannot parse them.

In regards to that, I'd like to reserve a proper action for a later PR. It should improve error handling in such cases, and make query analysis more explicit. Then it will be possible to configure Acra behavior for queries which are too hard to analyze: either we block the query and report an error, or let the query through unchanged, or change whatever we can change and leave out the rest, or something else.

encryptor/queryDataEncryptor.go
ilammy: I guess I'm in a hurry and have an affinity for typos today :joy:

decryptor/base/observers.go
```diff
@@ -61,11 +62,50 @@ func NewOnQueryObjectFromQuery(query string) OnQueryObject {
...
+
+// Supported values of BoundValueEncoding.
+const (
+	BindText BoundValueEncoding = iota
+	BindBinary
+)
+
+type boundValue struct {
+	data     []byte
+	encoding BoundValueEncoding
```
iamnotacake, line 83: PostgreSQL calls these fields `format`, and not `encoding`. See [this link](https://www.postgresql.org/docs/8.2/protocol-message-formats.html). Consider renaming this field to `format`, as well as `BoundValueEncoding` -> `BoundValueFormat` and `Encoding()` -> `Format()` in `type BoundValue interface`

ilammy: > Consider renaming [encoding] field to `format`

Make sense. Renamed.

ilammy:
I've found that PostgreSQL executors are useful in Acra EE as well while I've been working on porting all these changes there. Thus, I have extracted a common PostgreSQL fixture into yet another mix-in `BaseBinaryPostgreSQLMixin` which sets up two asyncpg executors and limits test execution to PostgreSQL.

I have also found an... well, a *working* approach to compile SQLAlchemy query object into something suitable for asyncpg. The `BaseBinaryPostgreSQLMixin` also includes a `compileQuery` method which converts SQLAlchemy query object and parameter dictionary (as used everywhere in tests) into a query string and parameter list that can be passed to asyncpg.

This should come handy when writing new tests for PostgreSQL prepared statements. Just inherit from `BaseBinaryPostgreSQLMixin` and use `self.executor1` or `self.executor2` to run your SQLAlchemy queries.

@Lagovas, @iamnotacake, please give a look to recently pushed changes, starting with https://github.com/cossacklabs/acra/pull/429/commits/51432a1a5930aa7f60aaaac4cc340aae6eee54fd.

tests/test.py
```diff
@@ -1340,6 +1340,31 @@ def testReadAcrastructInAcrastruct(self):
         self.assertEqual(row['empty'], b'')
 
 
+class BaseBinaryPostgreSQLMixin(BaseTestCase):
```
Lagovas, line 1343: if it is mixin then it should not be inherited from BaseTestCase to allow child classes inherit only methods of this mixin. It is allow to correctly inherit several classes with separate logic. If it should inherit from some other base class then it is "SpecificBaseClass". But it's is okay if one mixin inherited from another mixin which extends his API.
Good example of this way structuring we can see in [django views](https://github.com/django/django/blob/master/django/views/generic/detail.py), where exists one base class and a lot of mixins, which you can inherit and create your own special base class with blackjack and other things ))

ilammy: :thinking: I've inherited it from `BaseTestCase` so that calling `super().setUp()` and `super().checkSkip()` were predictable.

Would it be more accurate to call this class `BaseBinaryPostgreSQLTestCase` or something?

Lagovas: I think yes, will be okay. User who will inherit from this class will undertstand that it some base class and probably as other in a project do something in `setUp` and override it

Lagovas (approved):

iamnotacake (approved):

PR #430, opened 2020-10-19
Basic support of text format for parameters

ilammy:
It turns out that clients don't really use text formats for data types which are interesting to Acra CE's transparent encryption. It works only with byte arrays which are typically using binary format.

Text format may be occasionally used for types interesting to Acra EE: currently integers and floats. However, Acra EE's encryptors *expect* such data in text format in the first place.

Therefore, for text format of parameters we need to pass them as is. It's the binary format which requires more attention. If the parameter is integer – it still must be encoded as text for the encryptor, even if the parameter is binary.  We'll need to know what type the parameter has and then possibly decode and reencode it after encryption. Right now that's not possible. We continue passing binary data as is.

Passing through text parameters lets Acra EE's encryptors work correctly when text formats are used for interesting data types. Let's add this part at first and then get back to binary parameter recoding.

#### Text format support for "asyncpg" executor

Teach `AsyncpgExecutor` to use text format for some types of prepared statement parameters (integers and floats). This can be controlled with a new optional field `format` in the `ConnectionArgs`. Set it to `"text"` to request a text format. The default is to use binary format.

`BaseBinaryPostgreSQLTestCase` supports a class variable `FORMAT` which is the format passed to `AsyncpgExecutors` it creates. By default it uses binary format, but you can set it to `"text"` per test case.

Acra CE currently does not use this functionality, but Acra EE uses the base class from Acra CE and needs this.

**asyncpg** does not seem to support text format natively for anything other than strings. The current implementation is arguably a hack, but well, it works for our needs.

### Final notes

Required reviews:

- [ ] @Lagovas
- [ ] @iamnotacake (not blocker)

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -611,9 +611,11 @@ class ProcessStub(object):
         pass
 
 
-ConnectionArgs = collections.namedtuple(
-    "ConnectionArgs", ["user", "password", "host", "port", "dbname",
-                       "ssl_ca", "ssl_key", "ssl_cert"])
+ConnectionArgs = collections.namedtuple("ConnectionArgs",
+    field_names=["user", "password", "host", "port", "dbname",
+                 "ssl_ca", "ssl_key", "ssl_cert", "format"],
+    # 'format' is optional, other fields are required.
+    defaults=[None])
 
 
 class QueryExecutor(object):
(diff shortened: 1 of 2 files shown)


tests/test.py
```diff
@@ -719,11 +721,33 @@ def _connect(self, loop):
...
+                    decoder=float,
+                    format='text')
+            )
+
     def execute_prepared_statement(self, query, args=None):
         if not args:
             args = []
         loop = asyncio.get_event_loop()
         conn = self._connect(loop)
+        if self.connection_args.format == 'text':
```
Lagovas, line 749: can we use some constants instead hardcoded values? it force to assign them somewhere together and we may see which formats are used/supported. otherwise we should search by `.format*` where and which formats used in the code.
And as I understand, text format added here for acra-ee and for now not used in acra-ce, correct?

ilammy: > can we use some constants instead hardcoded values?

Yeah, that's a good idea.

What would be the best way to introduce such constants? I though about adding them to `AsyncpgExecutor` class like this:
```python
class AsyncpgExecutor(QueryExecutor):

    TextFormat = 'text'
    BinaryFormat = 'binary'
```
so that we can do
```python
    def execute_prepared_statement(self, query, args=None):
        conn = self._connect(loop)
        if self.connection_args.format == TextFormat:
            self._set_text_format(conn)
```
and then something like this:
```python
class BaseBinaryPostgreSQLTestCase(BaseTestCase):

    FORMAT = AsyncpgExecutor.BinaryFormat
```

Is this an acceptable approach?

ilammy: >  And as I understand, text format added here for acra-ee and for now not used in acra-ce, correct?

Yes. Acra CE currently doesn't use text format. Well, we can check it in transparent encryption tests, for example. I've been using them for debugging. Obviously, since Acra CE does not modify integers, the tests should behave exactly the same with either format. Should we test this format switch in Acra CE?

Lagovas: > Is this an acceptable approach?

Yes, due to missing enum support in python, common way to define class level variables as constants

> Should we test this format switch in Acra CE?

IMHO, will be good. probably it will save us from some mistakes when we will try to touch text format in acra-ce too.

ilammy: I've added both constants and a new test with text format.

Lagovas (approved):

PR #431, opened 2020-10-29
TLS validation on new connections

iamnotacake:
Here's a (very much) draft PR that shows current progress on implementing T1830.
It will be rebased and commits will be more clean before I make it a non-draft one. A lil bit messy right now.
Now I'm more interested in rough general review about the changes in common, don't waste too much time.
Maybe I'm doing something completely wrong. You may wanna look at `checkOCSP()` function, it contains logic that decides whether we should allow a certificate.

Implemented:
* Certificate generation script being able to put OCSP and CRL URLs into certificates
* Some certificates are updated, DB-related ones stay as is to avoid having to update their docker containers right now
* Parsing of few OCSP- and CRL-related command line options
* Structs for storing configuration related to OCSP and CRL
* Callback for TLS wrapper that uses OCSP configuration and performs the check

Not yet implemented:
* Validation using CRL (including caching, CRL doesn't provide any benefit if we download it on every check)
* Anything related to OCSP Stapling

Known issues:
* Too many log messages, they are mostly for debug
* Errors that are not constants, but are created when needed
* Different strings, including error messages and help about CLI options, they should be more intuitive
* In checkOCSP(), avoid asking the same server twice


tests/test.py
```diff
@@ -1036,6 +1041,37 @@ def fork_connector(self, connector_port: int, acraserver_port: int,
...
+        args = {
+            'port': port,
+            'index': TEST_TLS_OCSP_INDEX,
+            'rsigner': TEST_TLS_OCSP_CERT,
+            'rkey': TEST_TLS_OCSP_KEY,
+            'CA': TEST_TLS_CA,
+        }
+
+        cli_args = sorted(['-{}={}'.format(k, v) for k, v in args.items()])
+        print('openssl ocsp args: {}'.format(' '.join(cli_args)))
```
Lagovas, line 1058: better to use `logging` module instead print. it will allow us to change logs level and filter some levels if need.

tests/test.py
```diff
@@ -1036,6 +1041,37 @@ def fork_connector(self, connector_port: int, acraserver_port: int,
...
+
+        process = self.fork(lambda: subprocess.Popen(['openssl', 'ocsp'] + cli_args))
+
+        # Openssl OCSP server exitx on first invalid request, let's skip connection check for now
```
Lagovas, line 1062: oh crap )) but I predict that when acra-server/connector will check ocsp URI's by test query to OCSP server, our tests will start blinking, because sometimes OCSP server may up longer and acra will check before it up.
So better to make some valid request here by some python library or create some template of request as byte array and hardcode it here, and make correct HTTP request with payload from template.

cmd/acra-connector/acra-connector.go
```diff
@@ -397,7 +397,18 @@ func main() {
 	if connectorMode == connector_mode.AcraServerMode {
 		if *useTLS {
 			log.Infof("Selecting transport: use TLS transport wrapper")
-			tlsConfig, err := network.NewTLSConfig(network.SNIOrHostname(*tlsAcraserverSNI, *acraServerHost), *tlsCA, *tlsKey, *tlsCert, tls.ClientAuthType(*tlsAuthType))
+
+			ocspConfig, err := network.NewOCSPConfig("", "yes", "prefer")
```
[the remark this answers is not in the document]
iamnotacake: Currently, there is a stub in connector to make TLS config creation pass, but w/o CLI args parsing as in acra-server.

That's because I've noticed that even with TLS enabled in tests, `get_connector_tls_params()` is not executed.
See test class `SSLPostgresqlMixin`, it force defines `with_tls()` to return `False`. I've only tested on job `postgresql-ssl` and it seems to be the only one that redefines `with_tls()`

cmd/acra-server/acra-server.go
```diff
@@ -123,6 +123,13 @@ func main() {
 	tlsDbSNIOld := flag.String("tls_db_sni", "", "Expected Server Name (SNI) from database (deprecated, use \"tls_database_sni\" instead)")
 	tlsDbCert := flag.String("tls_database_cert", "", "Path to client TLS certificate shown to database during TLS handshake (overrides \"tls_cert\")")
 	tlsDbKey := flag.String("tls_database_key", "", "Path to private key of the TLS certificate used to connect to database (see \"tls_database_cert\")")
+	tlsOcspUrl := flag.String("tls_ocsp_url", "", "OCSP service URL")
```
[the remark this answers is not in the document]
iamnotacake: Oh, didn't know I can do it like that. Parsing args from different modules...

Anyway, here we need values right in `main()` function, to create two TLS configs, with slightly different OCSP config

cmd/acra-server/acra-server.go
```diff
@@ -242,7 +249,28 @@ func main() {
...
+			ocspClientConfig, err = network.NewOCSPConfig(*tlsOcspUrl, *tlsOcspRequired, *tlsOcspFromCert)
+		}
+		if err != nil {
+			log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorWrongConfiguration).
+				Errorf("Configuration error: invalid OCSP config")
+			os.Exit(1)
+		}
+		log.Infof("OCSP: client config: %s", ocspClientConfig.Describe())
+
+		crlConfig, err := network.NewCRLConfig(*tlsCrlUrl, *tlsCrlFromCert)
```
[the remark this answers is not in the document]
iamnotacake: `tlsCrlUrl == ""` -> `tlsCrlCromCert == "use"` (default value) -> use one from certificate (if any)

However, we may need to ignore CRL in such cases, if, for example, OCSP confirms the certificate

network/ocsp.go
```diff
@@ -0,0 +1,232 @@
...
+		return nil, err
+	}
+	ocspUrl, err := url.Parse(ocspServer)
+	if err != nil {
+		return nil, err
+	}
+	httpRequest.Header.Add("Content-Type", "application/ocsp-request")
+	httpRequest.Header.Add("Accept", "application/ocsp-response")
+	httpRequest.Header.Add("host", ocspUrl.Host)
+	httpClient := &http.Client{}
```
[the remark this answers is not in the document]
iamnotacake: omg, one more TLS config, to query OCSP over HTTPS, and it probably needs its own OCSP config

network/ocsp.go
```diff
@@ -0,0 +1,232 @@
...
+	ocspFromCertIgnore
+)
+
+type OCSPConfig struct {
+	url      string
+	required int // ocspRequired*
+	fromCert int // ocspFromCert*
+}
+
+func NewOCSPConfig(uri, required, fromCert string) (*OCSPConfig, error) {
```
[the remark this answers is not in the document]
iamnotacake: With current implementation — no, in future — probably, quite easy to extend

tests/test.py
iamnotacake: > sometimes OCSP server may up longer and acra will check before it up

It shouldn't be slow actually. But I agree, we may need more stable OCSP server instead of openssl.

Lagovas: we started call `check_connection` after each fork not just for fun) it was a frequent case on slow machines (circleci for example) when acra hadn't started yet when testcase tried to connect.

cmd/acra-server/acra-server.go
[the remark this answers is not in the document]
Lagovas: oh, I thought that empty string will return err on url.Parse call... but not. so it's okay
And yes, we should allow to ignore crl or skip any crl checks (for performance reasons, for example in local tests)

tests/test.py
```diff
@@ -1056,6 +1092,11 @@ def get_connector_api_connection_string(self, port=None):
             port = self.CONNECTOR_API_PORT_1
         return get_connector_connection_string(port)
 
+    def get_ocsp_server_connection_string(self, port=None):
+        if not port:
+            port = self.OCSPSERVER_PORT
+        return 'http://127.0.0.1:{}'.format(port)
```
ilammy, line 1098: I can imagine a (perverse) desire to run tests with a real OCSP server running elsewhere, so I guess it wouldn't hurt to make the hostname configurable too.

tests/test.py
```diff
@@ -1222,9 +1267,15 @@ def tearDown(self):
...
                      getattr(self, 'acra', ProcessStub())]
+
+        if self.with_tls():
+            processes.append(getattr(self, 'ocsp_server', ProcessStub()))
+
         stop_process(processes)
         send_signal_by_process_name('acra-server', signal.SIGKILL)
         send_signal_by_process_name('acra-connector', signal.SIGKILL)
+        if self.with_tls():
+            send_signal_by_process_name('openssl', signal.SIGKILL)
```
ilammy, line 1278: Hm... While it might be okay to blanket send SIGKILLs to `acra-*` things, doing so with all `openssl` instances is a bit risky.

@Lagovas, do you remember why these `send_signal_by_process_name()` calls are necessary? `stop_process()` already attempts to do graceful termination with SIGTERM and then issues a SIGKILL after a timeout, sending them out to specified PIDs, not just to all processes with a matching name.

network/ocsp.go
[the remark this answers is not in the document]
ilammy: Hm... Isn't this prone to recursion? IIRC, OCSP servers usually work over HTTP exactly for this reason: so that you don't have to special-case TLS connections to OCSP servers to skip OCSP checks for themselves.

network/tls_wrapper.go
```diff
@@ -108,14 +132,49 @@ func NewTLSConfig(serverName string, caPath, keyPath, crtPath string, authType t
 		}
 		certificates = append(certificates, cer)
 	}
+
+	verifyPeerCertificate := func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
```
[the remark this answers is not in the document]
ilammy: How about splitting certificate verification into its own interface:

```go
type CertificateVerifier interface {
        Verify(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error
}
```
Then we can provide a simple adapter from `CertificateVerifier` to whatever `tls.Config` wants while implementing certificate verification elsewhere. This enables building combinators for verification.

No-op verification, censoring by CRLs, checking via OCSPs:

```go
func NewNopVerifier() CertificateVerifier
func NewCRLVerifier(*CRLConfig) CertificateVerifier
func NewOCSPVerifier(*OCSPConfig) CertificateVerifier
```

Chained verification and verification groups:

```go
func AnySuccessful([]CertificateVerifier) CertificateVerifier
func AllSuccessful([]CertificateVerifier) CertificateVerifier
func ParallelAllSuccessful([]CertificateVerifier) CertificateVerifier
func ParallelAnySuccessful([]CertificateVerifier) CertificateVerifier
```

Individual leaf verifiers and their combinators are way easier to test separately than a monolith.

network/ocsp.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	return &OCSPConfig{url: uri, required: requiredVal, fromCert: fromCertVal}, nil
+}
+
+func (c *OCSPConfig) Describe() string {
+	return fmt.Sprintf("url=%s, required=%d, fromCert=%d", c.url, c.required, c.fromCert)
+}
+
+func queryOCSP(commonName string, clientCert, issuerCert *x509.Certificate, ocspServer string) (*ocsp.Response, error) {
```
ilammy, line 116: It would be nice to extract this into an interface:
```go
type OcspServer interface {
    Query(commonName string, clientCert, issuerCert *x509.Certificate) *ocsp.Response, error
}
```
This allows separating verification logic from query combination (instead of hardwiring it in `checkOCSP`). More importantly, this enables writing unit tests for OCSP query combianation *and* certificate verification as a whole, because you will be able to write stub OCSP servers that way.

ilammy:
Also, I have to note that the changes are well-commented. Thank you for being merciful to the people of the future reading this!

tests/test.py
Lagovas: actually some time we had issues with zombi processes after tests, when forked processes didn't stop by sigterm or some test cases incorrectly sent sigkill (forgot or exited before sigkill call)

network/crl.go
```diff
@@ -48,12 +48,27 @@ func NewCRLConfig(uri, fromCert string) (*CRLConfig, error) {
...
+}
+
+type DefaultCRLVerifier struct {
+	Config CRLConfig
+}
+
+func (v DefaultCRLVerifier) Verify(chain []*x509.Certificate) (int, error) {
+	if len(v.Config.uri) == 0 && v.Config.fromCert == crlFromCertIgnore {
+		return 0, nil
+	}
```
ilammy, line 132: This is somewhat important decision – skip CRL verification because the configuration does not contain one – I'd probably log it with *debug* error for each individual certificate. That makes it possible to see which verifications are performed as they go, if the administrator is interested.

In general, I'd dump the entire decision-making process for verification with debug/verbose logs. You never know when that comes handy, so it's better have it just in case.

network/ocsp.go
```diff
@@ -144,26 +153,35 @@ func queryOCSP(commonName string, clientCert, issuerCert *x509.Certificate, ocsp
...
 }
 
+type OCSPVerifier interface {
+	// Verify returns number of confirmations or error.
+	// The error is returned only if it is critical, for example:
+	// - the certificate was revoked
+	// - the certificate is not known by OCSP server and we requested tls_ocsp_required == "yes" or "all"
+	// - if we were unable to contact OCSP server(s) but we really need the response, tls_ocsp_required == "all"
+	Verify(chain []*x509.Certificate) (int, error)
+}
```
ilammy, line 163: This interface seems to be the same as `CRLVerifier`. Do you plan for them to be different eventually?

In my opinion, certificate verification is something generic: it can be CRL blacklists, interactive checks by OCSP, or even interactive query to the user if Acra were a GUI app. So it makes more sense, I think, to have a generic enough `Verifier` interface with assorted implementations for it. And documentations comments for particular `Verify()` implementation might describe their specifics.

network/crl.go
```diff
@@ -53,6 +58,59 @@ func NewCRLConfig(uri, fromCert string) (*CRLConfig, error) {
 	return &CRLConfig{uri: uri, fromCert: fromCertVal}, nil
 }
 
+type CRLClient interface {
+	// Fetch fetches and parses CRL from passed URI (can be either http:// or file://)
+	Fetch(uri string) (*pkix.CertificateList, error)
+}
```
ilammy, line 64: One can argue that it might be better to have the interface like this:

```go
type CRLClient interface {
        FetchCRL() (*pkix.CertificateList, error)
}
```
where the client is configured to always fetch from a particular location (and might be reconfigured later, if needed). This separates sources (local files vs. networks) which can be just different implementations, if necessary, without a giant switch in the current one.

Such interface also does not expose the (arguably) unnecessary coupling of CRLs to URIs. Right now the interface insists that a CRL *must* be identified with a URI. This might not be the case, for example, with test stubs, or maybe some fixed CRL hardwired into Acra.

network/crl.go
```diff
@@ -61,6 +119,11 @@ type CRLVerifier interface {
 
 type DefaultCRLVerifier struct {
 	Config CRLConfig
+	Client CRLClient
+	// key = URI of cached CRL
+	// value = parsed CRL
+	// XXX maybe hide it behind mutex?
+	cache map[string]*pkix.CertificateList
```
ilammy, line 126: IIRC, Acra *can* accept and make TLS connections from different goroutines, so it makes sense to synchronize mutable data. Go has `sync.Map` which might be useful instead of explicit locking, though.

@Lagovas, please correct me.

network/crl.go
Lagovas: It's okay if we rely only on static configured crl URIs. But we have one more source of URI: client's certificate. And when we accept new connection from client and he passes certificate with [crlDistributionPoint](https://tools.ietf.org/html/rfc5280#section-4.2.1.13), then we should use some client to fetch CRL data.
I discovered from rfc that crl URI may be not only HTTP URL, it may be ldap access point too, and probably other protocols for CRL fetching.

network/crl.go
Lagovas: Yes, at first we accept connection in main goroutine, then we pass it to new goroutine which may do tls handshake, plus will create new connection to database. And both of them will use verifiers. So yes, we should use thread-safe shared struct. And better to use sync.Map or mutex.
Plus as I told at standup, better to introduce interface for cache like 
```
type VerifierCache interface {
  Get(key []byte)([]byte, error)
  Put(key, data []byte) error
}
```
to easily change implementation in a future to some distributed storage.  Plus we should separate cache for each verifier and for each purpose: ocsp/crl. It will depends on implementation. By key prefix if it will store in redis, or separate tables/names in tables, or tree structure in boltdb with separate buckets for each purpose/direction of connections

network/ocsp.go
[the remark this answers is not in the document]
Lagovas: I missed, that response of ocsp should be signed. So we will validate it by key. Plus ocsp request contains only hashes without any information about certificates (but allow to use tables of hashes of public certificates for matching...). So actually yes, we can request it by http.

network/ocsp.go
Lagovas: Agree. Probably better to rely on `VerifyPeerCertificate` function signature: `func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error`, to provide access to all data from connection which may be used for all our verifiers, and to avoid any adapters for that.

network/tls_wrapper.go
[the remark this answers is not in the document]
Lagovas: Looks great

cmd/acra-server/acra-server.go
```diff
@@ -265,16 +267,21 @@ func main() {
 
 		ocspClientVerifier := network.DefaultOCSPVerifier{Config: *ocspClientConfig, Client: &network.DefaultOCSPClient{}}
 
-		crlConfig, err := network.NewCRLConfig(*tlsCrlUrl, *tlsCrlFromCert)
+		var crlClientConfig *network.CRLConfig
+		if len(*tlsCrlClientUrl) > 0 {
```
Lagovas, line 271: why not compare with empty string? `*tlsCrlClientUrl != ""`, because more often we check by len() values which are some containers and if not to look what type is of tlsCrlClientUrl variable, a reader may think that it's some array of strings.

cmd/acra-server/acra-server.go
```diff
@@ -265,16 +267,21 @@ func main() {
...
+			crlClientConfig, err = network.NewCRLConfig(*tlsCrlClientUrl, *tlsCrlFromCert)
+		} else {
+			crlClientConfig, err = network.NewCRLConfig(*tlsCrlUrl, *tlsCrlFromCert)
+		}
 		if err != nil {
 			log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorWrongConfiguration).
-				Errorln("Configuration error: invalid CRL config")
+				Errorf("Configuration error: invalid CRL config")
```
Lagovas, line 278: why `Errorf` instead `Errorln` if we don't pass any param? If sometimes we back to the standard `log` package, it will print line without newline character)

cmd/acra-server/acra-server.go
```diff
@@ -313,7 +320,21 @@ func main() {
 
 		ocspDbVerifier := network.DefaultOCSPVerifier{Config: *ocspDbConfig, Client: &network.DefaultOCSPClient{}}
 
-		dbTLSConfig, err = network.NewTLSConfig(network.SNIOrHostname(*tlsDbSNI, *dbHost), *tlsDbCA, *tlsDbKey, *tlsDbCert, tls.ClientAuthType(*tlsDbAuthType), ocspDbVerifier, crlVerifier)
+		var crlDbConfig *network.CRLConfig
+		if len(*tlsCrlDbUrl) > 0 {
```
Lagovas, line 324: same here. it's okay to check by `len` if we will split values to an array of strings by some separator.

cmd/acra-server/acra-server.go
```diff
@@ -313,7 +320,21 @@ func main() {
...
+		}
+		if err != nil {
+			log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorWrongConfiguration).
+				Errorf("Configuration error: invalid CRL config")
+			os.Exit(1)
+		}
+
+		crlDbVerifier := network.DefaultCRLVerifier{Config: *crlDbConfig, Client: network.DefaultCRLClient{}, Cache: &network.DefaultCRLCache{}}
+
+		dbTLSConfig, err = network.NewTLSConfig(network.SNIOrHostname(*tlsDbSNI, *dbHost), *tlsDbCA, *tlsDbKey, *tlsDbCert, tls.ClientAuthType(*tlsDbAuthType), ocspDbVerifier, crlDbVerifier)
```
Lagovas, line 337: soon we will need some config for TlsConfig creation )) because too much parameters)

network/crl.go
```diff
@@ -88,29 +88,58 @@ func (c DefaultCRLClient) Fetch(uri string) (*pkix.CertificateList, error) {
...
+}
+
+func (c *DefaultCRLCache) Get(key string) ([]byte, error) {
+	log.Debugf("CRL: cache: loading '%s'", key)
+	value, ok := c.cache.Load(key)
+	if ok {
+		value, ok := value.([]byte)
+		if !ok {
+			// should never happen since Put() only inserts []byte
+			return nil, errors.New("Unexpected value of invalid type")
```
Lagovas, line 121: can I ask you to add unit-test for this cache with check for this execution branch with this error?)

network/crl.go
```diff
@@ -120,19 +149,124 @@ type CRLVerifier interface {
...
+	confirmsByCertCRL := 0
+
+	// TODO avoid querying same CRL more than once, maybe create some list of checked CRLs (based on URI)
+
+	for {
+		if len(v.Config.uri) > 0 {
+			rawCRL, err := v.getCachedOrFetch(v.Config.uri)
+			if err != nil {
+				log.WithError(err).Debugf("CRL: Cannot get CRL from '%s'", v.Config.uri)
+				// XXX return error instead?
```
Lagovas, line 206: yes

network/crl.go
```diff
@@ -120,19 +149,124 @@ type CRLVerifier interface {
...
+				break
+			}
+
+			crl, err := x509.ParseCRL(rawCRL)
+			if err != nil {
+				log.WithError(err).Debugf("CRL: Cannot parse CRL from '%s'", v.Config.uri)
+				// XXX return error instead?
```
Lagovas, line 213: yes

network/crl.go
```diff
@@ -120,19 +149,124 @@ type CRLVerifier interface {
...
+				break
+			}
+
+			err = issuer.CheckCRLSignature(crl)
+			if err != nil {
+				log.WithError(err).Warnf("CRL: Failed to check signature for CRL at %s", v.Config.uri)
+				// XXX return error instead?
```
Lagovas, line 220: of course yes

network/crl.go
```diff
@@ -120,19 +149,124 @@ type CRLVerifier interface {
...
+				break
+			}
+
+			for i := range crl.TBSCertList.RevokedCertificates {
+				if crl.TBSCertList.RevokedCertificates[i].SerialNumber.Cmp(cert.SerialNumber) == 0 {
+					log.Warnf("CRL: Certificate %v was revoked at %v", cert.SerialNumber, crl.TBSCertList.RevokedCertificates[i].RevocationTime)
+					return 0, errors.New(fmt.Sprintf("Certificate %v was revoked at %v", cert.SerialNumber, crl.TBSCertList.RevokedCertificates[i].RevocationTime))
```
Lagovas, line 227: nice statement for one more unit-test

network/crl.go
```diff
@@ -120,19 +149,124 @@ type CRLVerifier interface {
...
+			if crl.TBSCertList.RevokedCertificates[i].SerialNumber.Cmp(cert.SerialNumber) == 0 {
+				log.Warnf("CRL: Certificate %v was revoked at %v", cert.SerialNumber, crl.TBSCertList.RevokedCertificates[i].RevocationTime)
+				return 0, errors.New(fmt.Sprintf("Certificate %v was revoked at %v", cert.SerialNumber, crl.TBSCertList.RevokedCertificates[i].RevocationTime))
+			}
+		}
+
+		confirmsByCertCRL += 1
+	}
+
+	// XXX with cache containing raw CRL, we have to parse and verify signature on every check,
```
Lagovas, line 268: nice idea, and better to use some LRU cache, to store only frequently requested CRL values. And we already have one [implementation](https://github.com/cossacklabs/acra/blob/master/keystore/lru/cache.go#L29) for that, used in keystore.
And I looked on lru implementation [doc](https://github.com/golang/groupcache/blob/master/lru/lru.go#L25) and see that it allow having unlimited size 
So maybe we can use our lru cache instead new one DefaultCRLCache? What do you think?

network/ocsp.go
```diff
@@ -208,6 +212,8 @@ func (v DefaultOCSPVerifier) Verify(chain []*x509.Certificate) (int, error) {
 	confirmsByConfigOCSP := 0
 	confirmsByCertOCSP := 0
 
+	// TODO avoid querying same OCSP more than once
```
Lagovas, line 215: yeah,  nice idea. probably to use some `map[string]struct{}{}` as set of unique values.

tests/test.py
```diff
@@ -1072,6 +1074,27 @@ def fork_ocsp_server(self, port: int, check_connection: bool=True):
...
 
+    def fork_crl_http_server(self, port: int, check_connection: bool=True):
+        logging.info("fork HTTP server with port {}".format(port))
+
+        http_server_connection = self.get_crl_http_server_connection_string(port)
+
+        cli_args = ['--bind', '127.0.0.1', '--directory', 'tests/ssl', str(port)]
+        print('python HTTP server args: {}'.format(' '.join(cli_args)))
+
+        process = self.fork(lambda: subprocess.Popen(['python3', '-m', 'http.server'] + cli_args))
```
Lagovas, line 1085: awesome python's standard library)

network/crl.go
iamnotacake: `CRLCache` will still have `[]byte` as value type, so maybe we need to change the interface? Or introduce one more interface to use between `CRLCache` and `CRLVerifier` and store some amount of parsed-verified CRLs

iamnotacake: Or... Did you mean add the LRU cache into `DefaultCRLVerifier` itself? Then I will only need to extend `DefaultCRLVerifier.getCachedOrFetch()` a little

Lagovas: > Or introduce one more interface to use between CRLCache and CRLVerifier and store some amount of parsed-verified CRLs

hm... actually it okay to provide one more interface for CRL/OCSP, which will wrap cache implementation and return typed values, to avoid extra parsing

> Did you mean add the LRU cache into DefaultCRLVerifier itself? Then I will only need to extend DefaultCRLVerifier.getCachedOrFetch() a little

not only add into, but better to hide DefaultCRLVerifier struct, probably make it private (to avoid misuse and allow initialize only by constructors) and add [factory method](https://github.com/cossacklabs/acra/blob/master/keystore/lru/cache.go#L45) which will accept size of cache as parameter and will initialize all that it need.

network/crl.go
```diff
@@ -120,19 +149,124 @@ type CRLVerifier interface {
...
+
+func (v DefaultCRLVerifier) getCachedOrFetch(uri string) ([]byte, error) {
+	crl, err := v.Cache.Get(v.Config.uri)
+	if crl != nil {
+		if err != nil {
+			return nil, err
+		} else {
+			return crl, nil
+		}
+	}
```
ilammy, line 163: This looks a bit too complex and does not handle the case where the cache itself fails. If `err != nil` then it is expected for `crl == nil`, and that case is not really handled.

I guess if we allow the cache to fail *and* want to handle this failure, `Get()` method should return `([]byte, bool, error)` which is handled like this:

```go
crl, ok, err := v.Cache.Get(v.Config.uri)
if err != nil {
	return nil, err
}
if ok {
	return crl, nil
}
// Else: no cache hit, fetch from network.
```
I don't really see why we'd care for the error here: if the cache fails, we go straight to querying the origin anyway. So I guess the `CRLCache` interface can be changed to this:

```go
type CRLCache interface {
	Get(key string) ([]byte, bool)     // true if cached, false if absent
	Put(key string, value []byte) bool // true if replaced, false if first insert
	Remove(key string) bool            // true if removed, false if never existed
}
```
I'd say that the cache is most likely going to be local, in-memory one anyway, so it does not have many failure modes that are important here.

Then the user code gets even more simple:
```go
crl, ok := v.Cache.Get(v.Config.uri)
if ok {
	return crl, nil
}
```

network/crl.go
```diff
@@ -120,19 +149,124 @@ type CRLVerifier interface {
...
+
+	crl, err = v.Client.Fetch(uri)
+	if crl != nil {
```
ilammy, line 166: Early exits are easier to read, IMO:
```go
crl, err = v.Client.Fetch(uri)
if err != nil {
	// place for your logging
	return nil, err
}

// Continue on the happy path...
```

cmd/acra-server/acra-server.go
ilammy: > soon we will need some config for TlsConfig creation

Well, I'd say it will be much more readable to have some `network.TLSConfig` with exported, named fields that you fill out, and then use
```go
func (c *TLSConfig) ToTLSConfig() (*tls.Config, error)
```
to convert that into Go structure while making all the necessary validation.

Lagovas: Using fields directly less flexible than to use methods, IMHO. For example configuring VerifyConnection handler. How we should configure this field with 2 handlers? Who is responsible to do this as chain of handlers? First added handler, second?
```
config.VerifyConnection = handler1
// some code
if config.VerifyConnection != nil {
   config.VerifyConnection = HandlerChain(config.VerifyConnection, handler2)
} else {
  config.VerifyConnection = handler2
}
```
yeah, it is an exaggerated example, but the main reason that better to encapsulate some logic with methods or in other way to allow change it in a future without rewriting client's code (in our case it's main.go functions in acra-ce and in acra-ee after) if we will change this logic

ilammy: Well... You can go
```go
config := &network.TLSConfig{
        // ...
        Verifier: VerifyAll(verifier1, verifier2),
}
```
I.e., build the verifier before even getting to constructing a TLS config. I doubt that it will be necessary to change the configuration midway, it seems mostly static thing. The struct just gives proper names to components instead of them being a long string of unnamed constructor parameters.

Of course, the builder pattern can be used, but I don't feel that it's that popular in Go, and in this particular case of setting up TLS verification, it seems more natural to keep verifier combinators separate from TLS config rather than having something like
```go
config.AddVerifier(verifier1)
config.AddVerifier(verifier2)
```
because you can never predict all the ways you'd want to combine the verifiers and implement the assortment of methods for setting them up in TLS config itself. It's more flexible to have the combinators freestanding and then just pass the resulting complete verifier to the TLS config.

Lagovas: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis - good article with pros/cons of several methods of configuration. Cons for using config as struct is empty values and default configuration.
Actually, I prefer to try options) But anyway we need to use the same wrappers as you show like `VerifyAll(...)`

.circleci/check_configs.sh
```diff
@@ -9,6 +9,8 @@ function compare_configs() {
      cmp_status="$?"
      if [[ "${cmp_status}" != "0" ]]; then
         status=1
+        # Show diff in logs, easier to spot the difference
+        diff -U3 ${folder_a}/acra-${cmd}.yaml ${folder_b}/acra-${cmd}.yaml
```
ilammy, line 13: ❤️

network/cert_verifier.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	"crypto/x509"
+	"errors"
+	"fmt"
+	log "github.com/sirupsen/logrus"
+)
+
+// CertVerifier is a generic certificate verifier
+type CertVerifier interface {
+	Verify(chain []*x509.Certificate) (int, error)
```
Lagovas, line 12: why so difficult and return count of verifies?) we can return only errors and make several wrappers for such algorithms which will be related on confirm count
```
type CertVierifier interface {
  VerifyPeerCertificate func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error
}

type OCSPVerifier struct{}
func (verifier *OCSPVerifier) VerifyPeerCertificate func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {return nil}

type CRLVerifier struct{}
func (verifier *CRLVerifier) VerifyPeerCertificate func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {return nil}

// we can use this implementation for at least and for verifyAll
type AtLeastXVerifier struct {
  verifiers []CertVierifier
  minCount int
}
func NewAtLeastXVerifier(x int, verifiers []Verifiers)(AtLeastXVerifier, error){return nil}
func (verifier *AtLeastXVerifier) VerifyPeerCertificate func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
  count := 0
  for _, v := range verifier.verifiers {
    if err := v.VerifyPeerCertificate(rawCerts, verifiedChains); err != nil {
      count++
    }
  }
  if count < verifier.minCount {
    return ErrNotEnoughVerificationCount
  }
}

func NewVerifyAllVerification(verifiers []CertVerifier)(*AtLeastXVerifier, error){
  // reuse AtLeastXVerifier as special case
  return NewAtLeastXVerifier(len(verifiers), verifiers)
}
```

After that we can configure to use only OCSPVerifier, only CRLVerifier, or combine both with special requirements like verify all or X verifiers. And move logic of counting to separate implementations which implement same interface

network/connection_wrapper_test.go
```diff
@@ -31,6 +31,25 @@ import (
 var TestClientID = []byte("client")
 var TestServerID = []byte("server")
 
+var TestCertVerifier CertVerifier
+
+func init() {
```
Lagovas, line 36: why not to do it with factory method which use where it need? to not clog global scope of variables.
```
func NewTestCertVerifier(t *testing.T) CertVerifier{
ocspConfig, err := NewOCSPConfig("", "yes", "prefer")
	if err != nil {
		t.Fatal("Cannot create OCSP config")
	}

	crlConfig, err := NewCRLConfig("", "use")
	if err != nil {
		t.Fatal("Cannot create CRL config")
	}

	TestCertVerifier, err = NewCertVerifierFromConfigs(ocspConfig, crlConfig)
	if err != nil {
		t.Fatal("Cannot create certificate verifier")
	}
        return TestCertVerifier
}
```

Plus you use panics and... I don't know how test runner will handle it. Is it will recover and continue execution or will stop on first panic...

network/cert_verifier.go
```diff
@@ -0,0 +1,95 @@
...
+		if err != nil {
+			return confirmsTotal, err
+		}
+		confirmsTotal += confirms
+		if confirms > 0 {
+			confirmsByVerifiers++
+		}
+		// CertVerifierAtLeast will stop checking once we got >0 confirms from
+		// at least v.minConfirmsRequired verifiers in v.verifiers
+		if confirmsByVerifiers >= v.minConfirmsRequired {
```
ilammy, line 108: What if we have N verifiers, require at least M (where M < N), but the last verifier rejects the certificate?

To extend this question, what is the use case for `CertVerifierAtLeast`? What's the merit for the user to use M > 1 when they can't quite choose which verifiers they'll end up skipping?

I'd ask @shadinua for comments on common verification pattern.

ilammy (commented):
Nice implementation with interfaces! 👍

cmd/acra-server/acra-server.go
```diff
@@ -253,35 +253,35 @@ func main() {
 		}
 
 		var ocspClientConfig *network.OCSPConfig
-		if len(*tlsOcspClientUrl) > 0 {
-			ocspClientConfig, err = network.NewOCSPConfig(*tlsOcspClientUrl, *tlsOcspRequired, *tlsOcspFromCert)
+		if *tlsOcspClientURL != "" {
```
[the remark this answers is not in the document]
ilammy: @Lagovas, I'm afraid I don't quite follow you. Could you please elaborate with a practical example?

Let's take OCSP URL for example. I can see three possible parameters here:

1. `tls_ocsp_url` — set URL for both client and DB
2. `tls_client_ocsp_url` — set URL only for certificates used by the client
3. `tls_db_ocsp_url` — set URL only for certificates used by the DB

which can be combined in the following ways, as I see it:

1. Provide only one: `tls_ocsp_url`.
    - Common configuration, only one OCSP authority.
2. Use just two: `tls_client_ocsp_url` and  `tls_db_ocsp_url`.
    - Distinct configurations possible, may use different OCSP servers.
    - If configurations are the same, specify both with the same value.
    - Easily extendable for more configurations if needed.
3. All three: `tls_ocsp_url` + `tls_client_ocsp_url` & `tls_db_ocsp_url`.
    - Distinct configurations possible, may use different OCSP servers if needed.
    - If the configuration is the same, specify just common `tls_ocsp_url`.
    - Easily extendable for more configurations if needed.

Which one do you suggest to implement? If there's a fourth option, please note it as well.

I personally prefer the option 3 as it can be convenient for easy configuration when there is only one authority, but can be easily used in advanced mode to override configuration for client/DB if necessary. However, it also introduces some complexity so the option 2 is quite viable as well (albeit not so convenient on the command-line, for example). Option 1 is inconsistent with the rest of the TLS configuration now, so I wouldn't consider it.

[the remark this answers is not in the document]
Lagovas: I thought that enough only 2 case with 2 separate values. it not hard to copy paste same value for both parameters. but... probably u r right and we can simplify configuration for simple configurations with one ocsp/crl parameter for both sides

network/cert_verifier.go
ilammy: Okay, I'd support not returning the successful verification count.

While it seems like a good thing to write into logs, I don't think Acra can make any decisions based on this number. When the verifier has been configured and is used, you don't even know how many internal verifiers there are inside. So you can't really tell what verification count is normal and what is not.

The only place where you can – sorta-kinda – look and act on the verification behavior is *the implementation* of the verifier: composite verifiers know how many direct children they have and which of them fail. However, they have no clue how many child verifiers their children have, just like the caller of the composite verifier does not know it.

So I concur with @Lagovas here and support returning an error alone. This makes verification a clear-cut process:

- Either the certificate is okay and we allow the connection.
- Or the certificate is now okay and we deny the connection.

There is no third option of "well, maybe" when some sub-verifiers don't return a conclusive result and the composite verifiers defers the decision to the caller by returning `nil` error but not-quite-successful verification count (which the caller cannot distinguish).

If only an error is returned, the verifier is responsible for making the decision whether it's okay or not to trust the certificate. The conditions of trust are configurable by the administrator.

network/connection_wrapper_test.go
ilammy: >  don't know how test runner will handle it. Is it will recover and continue execution or will stop on first panic

That'll kill the test runner even before it starts running tests.

Well, I'd avoid using global test fixtures which can cause weird *side-effects* later with concurrent test execution.

network/crl.go
```diff
@@ -0,0 +1,380 @@
...
+		if err != nil {
+			return nil, err
+		}
+		crlURL, err := url.Parse(uri)
+		if err != nil {
+			return nil, err
+		}
+		httpRequest.Header.Add("Accept", "application/pkix-crl, application/pem-certificate-chain")
+		httpRequest.Header.Add("host", crlURL.Host)
+		httpClient := &http.Client{}
```
Lagovas, line 105: will be better if we will initialize client object at DefaultCRLClient initialization and will reuse it.
Plus will be great to set up timeouts for HTTP client to avoid long requests and control it
https://golang.org/pkg/net/http/#Client - We can configure `client.Timeout` parameter by some default value (better to use constant)
Plus we can configure transport object which should be used with all timeouts, and tls.Config if we will need it in a future

network/crl.go
```diff
@@ -0,0 +1,380 @@
...
+	case crlFromCertIgnoreStr:
+		fromCertVal = crlFromCertIgnore
+	default:
+		return nil, errors.New("Invalid `tls_crl_from_cert` value '" + fromCert + "', should be one of 'use', 'ignore'")
+	}
+
+	if uri != "" {
+		// Since this is CRL configuration alone, we don't have access to cache yet;
+		// so let's just download the CRL and forget about it
+		_, err = DefaultCRLClient.Fetch(DefaultCRLClient{}, uri)
```
Lagovas, line 73: is it tested and works? as I see, CRLClient.Fetch accepts only one parameter, not 2

network/crl.go
```diff
@@ -0,0 +1,380 @@
...
+
+// CRLCache is used to store fetched CRLs to avoid downloading the same URI more than once
+type CRLCache interface {
+	Get(key string) ([]byte, error)
+	Put(key string, value []byte) error
+	Remove(key string) error
+}
+
+// DefaultCRLCache is a default implementation of CRLCache, internally stores items in sync.Map
+type DefaultCRLCache struct {
```
Lagovas, line 141: can we reuse cache from keystore to not introduce one more and maintain less?
https://github.com/cossacklabs/acra/blob/master/keystore/lru/cache.go - they have similar interfaces on 50% )
plus I don't see where we actually use this DefaultCRLCache

network/crl.go
```diff
@@ -0,0 +1,380 @@
...
+// ParsedCRLCache is similar to CRLCache, but works on top of it and stores parsed and verified CRLs
+type ParsedCRLCache interface {
+	Get(key string) (*pkix.CertificateList, error)
+	Put(key string, value *pkix.CertificateList) error
+	Remove(key string) error
+}
+
+// PRLRUParsedCRLCache is an implementation of ParsedCRLCache that uses LRU cache inside
+type LRUParsedCRLCache struct {
+	cache lru.Cache
```
Lagovas, line 186: why not to re-use - https://github.com/cossacklabs/acra/blob/master/keystore/lru/cache.go this our cache implementation, which has same struct?) and only make method wrappers which parse bytes to final output structs

network/crl.go
```diff
@@ -0,0 +1,380 @@
...
+	if crl != nil {
+		if err != nil {
+			// non-empty result + error, should never happen
+			return nil, err
+		}
+
+		return crl, nil
+	}
+
+	// If not found, search in v.Cache, if found then parse, verify and store in v.ParsedCache
```
Lagovas, line 254: why we use 2 cache instead 1? we fetch data, parse and process it, and put it into "ParsedCache". Why we need to put it into the unparsed cache? which benefits it provides?

network/crl.go
```diff
@@ -0,0 +1,380 @@
...
+func (v DefaultCRLVerifier) Verify(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+	for _, chain := range verifiedChains {
+		log.Debugf("CRL: Verifying '%s'", chain[0].Subject.String())
+
+		if len(v.Config.uri) == 0 && v.Config.fromCert == crlFromCertIgnore {
+			log.Debugln("CRL: Skipping check since no config URI specified and we were told to ignore URIs from certificate")
+			return nil
+		}
+
+		cert := chain[0]
```
Lagovas, line 324: what if there empty chain? can we test it in unittests?

network/crl.go
```diff
@@ -0,0 +1,380 @@
...
+		issuer := chain[1]
```
Lagovas, line 325: same, what if there only client's certificate, without any intermediate? it is a common case after misconfigurations.

network/crl.go
```diff
@@ -0,0 +1,380 @@
...
+			crl, err := v.getCachedOrFetch(v.Config.uri, issuer)
+			if err != nil {
+				log.WithError(err).Debugf("CRL: Cannot get CRL from '%s'", v.Config.uri)
+				return err
+			}
+
+			for _, revokedCertificate := range crl.TBSCertList.RevokedCertificates {
+				if revokedCertificate.SerialNumber.Cmp(cert.SerialNumber) == 0 {
+					log.Warnf("CRL: Certificate %v was revoked at %v", cert.SerialNumber, revokedCertificate.RevocationTime)
+					return fmt.Errorf("Certificate %v was revoked at %v", cert.SerialNumber, revokedCertificate.RevocationTime)
```
Lagovas, line 343: need unit test (and separate error variable for it)

network/crl_test.go
```diff
@@ -0,0 +1,392 @@
...
+		Addr:    addr,
+		Handler: mux,
+	}
+	go func() {
+		httpServer.ListenAndServe()
+	}()
+	defer httpServer.Close()
+
+	// Give the server time to start, otherwise this test will fail sometimes
+	time.Sleep(time.Millisecond * 50)
```
Lagovas, line 184: better to add one more function (like check_connection in our python tests) which will test HTTP server's startups in a loop with some retry count and sleeps, after which return control or call t.Fatal. 
We had situations when on local machine was enough 50ms to start some service but had blinking tests in CI environments where tests running on slower machines and in docker containers

network/crl_test.go
```diff
@@ -0,0 +1,392 @@
...
+	validVerifiedChains[0] = append(validVerifiedChains[0], cert)
+	cert, err = pemToX509Cert(validRawCerts[1])
+	if err != nil {
+		t.Fatalf("Failed to parse test PEM certificate: %v", err)
+	}
+	validVerifiedChains[0] = append(validVerifiedChains[0], cert)
+
+	t.Logf("validVerifiedChains[0][0].Subject = %s", validVerifiedChains[0][0].Subject.String())
+	t.Logf("validVerifiedChains[0][1].Subject = %s", validVerifiedChains[0][1].Subject.String())
+	if err = validVerifiedChains[0][0].CheckSignatureFrom(validVerifiedChains[0][1]); err != nil {
```
Lagovas, line 351: what we test here?) in this testcase we should test our DefaultCRLVerifier and be sure that it use valid test mixtures. If we need check test data, better to do it in separate function (or better check it once manually to be sure that it is valid and it's enough for that. it can't change their state after that from valid to invalid ^ )

network/crl_test.go
```diff
@@ -0,0 +1,392 @@
...
+	cert, err := pemToX509Cert(validRawCerts[0])
+	if err != nil {
+		t.Fatalf("Failed to parse test PEM certificate: %v", err)
+	}
+	validVerifiedChains[0] = append(validVerifiedChains[0], cert)
+	cert, err = pemToX509Cert(validRawCerts[1])
+	if err != nil {
+		t.Fatalf("Failed to parse test PEM certificate: %v", err)
+	}
+	validVerifiedChains[0] = append(validVerifiedChains[0], cert)
```
Lagovas, line 347: all of these preparations make it harder to understand what and how we test here our DefaultCRLVerifier. 
Better to move it to separate method like `getValidVerifiedCertificateChain(t *testing.T) []*x509.Certificate` and call it here by 1 line instead 10.

network/crl_test.go
```diff
@@ -0,0 +1,392 @@
...
+
+	err = crlVerifier.Verify(validRawCerts, validVerifiedChains)
+	if err != nil {
+		t.Fatalf("Unexpected error for valid certificate: %v", err)
+	}
+
+	//
+	// Test invalid certificate chain that contains revoked certificate
+	//
+	invalidRawCerts := [][]byte{}
```
Lagovas, line 363: same, it's okay to move it to separate method like `getInvalidVerifiedChain`. It will allow reusing of this method in some other test methods if it will need it. And it will simplify the reading testcase.

network/ocsp.go
```diff
@@ -0,0 +1,273 @@
...
+		_, err := url.Parse(uri)
+		if err != nil {
+			return nil, err
+		}
+
+		// TODO: Do some request to `uri`, log warn if failed
+	}
+
+	var requiredVal int
+	switch required {
```
Lagovas, line 89: better to refactor it as map[string]int as global variable and there just use like `ocspRequiredMap[required]`.
```
// --tls_ocsp_from_cert=<use|trust|prefer|ignore>
const (
	// Use OCSP servers listed in certificate (if any), try them after the one
	// configured from CLI/config
	ocspFromCertUseStr = "use"
	// Query servers listed in certificate and don't perform further requests
	// if one respons with "ok, valid"
	ocspFromCertTrustStr = "trust"
	// Query servers listed in certificate before the one from config
	ocspFromCertPreferStr = "prefer"
	// Ignore OCSP servers listed in certificates
	ocspFromCertIgnoreStr = "ignore"
)

const (
	ocspFromCertUse int = iota
	ocspFromCertTrust
	ocspFromCertPrefer
	ocspFromCertIgnore
)
var ocspFromCertMap = map[string]int {
  ocspFromCertUseStr: ocspFromCertUse,
  ....
}
```
At first, we will link string and int values together close to constant declaration. At second, it will allow to reuse this logic in tests, other parts of code and not require to copy-paste this switch.

network/ocsp.go
```diff
@@ -0,0 +1,273 @@
...
+	case ocspRequiredNoStr:
+		requiredVal = ocspRequiredNo
+	case ocspRequiredAllStr:
+		requiredVal = ocspRequiredAll
+	default:
+		return nil, errors.New("Invalid `ocsp_required` value '" + required + "', should be one of 'yes', 'no', 'all'")
+	}
+
+	var fromCertVal int
+	switch fromCert {
```
Lagovas, line 101: and same for other switches. here we should only check that it's valid value by `requiredVal, ok := requiredValueMap[requiredVal]` and check `ok` variable.

network/ocsp.go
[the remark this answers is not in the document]
Lagovas: here the same advice, to move client initialization to constructor with pre-defined timeouts/transport object.

network/ocsp.go
```diff
@@ -0,0 +1,273 @@
...
+	url      string
+	fromCert bool
+}
+
+// Verify ensures certificate is not revoked by querying configured OCSP servers
+func (v DefaultOCSPVerifier) Verify(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+	for _, chain := range verifiedChains {
+		log.Debugf("OCSP: Verifying '%s'", chain[0].Subject.String())
+
+		cert := chain[0]
```
Lagovas, line 199: we should validate all input from unknown sources. We don't sure that there are at least 2 values in array.

network/ocsp.go
```diff
@@ -0,0 +1,273 @@
...
+		issuer := chain[1]
+
+		for _, ocspServer := range cert.OCSPServer {
```
Lagovas, line 202: imho, it is a redundant code, because it not interesting when we ignore it and waste extra time. plus we log it in else branch below

network/ocsp.go
```diff
@@ -0,0 +1,273 @@
...
+
+			response, err := v.Client.Query(cert.Issuer.CommonName, cert, issuer, serverToCheck.url)
+			if err != nil {
+				log.WithError(err).Warnf("Cannot query OCSP server at %s", serverToCheck.url)
+
+				if v.Config.required == ocspRequiredAll {
+					return errors.New("Cannot query OCSP server, but --tls_ocsp_required=all was passed")
+				}
+
+				continue
```
Lagovas, line 245: we should use some counter or flag that at least one check was successful. or we can loop over all values, get error from all of them and not get any valid check.

network/tls_wrapper.go
```diff
@@ -108,14 +137,28 @@ func NewTLSConfig(serverName string, caPath, keyPath, crtPath string, authType t
...
+		err := certVerifier.Verify(rawCerts, verifiedChains)
+
+		if err != nil {
```
Lagovas, line 144: can we log it by one line without extra if statements like `log.WithErorr(err).WithField("status", err == nil).Debugln("verifyPeerCertificate")` ? and return error
Then we will have log message like:
```
time=2020-10-10T10:10:10 verifyPeerCertificate status=true error=nil

or

time=2020-10-10T10:10:10 verifyPeerCertificate status=false error=somethingWrong
```

tests/test.py
Lagovas: ping

tests/ssl/generate_tls_keys.sh
```diff
@@ -17,33 +19,78 @@ openssl req -x509 \
     -config "$(dirname $0)/openssl.cnf" -extensions "v3_ca" \
     -days ${EXPIRE}
 
-declare -a names=("mysql" "postgresql" "acra-writer" "acra-server")
+# certificate database will be regenerated
+rm -f ${OUT_DIR}/index.txt
+
+declare -a names=("mysql" "postgresql" "acra-writer" "acra-writer-revoked" "acra-server" "ocsp")
```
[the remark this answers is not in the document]
Lagovas: ping

tests/ssl/generate_tls_keys.sh
```diff
@@ -17,33 +19,78 @@ openssl req -x509 \
...
 for name in "${names[@]}"
 do
+    subj="${SUBJ}/CN=Test leaf certificate ($name)"
     mkdir -p "${OUT_DIR}/${name}"
     # create private key
+    [ "$SKIP_EXISTING" = 0 ] || [ ! -f "${OUT_DIR}/${name}/${name}.key" ] && \
+    echo "==> Generating private key for ${name}" && \
```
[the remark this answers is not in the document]
Lagovas: i'm agree

cmd/acra-server/acra-server.go
```diff
@@ -123,6 +123,15 @@ func main() {
 	tlsDbSNIOld := flag.String("tls_db_sni", "", "Expected Server Name (SNI) from database (deprecated, use \"tls_database_sni\" instead)")
 	tlsDbCert := flag.String("tls_database_cert", "", "Path to client TLS certificate shown to database during TLS handshake (overrides \"tls_cert\")")
 	tlsDbKey := flag.String("tls_database_key", "", "Path to private key of the TLS certificate used to connect to database (see \"tls_database_cert\")")
+	tlsOcspURL := flag.String("tls_ocsp_url", "", "OCSP service URL")
+	tlsOcspClientURL := flag.String("tls_ocsp_client_url", "", "OCSP service URL, for client certificates only")
+	tlsOcspDbURL := flag.String("tls_ocsp_database_url", "", "OCSP service URL, for database certificates only")
+	tlsOcspRequired := flag.String("tls_ocsp_required", "yes", "Whether we need OCSP response in order to accept certificate")
```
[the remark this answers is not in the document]
iamnotacake: Agree, but list of possible values is not enough. We should also show their description somewhere. But that seems like too many words to put in last argument of `flag.String()`

network/crl.go
iamnotacake: This is just a different way of writing
```go
crlClient := DefaultCRLClient{}
_, err = crlClient.Fetch(uri)
```

network/crl.go
iamnotacake: > can we reuse cache from keystore to not introduce one more and maintain less?

That one was designed to store `[]byte` while we want to store `pkix.CertificateList`. `Get()` tries to cast value to `[]byte` before returning it. Also, calling `NewCacheKeystoreWrapper()` in context of CRL sounds confusing. Much easier to copy the implementation and change it a little.

>  plus I don't see where we actually use this DefaultCRLCache

It is used to store fetched CRLs (as raw byte arrays), while `LRUParsedCRLCache` stores (a limited amount of) parsed and verified CRLs. You've mentioned once that we may want to use some kind of database instead of in-memory `DefaultCRLCache`.

network/crl.go
iamnotacake: Well, I guess it's not possible to have completely empty chain in TLS. There should be at least one certificate. Cannot prove it right now, gotta dig into docs. But you gave me an idea: the issuer (`chain[-1]`) may be empty in case of single-certificate chain.

network/crl_test.go
```diff
@@ -0,0 +1,350 @@
...
+	"io/ioutil"
+	"net/http"
+	"os"
+	"testing"
+	"time"
+)
+
+func getTestCACert() []byte {
+	// tests/ssl/ca/ca.crt
+	return []byte(`
```
ilammy, line 34: Those certificates might be a bit painful to update once they expire (or rather, when something inside them becomes deprecated).

I guess it would make more sense to just read them from the files directly. However, it's a bit tricky to do in all the contexts the tests are executed (IDEs, CI runners...) I remember @Lagovas mentioning once that Go is getting some magic macro that will expand into file's content at compile time. It would be nice to leave a TODO for future refactoring, I think.

network/crl_test.go
iamnotacake: Okay, will add some kind of loop. However, those problems you are talking about, they are more related to multi-container or at least multi-process setups while here we have everything in a single process.

network/crl_test.go
iamnotacake: It's more like a small sanity check to make sure the first one is issued by the second one

network/crl_test.go
```diff
@@ -0,0 +1,350 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return cert, nil
+}
+
+func TestDefaultCRLClientHTTP(t *testing.T) {
+	crl := getTestCRL()
+	addr := "127.0.0.1:45318"
```
ilammy, line 165: I guess it would be (technically) better to use `:0` here to let the OS choose a random unused port for the server.

First you manually create a listening socket:
```go
listener, err := net.Listen("tcp", ":0")
```
then you pass it to the server:
```go
go httpServer.Serve(listener)
```
and connect later to the listener's address:
```go
uri := fmt.Sprintf("http://%s/test_crl.pem", listener.Addr().String())
```

This approach also has a hidden benefit of creating the listening socket beforehand, as opposed to using `ListenAndServe` in a goroutine. That way the client can start connecting right away, as the listening socket is already there, so you don't need a `time.Sleep()` call to avoid errors.

network/crl_test.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+}
+
+func TestDefaultCRLClientFile(t *testing.T) {
+	crl := getTestCRL()
+
+	file, err := ioutil.TempFile("/tmp", "go_test_*.crl")
+	if err != nil {
+		t.Fatalf("Cannot create temporary file to store CRL: %v", err)
+	}
```
ilammy, line 221: I guess an empty string instead of `"/tmp"` will be slightly better here, letting Go choose the default temporary directory for the system.

ilammy (commented):
I like that we get more tests! ❤️

network/crl_test.go
iamnotacake: Wow, didn't know you can let OS decide the port to listen on

network/crl.go
ilammy: I agree that this makes things complicated. It's not really obvious which one of those you need:
```go
	Cache       CRLCache
	ParsedCache ParsedCRLCache
```
and what's the difference.

I believe it would be enough to cache only parsed results as that is our happy path. On a cache hit we get `*pkix.CertificateList` right away. On a cache miss we take the slow path: do the network request, parse the results, and cache them only after success.

cmd/acra-server/acra-server.go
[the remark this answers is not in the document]
Lagovas: actualy, it's okay to write something `"Whether we need OCSP response in order to accept certificate. Accept next values: %s | %s | %s"` and place constant values instead placeholders.

network/crl.go
Lagovas: wow, didn't see such kind of calls in golang)

tests/ssl/openssl.cnf
```diff
@@ -14,10 +14,30 @@ distinguished_name = req_dn
 basicConstraints = CA:false
 keyUsage = nonRepudiation, digitalSignature, keyEncipherment
 subjectAltName = DNS:localhost
+authorityInfoAccess = OCSP;URI:http://127.0.0.1:8888
+crlDistributionPoints = URI:http://127.0.0.1:8889/crl.pem
```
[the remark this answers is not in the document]
shad: I don't think I'm the "shad" you're looking for 😬 .  Perhaps @shadinua ?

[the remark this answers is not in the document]
ilammy: Oh, *that* why there is no response 😂

I'm terribly sorry for the mistake. But well... welcome to our pull request 😁

Lagovas (commented):
p.s. please, let us know when you finished the changes and it is ready for review) to not review every commit which may not have some changes but you already know about them and planned to do

storojs72:
Let me outline my thoughts on these validations.

Now (after our discussion with voice and brief reviewing commits), it seems that I understand the introduced caching. In case when we configured AcraServer to use `crl` as endpoint (or if application presents the TLS certificate with `crl` field specified as the endpoint) it allows us to avoid downloading CRL list all the time when connection is established. And I would agree with @shadinua that caching should have minimal live time (minutes) - that seems to be useful for cases when application creates a lot of (probably even simultaneous) connections to AcraServer and we don't need to load CRL from our CA and parse it all the time when new connection comes. We do not use caching for OCSP, since this protocol in intended to be "real-time" in a sense that when we ask CA that we trust - "is this ceritificate is valid (not revoked)" - the response should be new and "fresh" all the time. Please, correct me if I'm wrong in my understanding here.

The second point is configuring TLS validation. We should pay some attention on trust model. As we discussed, we intend to have three options for configuring this on AcraServer (`crl` file, `crl` endpoint, `ocsp` endpoint). That means that when client application connects - AcraServer verifies it by checking presented certificates to be issued by CA that is under our control (that AcraServer trusts) - AcraServer verifies signature of the certificate. Now additionally we will have the functionality to extend and enforce this verification by checking that application's certificate is not revoked using CRL/OCSP provider that we also trust  - sometimes CA and CRL/OCSP provider may be different entities. From [RFC5280](https://tools.ietf.org/html/rfc5280):
```
If the certificate issuer is not the CRL issuer, then the cRLIssuer field MUST be 
present and contain the Name of the CRL issuer.  If the certificate issuer is also 
the CRL issuer, then conforming CAs MUST omit the cRLIssuer field and MUST 
include the distributionPoint field
``` 

Alternatively, if client's certificate contains CRL/OCSP fields, and in case when AcraServer's CRL/OCSP endpoints are different, I think that AcraServer should at least warning about it (or probably even better - to make this configurable). That is because, AcraServer trusts the external CA that issued client's certificate (which is checked by signature verification) but may not trust to the CRL/OCSP providers that those external CA trusts. This is the rare case when maximum level of paranoia is activated (level 4 in Go I think - when we mandate validation of client application certificates)

And one more "performance" recommendation - when client application validates AcraServer's certificate's via OCSP (some application may also want to do it) - we can use OCSP stapling - a state-of-the-art technique that can be described as follows. When initial TLS handshake is establishing, AcraServer sends OCSP response together with its TLS certificate - which allows client application to skip the steps of requesting OCSP on its own. According to  [Cloudflare's blogpost](https://blog.cloudflare.com/ocsp-stapling-how-cloudflare-just-made-ssl-30/) this allows ~30% boost in establishing new connections. [This another blogpost](https://blog.cloudflare.com/high-reliability-ocsp-stapling/) describes the mechanics of stapling in details. We probably can implement it in separate engineering task (even as enterprise feature)

tests/ssl/openssl.cnf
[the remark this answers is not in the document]
shadinua: > Isn't localhost better? Like, for built-in IPv6 support.
Of course, in general case it is preferable to use domain names instead of IPs when it possible.

But. During a protracted IPv4 → IPv6 transition period that multiplied by a rather complex resolving mechanism in *nix systems, sometimes we can observe a lot of strange side effects, when `localhost` may not be functionally equal neither to `127.0.0.1` nor to `::1`.

So I wouldn't say that using `localhost` instead of `127.0.0.1` make things better in that particular case.

Generally speaking I'd suggest to use:
* `localhost`/`localhost6` — when we either do production-ready installations or we want to test not only the product, but the whole interaction with the base OS
* `127.0.0.1` — when we want to test exceptionally the product and be sure that the [mis]configuration resolving subsystem on base OS will not affect on the flow; today, by default, socket on `127.0.0.1` will be created even if we specify `::1` as IP to bind, so in observable future this approach should work; as a bonus — this approach is a bit faster that also may be useful in tests

network/ocsp.go
```diff
@@ -85,54 +85,51 @@ func NewOCSPConfig(uri, required, fromCert string) (*OCSPConfig, error) {
...
-		requiredVal = ocspRequiredNo
-	case ocspRequiredAllStr:
-		requiredVal = ocspRequiredAll
-	default:
-		return nil, errors.New("Invalid `ocsp_required` value '" + required + "', should be one of 'yes', 'no', 'all'")
+	requiredValValues := map[string]int{
+		ocspRequiredYesStr: ocspRequiredYes,
+		ocspRequiredNoStr:  ocspRequiredNo,
+		ocspRequiredAllStr: ocspRequiredNo,
 	}
```
ilammy: I wonder if there is any merit in having parallel lists of `ocspRequired${foo}Str` and `ocspRequired${foo}` instead of just using a single set of named string constants everywhere...

cmd/acra-connector/acra-connector.go
```diff
@@ -404,7 +404,7 @@ func main() {
 				log.WithError(err).Fatalln("Cannot create OCSP config")
 			}
 
-			crlConfig, err := network.NewCRLConfig("", "use")
+			crlConfig, err := network.NewCRLConfig("", "use", 16)
```
ilammy: I'd leave some TODO here to make it configurable later.

Also, I believe AcraTranslator needs one too.

ilammy (commented):
Cute tests :3

cmd/acra-connector/acra-connector.go
```diff
@@ -229,6 +229,14 @@ func main() {
...
+	tlsOcspURL := flag.String("tls_ocsp_url", "", "OCSP service URL")
+	tlsOcspRequired := flag.String("tls_ocsp_required", "denyUnknown", "Whether we need OCSP response in order to accept certificate")
+	tlsOcspFromCert := flag.String("tls_ocsp_from_cert", "prefer", "How should we treat OCSP server described in certificate itself")
+	tlsOcspCheckWholeChain := flag.Bool("tls_ocsp_check_whole_chain", false, "Whether to check whole certificate chain, false = only end certificate")
+	tlsCrlURL := flag.String("tls_crl_url", "", "CRL URL")
+	tlsCrlFromCert := flag.String("tls_crl_from_cert", "use", "How should we treat CRL URL described in certificate itself")
+	tlsCrlCacheSize := flag.Int("tls_crl_cache_size", 16, "Size of in-memory LRU cache for storing fetched CRLs, 0 = unlimited")
+	tlsCrlCacheTime := flag.Int("tls_crl_cache_time", 60, "How long cached CRL is considerend valid, seconds, max = 300")
```
ilammy, line 239: Making coordinated changes between multiple binaries may be bothersome, so it would be nice to pack all these command-line parameters into some shared place.

An approach that I have been using at times is to declare some
```go
type OCSPParameters struct {
        URL string
        // ...
}
```
which has a `Register()` method to register the command-line parameters, then have a global one somewhere.

This approach allows for command-line reuse *and* you can pass a single parameter struct to `network.NewOCSPConfig()` constructor instead of adding more and more parameters.

Or it might be possible to reuse existing `OCSPConfig` struct for that...

tests/ssl/Makefile
```diff
@@ -0,0 +1,233 @@
...
+# $1 => group-name
+# $2 => name
+# $3 => key size in bits (leave empty for default, 2048)
+define GenerateRSAKey
+$(1)/$(2)/$(2).key:
+	@mkdir -p $$(shell dirname "$$@")
+	@echo -e "\e[1;37m==> Generating private key $(1)/$(2)/$(2).key\e[0m"
+	[ -e "$(1)/$(2)/$(2).key" ] && echo "SKIP" || \
+	openssl genrsa -out "$(1)/$(2)/$(2).key" $(if $(3),$(3),2048) && \
+	chmod 400 "$(1)/$(2)/$(2).key"
```
ilammy, line 19: Does it _**need**_ to be a makefile? I mean, at this point, a Perl script might be more readable and use fewer `$`. How about just a Bash script to regenerate everything? It's not like we need to track dependencies too closely here.

ilammy (commented):
Alrighty... At this point I have personally stopped paying much attention to the actual changes. I believe, it's time to wrap it up because reviewing the end result is going to be hard, with 4k lines added and counting (most of it are certificates, but nevertheless).

Also, with 130+ comments here, I doubt that it will be easy to keep track of which have been resolved and which haven't.

I'd suggest @iamnotacake to perform a rebase and mop up, then see if there are some parts that can be submitted on their own, and push them out for review while working on the remaining things in a different branch.

network/ocsp.go
```diff
@@ -172,7 +173,7 @@ func (c DefaultOCSPClient) Query(commonName string, clientCert, issuerCert *x509
 	if err != nil {
 		return nil, err
 	}
-	ocspResponse, err := ocsp.ParseResponse(output, issuerCert)
+	ocspResponse, err := ocsp.ParseResponse(output, nil)
```
Lagovas, line 176: why nil?

network/ocsp_test.go
```diff
@@ -256,6 +259,80 @@ func TestDefaultOCSPClient(t *testing.T) {
...
 }
 
+func testWithConfigAndValidChain(t *testing.T, ocspConfig *OCSPConfig, rawCerts [][]byte, verifiedChains [][]*x509.Certificate) {
+	ocspClient := DefaultOCSPClient{}
+
+	ocspVerifier := DefaultOCSPVerifier{Config: *ocspConfig, Client: ocspClient}
+
+	err := ocspVerifier.Verify(rawCerts, verifiedChains)
+	if err != nil {
+		t.Fatalf("Unexpected error for valid certificate: %v", err)
```
Lagovas, line 269: please, add "\n" to the end of string because in the output it will be concatenated to the previous string

tests/test.py
```diff
@@ -922,6 +924,7 @@ class BaseTestCase(PrometheusMixin, unittest.TestCase):
     ACRASERVER_PORT = int(os.environ.get('TEST_ACRASERVER_PORT', 10003))
     OCSP_SERVER_PORT = int(os.environ.get('TEST_OCSP_SERVER_PORT', 8888))
     CRL_HTTP_SERVER_PORT = int(os.environ.get('TEST_HTTP_SERVER_PORT', 8889))
+    CRL_PATH = abs_path(os.environ.get('TEST_CRL_PATH', 'tests/ssl'))
```
Lagovas, line 927: not obvious why crl path is a folder, not a file

network/ocsp_test.go
```diff
@@ -256,6 +259,80 @@ func TestDefaultOCSPClient(t *testing.T) {
...
+	uri := fmt.Sprintf("http://%s", addr)
+
+	// TODO generate these chains by reading files, like getOCSPTestCase() does
+	validRawCerts, validVerifiedChains := getValidTestChain(t)
+	invalidRawCerts, invalidVerifiedChains := getInvalidTestChain(t)
+
+	//
+	// Test with default config, certificates contain OCSP server inside
+	//
+	ocspConfig, err := NewOCSPConfig("", ocspRequiredYesStr, ocspFromCertUseStr, false)
```
Lagovas, line 314: better to split it to 3 tests even if we will need to copy paste some code. it will simplify to refactor separate cases and it will not touch other cases. plus it will simplify understanding where error occurred, when output will be something like:
```
TestDefaultOCSPVerifierValidChain  Success
TestDefaultOCSPVerifierInvalidChain Failure
TestDefaultOCSPVerifierEmptyOCSPServer Failure
```
instead `TestDefaultOCSPVerifier Failure`

cmd/acra-connector/acra-connector.go
iamnotacake: Sounds like a good idea, but one issue is that there are two URIs: for client-side and for DB side. And when parsing CLI args we cannot know which one we need in this specific `OCSPConfig` because there'll be two of them, each for different `TLSConfig`.

network/crl.go
```diff
@@ -0,0 +1,59 @@
...
+	crlFromCertIgnore
+)
+
+type CRLConfig struct {
+	uri      string
+	fromCert int // crlFromCert*
+}
+
+func NewCRLConfig(uri, fromCert string) (*CRLConfig, error) {
+	_, err := url.Parse(uri)
```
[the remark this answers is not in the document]
iamnotacake: There was a problem that currently used OCSP server in integration tests will die when invalid HTTP request is performed. However, doing HEAD seems like a possible solution. Of not, maybe serializing some `ocsp.Request` (which cert should we query?) will do the job.

network/crl.go
iamnotacake: Didn't reuse one from `keystore`  since:
\- key/value are `string`/`[]byte` (like in `func (cache *Cache) Add(keyID string, keyValue []byte)`) while we need `pkix.CertificateList` (or even some custom struct, to contain CRL and fetch time)
\- that's a small piece of code, it cannot be properly reused and I don't wanna change it just to make it fit into both keystore needs and CRL cache needs, potentially breaking something else

network/crl.go
iamnotacake: You mean artificially put some other type there and then `Get()` it?

This branch should be unreachable because as comment says we only use `Put()` to store things and `Put()` is limited to this particular type (`*CRLCacheItem`)

ilammy: IMO, in case of weakly typed structures it's perfectly okay to write
```go
// Cast is always successful since Put() only inserts []byte
value, _ = value.([]byte)
```
with no failure path, as it is not possible.

network/ocsp.go
iamnotacake: Actually, this was a mistake. First, I thought there should be `issuer` certificate, as default OCSP responder certificate (if the responder didn't send own one attached to response). Then I misunderstood some things and put `nil` here.
What we have now: `ocsp.ParseResponse(output, issuerCert)`. So it [tries](https://github.com/golang/crypto/blob/c1f2f97bffc9/ocsp/ocsp.go#L550) to use attached cert if exists, otherwise tries passed one (`issuerCert`) who signed the cert we ask about.
Inside `ParseResponse()` there is a check that if `issuer != nil` then attached OCSP responder cert should be signed by `issuer`. This way, we only accept OCSP responses signed by either issuer itself or some intermediate cert signed by this exact issuer.

tests/ssl/generate_tls_keys.sh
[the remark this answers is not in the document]
iamnotacake: Renamed to `ocsp-responder` in newer [not yet pushed] version.
Also got that new "groups" of certificates (related files/dirs: `tests/ssl/Makefile`, `tests/ssl/group-{default,with-intermediate}`) where it is also called `ocsp-responder` and `intermediate-ocsp-responder` respectively. Though I'm not sure that kind of overcomplicated layout will be needed, if everything we need can be done with this script.

network/crl.go
```diff
@@ -0,0 +1,380 @@
...
+	default:
+		return nil, errors.New("Invalid `tls_crl_from_cert` value '" + fromCert + "', should be one of 'use', 'ignore'")
+	}
+
+	if uri != "" {
+		// Since this is CRL configuration alone, we don't have access to cache yet;
+		// so let's just download the CRL and forget about it
+		_, err = DefaultCRLClient.Fetch(DefaultCRLClient{}, uri)
+		if err != nil {
+			log.WithError(err).Warnf("CRL: Cannot fetch configured URI '%s'", uri)
```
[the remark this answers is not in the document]
iamnotacake: For this I've put a TODO:

	// TODO return error after issues with failing tests are fixed;
	//      CRL HTTP server is starting, connection is checked, then Acra is starting
	//      but somehow checking connection *here* fails thus failing the tests;
	//      everything else seems working since real requests to configured server
	//      are successful (when CRL verification is performed)
	// return nil, fmt.Errorf("CRL: Cannot fetch configured URI '%s'", uri)

PR #435, opened 2020-11-23
TLS authentication with certificates as identifiers

Lagovas:
Extended current TLS authentication where acra may starts with specified from cli params client_id which used for all incoming connections. This PR extends it and allow to start acra without static client_id and use data from certificates to map them to keystore identifiers/keys. PR contains 2 implementations of extracting data from certificates: common name and serial numbers, which should be chosen at startup (cli params not added yet). And one implementation of converting these values to acceptable format of ids using hex encoder (to allow non ascii printable values like serial numbers) and hash function (to allow values which longer than 256 length).
A little bit refactored unit-tests to re-use same common function which verify wrapper's logic, added helpers to generate tls certificates in golang to test all specific cases related to certificate fields (@iamnotacake you can re-use them if need^)

Done:
* validation incoming certificate (deny CA certificates or with incorrect keyUsage/extKeyUsage values)
* identifier fetching from certificates (distinguished name, serial numbers)
* encoding identifier values to acceptable formats of clientID by sha512 and hex encoding

Now AcraServer can use certificates from acra-connector connections when used `tls_client_id_from_cert` + `acraconnector_tls_transport_enable`, or from app's connection when used `tls_client_id_from_cert` + `acraconnector_transport_encryption_disable`. In both cases should be configured all necessary TLS params with ca/cert/key values.
Plus user can configure which strategy of mapping certificate metadata to clientID to use. To configure should set `tls_identifier_extractor_type` with `distinguished_name` or `serial_number` value. 
In first case will be used "Subject" field ordered and formatted according to [RFC2253](https://tools.ietf.org/html/rfc2253), hashed with sha512 (as hash function which provide longest digest and least collision probability) and encoded to hex value.
In second cases will be used serial number as 20-bytes integer value in big-endian order hashed with same function sha512 and encoded to hex value.
First strategy good to use when certificates may be rotated and allow to avoid rotation keys and accordingly data too.
Second strategy good for environments when Subject field not used in certificates.


network/tls_authentication.go
```diff
@@ -1,10 +1,73 @@
...
+	"github.com/cossacklabs/acra/keystore"
 )
 
+// CertificateIdentifierExtractor interface for implementations which should return identifier used for client's identification
+type CertificateIdentifierExtractor interface {
+	GetCertificateIdentifier(certificate *x509.Certificate) ([]byte, error)
+}
+
+// CommonNameExtractor implementation for CertificateIdentifierExtractor interface, which return CommonName as client's identifier
+type CommonNameExtractor struct{}
```
ilammy, line 18: I don't want to be that nitpicking jerk with 50 spelling comments for each PR, but the `certificate.Subject` this extractor returns is the "distinguished name". The "common name" is a part of a distinguished name. I believe this extractor should be renamed to avoid confusion. (Or maybe you intended to extract *just* the common name?)

network/tls_authentication_test.go
```diff
@@ -1,10 +1,76 @@
...
+	}
+	return workingDirectory
+}
+
+func getAcraWriterTestx509Certificate(t *testing.T) *x509.Certificate {
+	certPath := filepath.Join(getWorkingDirectory(t), "tests/ssl/acra-writer/acra-writer.crt")
+	certificateData, err := ioutil.ReadFile(certPath)
+	if err != nil {
+		t.Fatal(err)
+	}
```
ilammy: Hm.... Judging by the usage pattern of `getWorkingDirectory()`, why not convert it into some
```go
func readTestCertificateData(t *testing.T, path string) []byte {
	workingDirectory, err := os.Getwd()
	if err != nil {
		t.Fatal(err)
	}
	sslTestFolder := filepath.Join(workingDirectory, "tests/ssl")
	info, err := os.Lstat(sslTestFolder)
	// The "err" is likely to be os.ErrNotExist if the directory is mismatched,
	// so I guess both failures can be merged...
	if err != nil || !info.IsDir() {
		t.Fatalf("Current working directory '%s' is not Acra root directory", workingDirectory)
	}
	data, err := ioutil.ReadFile(filepath.Join(sslTestFolder, path))
	if err != nil {
		t.Fatal(err)
	}
	return data
}
```
which does all of the above checks and is easier to comprehend from tests?

I doubt that we'd need to check current working directory for anything else.

ilammy (commented):
I haven't read through the tests yet. However, overall idea of zero configuration is an appealing one.

(But I don't really like how unfriendly the resulting identifiers are. This is going to become a wart that everybody is aware of but it is impossible to change ever because customers will start depending on it.)

ilammy:
Also, projecting into the future, what will be a migration strategy for users which need to change their certificate subject?

Acra client IDs *per se* can be expected to be unchanging. However, the certificates might change due to some external reasons not connected with Acra. For example, the users might procure their certificates from a third-party CA, which one day adds or removes some field from distinguished names for newly issued certificates. The next certificate rotation will then result in a different client ID, resulting in unexpected service disruption.

It is easy to say "just don't change your certificate subject name", but that's kinda limiting and we'd better have a better answer. Some users might not want to get their own PKI just to run Acra without extra risks.

Do I remember correctly that client ID is **not** used for data encryption, or anything else other than locating the appropriate keys in the keystore? That is, if the certificate subject name changes, all the users need to do is to rename the keys to use the new computed client ID? If that's the case then certificates changes are not fatal. However, the users will need a way to 1) figure out client IDs for both old and new certificates, 2) rename the keys once they confirm the client ID change.

network/tls_authentication.go
Lagovas: yes, agree. I will change. main reason was to use "common name" because saw that this name usually used in discussing tls fields and identification

network/tls_authentication.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	if certificate.IsCA {
+		return ErrCACertificateUsed
+	}
+	// do we found any authentication keyUsage from KeyUsage field or from ExtKeyUsage parameters
+	// certificate may define any of them or both
+	definedAnyAuthenticationKeyUsage := false
+	keyUsageDefined := certificate.KeyUsage != 0
+	authenticationKeyUsage := certificate.KeyUsage&x509.KeyUsageDigitalSignature == 1
+	if keyUsageDefined && authenticationKeyUsage {
```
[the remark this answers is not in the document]
Lagovas: yes, agree) will simplify it

network/tls_authentication.go
```diff
@@ -0,0 +1,45 @@
...
+		definedAnyAuthenticationKeyUsage = true
+	}
+	extKeyUsageDefined := len(certificate.ExtKeyUsage) != 0
+	if extKeyUsageDefined {
```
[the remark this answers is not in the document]
Lagovas: +1, agree. I did it so because certificate.ExtKeyUsage may be nil and wasn't sure that forloop works with it. but checked now and it's okay)

Lagovas (commented):
> Acra client IDs per se can be expected to be unchanging. However, the certificates might change due to some external reasons not connected with Acra.

Yes, clientID should be unchanged, and this value used in acra-ee in tokenization as context data. And one of the future plans to merge clientID/zoneID usage and use them identically. And if user's want to use certificates as identifiers of keys, they need to take care of them in the same manner or choose another configuration. Certificates don't contain any unchanged data which may be used for that and we provide 2 ways (for now but maybe extendable, I thought, probably, to introduce plugins usage for that, but left it for later, because probably it will be not in demand by users). Serial numbers for cases if users don't use DN. But it force to rotate/change acra keys. And to use DN which allow to rotate certificate (new private key, new public key, signature, serial number) and re-use same DN identifier and avoid rotating/changing acra's keys.

And yes, you raised important question about instructions/tools for users that should allow changing DN values to not break their relation to acra's encryption keys. Probably it will be implemented in next PRs when we discuss interface and use cases of that.

ilammy:
> Yes, clientID should be unchanged, and this value used in acra-ee in tokenization as context data. And one of the future plans to merge clientID/zoneID usage and use them identically.

Another point here is that while client IDs do not affect the encrypted data, zone IDs *do* affect it (they are used as associated context for AcraStructs). Thus, if zone IDs need to change, the data will need to be re-encrypted.


> And to use DN which allow to rotate certificate (new private key, new public key, signature, serial number) and re-use same DN identifier and avoid rotating/changing acra's keys.

Here's an idea that came to me just now. How about letting the user choose which parts of the DN should be included into synthetic client ID? That way the users might get a chance to have some constant part of their certificates which they can commit to being immutable. Say, only the "organisation name" and "common name" parts, ignoring everything else. Though, this comes with issues on its own: how to deal with custom DNs, how to format the extracted data pieces, etc.


Anyhow, I don't feel that this PR is a place for design discussions so I'll leave it at that. Let's discuss this in chat or something. For now, I believe it's enough to mark this functionality experimental and require the users to explicitly opt-in.

network/tls_authentication.go
```diff
@@ -1,10 +1,73 @@
...
+	} else if len(identifier) > (keystore.MaxClientIDLength / 2) {
+		out = make([]byte, hex.EncodedLen(sha512.Size))
+		h := sha512.New()
+		if _, err := h.Write(identifier); err != nil {
+			return nil, err
+		}
+		identifier = h.Sum(nil)
+	} else {
+		out = make([]byte, hex.EncodedLen(len(identifier)))
+	}
```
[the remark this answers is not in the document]
Lagovas: hm... probably agree, will change to always hash DN. But not understand which prefixes you want to add? Which collisions it will help to avoid?

network/tls_authentication_test.go
Lagovas: first version was like that until I started tests from IDE and working directory automatically configured to `network/ ` when I run tests from network package) And I thought that other developers may meet these problems too and will have to change working directory every time. So decided to simplify it for all

Lagovas:
> Here's an idea that came to me just now. How about letting the user choose which parts of the DN should be included into synthetic client ID? That way the users might get a chance to have some constant part of their certificates which they can commit to being immutable. Say, only the "organisation name" and "common name" parts, ignoring everything else. Though, this comes with issues on its own: how to deal with custom DNs, how to format the extracted data pieces, etc.

Yeah, there a lot of ways to do it and that is why it implemented with interfaces and several implementations which may be replaced in a future. But for now we don't really know how users will use it and it is redundant to implement a lot of strategies when most of them will not use

ilammy:
> But not understand which prefixes you want to add? Which collisions it will help to avoid?

I was thinking about prefixing hashed identifier values with a constant:

     h := sha512.New()
    +h.Write([]byte("client ID:"))
     h.Write(identifier)
     identifier = h.Sum(nil)

so that the hashes produced here are less likely to collide with any other SHA-512 hashes, computed and used elsewhere for other purposes. This binds hashes to some sort of "context".

It's not strictly necessary, but I think it's still good hygiene for cases when hash functions are used as something other than a simple checksum.

Lagovas:
> I was thinking about prefixing hashed identifier values with a constant

Agree, will add. I thought that maybe you talk about prefix for result value, like `clientID_<hex_value>`, which will help to filter/search values in KV storages and separate from other values, or a little bit human-readable values for FS storage) But probably will add both prefixes

Lagovas:
> It's not strictly necessary, but I think it's still good hygiene for cases when hash functions are used as something other than a simple checksum.

Hm, I remembered why I didn't do it... Main thought was to not complicate generating these values and allow to automate it by existing tools, without any acra-tools. Because user will need to call acra-keystore and generate keys specifying correct client_id and it requires to know how to convert info from certificate to result value. If will be used tls authentication then user will know serial number or DN, for example, and then generate acra keys with this metadata. And it easier to just extract info from cert by `openssl` and pass it to `sha512sum` and use the result as `--client_id` param for keystore. 
But after changes, it will need more steps or we should provide some tool for that

ilammy:
> I thought that maybe you talk about prefix for result value, like clientID_<hex_value>

That too. Though, I don't know how much useful it will be since client IDs like "04f05825..." (128 characters!) are going to be very conspicuous, and I don't imagine why we'd need an easy way to filter out those specifically to see which 'normal' IDs remain.

> Hm, I remembered why I didn't do it... Main thought was to not complicate generating these values and allow to automate it by existing tools

You know, you just made me realize that those prefixes are probably not needed after all.

Let's look at it like this: if client ID syntax _was_ flexible enough, we'd be using DNs directly. The certificate content is basically public, so there is no point in 'hiding' the value of DN, it can be assumed to be known to the attacker (as well as any hash derived from it, in whatever complex algorithm is there). So why all the additional complexity?

Please disregard my comments about prefixes before hashing.

cmd/acra-server/acra-server.go
```diff
@@ -123,6 +124,8 @@ func main() {
 	tlsDbSNIOld := flag.String("tls_db_sni", "", "Expected Server Name (SNI) from database (deprecated, use \"tls_database_sni\" instead)")
 	tlsDbCert := flag.String("tls_database_cert", "", "Path to client TLS certificate shown to database during TLS handshake (overrides \"tls_cert\")")
 	tlsDbKey := flag.String("tls_database_key", "", "Path to private key of the TLS certificate used to connect to database (see \"tls_database_cert\")")
+	tlsUseClientIDFromCertificate := flag.Bool("tls_client_id_from_cert", false, "Use certificate info as clientID for encryption/decryption when used. Use certificate from connector's connection if acraconnector_tls_transport_enable used or from client's connection if acraconnector_transport_encryption_disable used")
+	tlsIdentifierExtractorType := flag.String("tls_identifier_extractor_type", network.IdentifierExtractorTypeDistinguishedName, fmt.Sprintf("Type of identifier extractor from TLS certificate to use as ClientID (%s|%s)", network.IdentifierExtractorTypeDistinguishedName, network.IdentifierExtractorTypeSerialNumber))
```
ilammy, line 128: How about making a list of all extractor types and formatting it here:

```go

fmt.Sprintf("...as ClientID (%s)", strings.Join(IdentifierExtractorTypes, "|"))

var IdentifierExtractorTypes []string = []string{
	IdentifierExtractorTypeDistinguishedName,
	IdentifierExtractorTypeSerialNumber,
}
```

That way it will be a bit more obvious and easy thing to add a new extractor type.

In some intermediate commit this has been located directly in the `network/tls_authentication.go` so it could be kinda okay, but now developers can easily forget to edit `cmd/acra-server/acra-server.go` (and whatever other files there are) when they add a new identifier extractor type.

tests/test.py
```diff
@@ -5167,6 +5217,408 @@ def testPgPlaceholders(self):
...
+class TestTLSAuthenticationWithConnectorBySerialNumber(TLSAuthenticationBySerialNumberMixin,
+                                                       TestTLSAuthenticationWithConnectorByDistinguishedName):
+    pass
+
+
+class TestTLSAuthenticationDirectlyToAcraByDistinguishedName(TLSAuthenticationByDistinguishedNameMixin, BaseTestCase):
+    """
+    Tests environment without connector, when client's app connect to db through acra-server with TLS and acra-server extracts clientID from client's certificate
+    instead using from --clientID CLI param
+    """
```
ilammy, line 5440: It seems like quite a few lines of code can be shared between `TestTLSAuthenticationWithConnectorByDistinguishedName` and `TestTLSAuthenticationDirectlyToAcraByDistinguishedName`.

ilammy (commented):
So far so good 💪 👍

cmd/acra-server/acra-server.go
```diff
@@ -123,6 +123,7 @@ func main() {
 	tlsDbSNIOld := flag.String("tls_db_sni", "", "Expected Server Name (SNI) from database (deprecated, use \"tls_database_sni\" instead)")
 	tlsDbCert := flag.String("tls_database_cert", "", "Path to client TLS certificate shown to database during TLS handshake (overrides \"tls_cert\")")
 	tlsDbKey := flag.String("tls_database_key", "", "Path to private key of the TLS certificate used to connect to database (see \"tls_database_cert\")")
+	tlsUseClientIDFromCertificate := flag.Bool("tls_client_id_from_cert", false, "Use certificate info as clientID for encryption/decryption when used. Use certificate from connector's connection if acraconnector_tls_transport_enable used or from client's connection if acraconnector_transport_encryption_disable used")
```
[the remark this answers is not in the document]
Lagovas: it switch connector to use TLS as transport instead secure session. but there several ways to get clientID from such communication: from static parameter --clientID or from certificate. If used `tls_client_id_from_cert`, then will be used from certificate. Otherwise should be provided non-empty `--clientID` (previous version).

cmd/acra-server/acra-server.go
```diff
@@ -277,25 +279,31 @@ func main() {
 			log.WithError(err).Errorln("Can't initialize identifier converter")
 			os.Exit(1)
 		}
-		tlsWrapper, err := network.NewTLSAuthenticationConnectionWrapper(clientTLSConfig, dbTLSConfig, network.CommonNameExtractor{}, idConverter)
+		tlsWrapper, err = network.NewTLSAuthenticationConnectionWrapper(dbTLSConfig, clientTLSConfig, network.CommonNameExtractor{}, idConverter)
```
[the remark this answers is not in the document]
Lagovas: because I mismatched order and tests failed because dbTLSConfig configured with SNI which verified in connection to db and useless for wrapping connection with app

tests/test.py
Lagovas: agree, removed duplicated `testAcrastruct....`

cmd/acra-server/acra-server.go
Lagovas: agree, added

decryptor/base/proxy.go
```diff
@@ -101,6 +94,34 @@ type ClientSession interface {
 	SetProtocolState(state interface{})
 }
 
+// TLSConnectionWrapper used by proxy to wrap raw connections to TLS when intercepts client/database request about switching to TLS
+// Reuse network.ConnectionWrapper to explicitly force TLS usage by name
+type TLSConnectionWrapper interface {
+	WrapDBConnection(ctx context.Context, conn net.Conn) (net.Conn, error)
+	WrapClientConnection(ctx context.Context, conn net.Conn) (net.Conn, []byte, error) // conn, ClientID, error
```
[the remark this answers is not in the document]
Lagovas: agree, updated

decryptor/mysql/response_proxy.go
```diff
@@ -174,8 +173,7 @@ func NewMysqlProxy(session base.ClientSession, decryptor base.Decryptor, setting
 		acracensor:             setting.Censor(),
 		clientConnection:       session.ClientConnection(),
 		dbConnection:           session.DatabaseConnection(),
-		clientTLSConfig:        tweakTLSConfigForMySQL(setting.ClientTLSConfig()),
-		dbTLSConfig:            tweakTLSConfigForMySQL(setting.DatabaseTLSConfig()),
```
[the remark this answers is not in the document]
Lagovas: no, removed. thanks for your attention

configs/acra-server.yaml
```diff
@@ -131,6 +131,9 @@ tls_client_ca:
 # Path to server TLS certificate presented to AcraConnectors (overrides "tls_cert")
 tls_client_cert: 
 
+# Use certificate info as clientID for encryption/decryption when used. Use certificate from connector's connection if acraconnector_tls_transport_enable used or from client's connection if acraconnector_transport_encryption_disable used
```
storojs72, line 134: Interesting that we try to remove `client_id` setting and add two configurations instead

network/tls_authentication_test.go
```diff
@@ -0,0 +1,245 @@
...
+		t.Fatal("Not denied certificate with acceptable KeyUsage/ExtKeyUsage")
+	}
+
+	// all purposes except digital signature
+	invalidKeyUsage := 0xFFFFFFFF ^ x509.KeyUsageDigitalSignature
+	// all purposes except ClientAuth
+	invalidExtKeyUsage := []x509.ExtKeyUsage{
+		x509.ExtKeyUsageAny,
+		x509.ExtKeyUsageServerAuth,
+		//x509.ExtKeyUsageClientAuth,
```
storojs72, line 135: Nice way to show what is excluded!

network/tls_authentication_test.go
```diff
@@ -0,0 +1,245 @@
...
+	if err := ValidateClientsAuthenticationCertificate(&x509.Certificate{IsCA: true}); err != ErrCACertificateUsed {
+		t.Fatal("Not denied certificate for CA purposes")
+	}
+	// test without any key usage
+	if err := ValidateClientsAuthenticationCertificate(&x509.Certificate{}); err != ErrMissedAuthenticationKeyUsage {
+		t.Fatal("Not denied certificate with acceptable KeyUsage/ExtKeyUsage")
+	}
+
+	// all purposes except digital signature
+	invalidKeyUsage := 0xFFFFFFFF ^ x509.KeyUsageDigitalSignature
```
storojs72, line 130: This seems to be tricky :)

storojs72 (commented):
I like this improvement. And since it may cause some inconveniences for users, as pointed by @ilammy , it is completely configurable (to use or not to use), so old users can still use existing approach. I think it is a right common direction in Acra development to be integrated closely with PKI. Much more users are familiar with TLS better than with Acra, and if we do not force them to handle one more security credential (`cliend_id`) they will say "Thanks" to us.

@Lagovas, I guess that this improvement will come to Translator and in enterprise version too in future PRs, right?

configs/acra-server.yaml
ilammy: > Interesting that we try to remove client_id setting and add two configurations instead

I guess in the future they will be enabled by default (making them opt-out, rather than opt-in). That way it will become zero configuration to most users.

ilammy (commented):
I approve the changes 👍 but I will hold onto the actual green check mark until the CI is green. Then this PR should be on a merge track.

configs/acra-server.yaml
Lagovas: actually we will not remove this parameter because it will allow to use acra-server instead acra-connector as close to app encryption proxy. for example if app can't encrypt data itself and client wants to setup acra-server close to app on the same node/container/machine and transparently encrypt data for "e2e" without app modification. For that they can run acra-server with `--client_id` parameter and with limited key storage they can do that)

cmd/acra-server/acra-server.go
```diff
@@ -123,6 +125,8 @@ func main() {
 	tlsDbSNIOld := flag.String("tls_db_sni", "", "Expected Server Name (SNI) from database (deprecated, use \"tls_database_sni\" instead)")
 	tlsDbCert := flag.String("tls_database_cert", "", "Path to client TLS certificate shown to database during TLS handshake (overrides \"tls_cert\")")
 	tlsDbKey := flag.String("tls_database_key", "", "Path to private key of the TLS certificate used to connect to database (see \"tls_database_cert\")")
+	tlsUseClientIDFromCertificate := flag.Bool("tls_client_id_from_cert", false, "Use certificate info as clientID for encryption/decryption when used. Use certificate from connector's connection if acraconnector_tls_transport_enable used or from client's connection if acraconnector_transport_encryption_disable used")
```
vixentael, line 128:
```suggestion
	tlsUseClientIDFromCertificate := flag.Bool("tls_client_id_from_cert", false, "Extract clientID from TLS certificate. Take TLS certificate from AcraConnector's connection if acraconnector_tls_transport_enable is TRUE; otherwise take TLS certificate from application's connection if acraconnector_transport_encryption_disable is TRUE")
```

cmd/acra-server/acra-server.go
```diff
@@ -123,6 +125,8 @@ func main() {
...
+	tlsIdentifierExtractorType := flag.String("tls_identifier_extractor_type", network.IdentifierExtractorTypeDistinguishedName, fmt.Sprintf("Type of identifier extractor from TLS certificate to use as ClientID (%s)", strings.Join(network.IdentifierExtractorTypesList, "|")))
```
vixentael, line 129:
```suggestion
	tlsIdentifierExtractorType := flag.String("tls_identifier_extractor_type", network.IdentifierExtractorTypeDistinguishedName, fmt.Sprintf("Decide which field of TLS certificate to use as ClientID (%s)", strings.Join(network.IdentifierExtractorTypesList, "|")))
```

cmd/acra-server/acra-server.go
```diff
@@ -271,19 +277,45 @@ func main() {
...
+			os.Exit(1)
+		}
+		useForClientID := *tlsUseClientIDFromCertificate && *noEncryptionTransport
+		proxyTLSWrapper = base.NewTLSConnectionWrapper(useForClientID, tlsWrapper)
+		log.WithField("use_client_id_from_cert", useForClientID).Infoln("Loaded TLS configuration")
+	}
+	if *useTLS {
+		if *tlsUseClientIDFromCertificate {
+			config.ConnectionWrapper = tlsWrapper
+			log.Println("Selecting transport: use TLS transport wrapper with clientID from certificates")
```
vixentael, line 302:
```suggestion
			log.Println("Selecting transport: use TLS transport wrapper, extract clientID from TLS certificate")
```

cmd/acra-server/acra-server.go
```diff
@@ -271,19 +277,45 @@ func main() {
...
+		} else {
+			log.Println("Selecting transport: use TLS transport wrapper with static clientID")
```
vixentael, line 304:
```suggestion
			log.Println("Selecting transport: use TLS transport wrapper with static clientID, provided as parameter")
```

vixentael (approved):

Lagovas:
@vixentael , I updated messages by 1 commit to avoid running tests by each commit from suggestions

vixentael (approved):
Thank you @Lagovas

storojs72 (approved):

CHANGELOG_DEV.md
```diff
@@ -1,3 +1,9 @@
+## 0.85.0 - 2020-12-08
+
+- Extended TLS support and mapping to clientID for client's key selection purposes
+  - Strategy of extraction metadata from certificates for mapping to clientID: `tls_identifier_extractor_type` (default: `distinguished_name` and `serial_number`)
```
iamnotacake, line 4: I guess we need to tell users that `and 'serial_number'` is another possible value, and it has nothing to do with default one, so maybe tell explicitly that `serial_number` is a completely different option?
```suggestion
  - Strategy of extraction metadata from certificates for mapping to clientID: `tls_identifier_extractor_type` (default: `distinguished_name`, another option: `serial_number`)
```

Lagovas: it's just changelog_dev, not public cli description of param. but okay)

ilammy (approved):

PR #436, opened 2020-12-03
TLS validation on new connections

iamnotacake:
#### Big refactor of `generate_tls_keys.sh`
* Put cert name into CN (like "Test leaf certificate (acra-server)")
* Generate certificate database for OpenSSL so its OCSP server can work with it, also required to generate CRL
* Add certs "ocsp-responder" and "acra-writer-revoked"; the second one is for tests to make sure acra-server handles revoked certificates
* Generate CRL (Certificate Revocation List) containing that one revoked certificate, signed by CA

#### Start OCSP & CRL servers when teststing with TLS enabled
Update `tests/test.py`:
* Add global vars containing OCSP certificate and key, as well as certificate database
* Add function to launch `openssl ocsp` server along with `acra-server` and `acra-connector`, but only if TLS is enabled
* Use Python's package `http` to create HTTP server, CRL verifier will be able to fetch CRL from there

### Implement all the things

This PR contains implementation of TLS certificate validators using using OCSP protocol (RFC6960) and CRLs (RFC5280). It includes mentioned validators implementation, as well as way to configure them in Acra (server and connector) using command line flags.

OCSP flags (with default value and description):
* `--tls_ocsp_url`, "", OCSP service URL
* `--tls_ocsp_client_url`, "", OCSP service URL, for client/connector certificates only
* `--tls_ocsp_database_url`, "", OCSP service URL, for database certificates only
* `--tls_ocsp_required`, "denyUnknown", How to treat certificates unknown to OCSP: <denyUnknown|allowUnknown|requireGood>
  `denyUnknown` -> deny certificate if OCSP server responded with "unknown"
  `allowUnknown` -> allow certificate not known by OCSP server, but all servers returning "unknown" is considered an error
  `requireGood` -> all OCSP servers we know about (the one from configuration and those listed in certificate) should respond with "good", otherwise reject the certificate
* `--tls_ocsp_from_cert`, "prefer", How to treat OCSP server described in certificate itself: <use|trust|prefer|ignore>
  `use` -> use OCSP servers from certificates, but after the one from configuration (if configured)
  `prefer` -> like `use` but try those from certificate first
  `trust` -> like `prefer`, but don't even try configured server if the one from certificate tells us the certificate is valid
  `ignore` -> don't use OCSP servers from certificates
* `--tls_ocsp_check_only_leaf_certificate`, false, Put 'true' to check only final/last certificate, or 'false' to check the whole certificate chain using OCSP

CRL flags (with default value and description):
* `--tls_crl_url`, "", URL of the Certificate Revocation List (CRL) to use
* `--tls_crl_client_url`, "", URL of the Certificate Revocation List (CRL) to use, for client/connector certificates only
* `--tls_crl_database_url`, "", URL of the Certificate Revocation List (CRL) to use, for database certificates only
* `--tls_crl_from_cert`, "use", How to treat CRL URL described in certificate itself: <use|trust|prefer|ignore>
  `use` -> use CRL distribution point(s) servers from certificates, but after the one from configuration (if configured)
  `prefer` -> like `use` but try those from certificate first
  `trust` -> like `prefer`, but don't even try configured CRL URL if the one from certificate does not contain the certificate
  `ignore` -> don't use CRL distribution points from certificates
* `--tls_crl_check_only_leaf_certificate`, false, Put 'true' to check only final/last certificate, or 'false' to check the whole certificate chain using CRL
* `--tls_crl_cache_size`, 16, How many CRLs to cache in memory (use 0 to disable caching)
* `--tls_crl_cache_time`, 0, How long to keep CRLs cached, in seconds (use 0 to disable caching, maximum: 300 s)

Please note that client- and DB-specific URIs are only configurable in acra-server; acra-connector does not have these "two sides" so there is only one URI for OCSP and one for CRL.

Integration tests (`tests/test.py`), when launching acra-server or acra-connector, add a couple of options to either enable both verifiers (if TLS was enabled in tests) or disable verification completely. Disabling by default is done to avoid situations where test certificate contains some OCSP or CRL URIs and Acra cannot access them since appropriate services were not started thus denying actually valid test certificates and failing the tests.

There are also unit tests (`network/{ocsp_test.go,crl_test.go}`) to cover the behavior of both verifiers.

### TODO:
- [X] Update error messages
- [X] Update help text about CLI flags
- [X] Add certificate group that contains intermedidate CA (it existed at some point, can be used in integration or go tests, but the tool that generated them was kinda too complex, and there were too many redundant files generated, I guess `generate_tls_keys.sh` should be updated, or some simple util written for that purpose instead)
- [ ] Add few integration tests (there were attempts, but the fact that `tests/test.py` sets up SSL/TLS configuration on script start, not inside tests, does not allow writing custom tests that use custom certificates, for example to test acra-server + acra-connector with revoked certificate; possible simple solution is to skip these new tests by default but run them (and only them) separately from inside the `.circleci/integration.sh` script after usual tests)
- [ ] Different random updates based on comments (mostly done)
- [X] Update CHANGELOG_DEV.md


iamnotacake:
This PR will likely have some little conflicts with #435, so one of us will have to merge `master` into PR's branch.
And based on all this mess, I guess #435 will be merged first and then I'll update this PR's branch (:

.circleci/check_configs.sh
```diff
@@ -9,6 +9,8 @@ function compare_configs() {
      cmp_status="$?"
      if [[ "${cmp_status}" != "0" ]]; then
         status=1
+        # Show diff in logs, easier to spot the difference
+        diff -U3 ${folder_a}/acra-${cmd}.yaml ${folder_b}/acra-${cmd}.yaml
```
vixentael, line 13: 👍

cmd/acra-connector/acra-connector.go
```diff
@@ -229,6 +229,14 @@ func main() {
...
+	tlsOcspURL := flag.String("tls_ocsp_url", "", "OCSP service URL")
+	tlsOcspRequired := flag.String("tls_ocsp_required", "denyUnknown", "Whether we need OCSP response in order to accept certificate")
+	tlsOcspFromCert := flag.String("tls_ocsp_from_cert", "prefer", "How should we treat OCSP server described in certificate itself")
+	tlsOcspCheckWholeChain := flag.Bool("tls_ocsp_check_whole_chain", false, "Whether to check whole certificate chain, false = only end certificate")
+	tlsCrlURL := flag.String("tls_crl_url", "", "CRL URL")
+	tlsCrlFromCert := flag.String("tls_crl_from_cert", "use", "How should we treat CRL URL described in certificate itself")
+	tlsCrlCacheSize := flag.Int("tls_crl_cache_size", 16, "Size of in-memory LRU cache for storing fetched CRLs, 0 = unlimited")
+	tlsCrlCacheTime := flag.Int("tls_crl_cache_time", 5, "How long cached CRL is considerend valid, seconds, max = 300")
```
vixentael, line 239: @shadinua suggests to start from 0, where 0 is "don't cache". Change description.

cmd/acra-server/acra-server.go
```diff
@@ -123,6 +123,18 @@ func main() {
...
+	tlsOcspDbURL := flag.String("tls_ocsp_database_url", "", "OCSP service URL, for database certificates only")
+	tlsOcspRequired := flag.String("tls_ocsp_required", "denyUnknown", "Whether we need OCSP response in order to accept certificate")
+	tlsOcspFromCert := flag.String("tls_ocsp_from_cert", "prefer", "How should we treat OCSP server described in certificate itself")
+	tlsOcspCheckWholeChain := flag.Bool("tls_ocsp_check_whole_chain", false, "Whether to check whole certificate chain, false = only end certificate")
+	tlsCrlURL := flag.String("tls_crl_url", "", "CRL URL")
+	tlsCrlClientURL := flag.String("tls_crl_client_url", "", "CRL URL, for client certificates only")
+	tlsCrlDbURL := flag.String("tls_crl_database_url", "", "CRL URL, for database certificates only")
+	tlsCrlFromCert := flag.String("tls_crl_from_cert", "use", "How should we treat CRL URL described in certificate itself")
+	tlsCrlCacheSize := flag.Int("tls_crl_cache_size", 16, "Size of in-memory LRU cache for storing fetched CRLs, 0 = unlimited")
+	tlsCrlCacheTime := flag.Int("tls_crl_cache_time", 5, "How long cached CRL is considerend valid, seconds, max = 300")
```
vixentael, line 137: same

cmd/acra-connector/acra-connector.go
```diff
@@ -229,6 +229,14 @@ func main() {
 	tlsCert := flag.String("tls_cert", "", "Path to certificate")
 	tlsAcraserverSNI := flag.String("tls_acraserver_sni", "", "Expected Server Name (SNI) from AcraServer")
 	tlsAuthType := flag.Int("tls_auth", int(tls.RequireAndVerifyClientCert), "Set authentication mode that will be used in TLS connection with AcraServer/AcraTranslator. Values in range 0-4 that set auth type (https://golang.org/pkg/crypto/tls/#ClientAuthType). Default is tls.RequireAndVerifyClientCert")
+	tlsOcspURL := flag.String("tls_ocsp_url", "", "OCSP service URL")
+	tlsOcspRequired := flag.String("tls_ocsp_required", "denyUnknown", "Whether we need OCSP response in order to accept certificate")
```
vixentael, line 233:
```suggestion
	tlsOcspRequired := flag.String("tls_ocsp_required", "denyUnknown", "Whether we need OCSP response in order to accept certificate; allowed options <denyUnknown|allowUnknown|all>")
```

cmd/acra-connector/acra-connector.go
```diff
@@ -229,6 +229,14 @@ func main() {
...
+	tlsOcspFromCert := flag.String("tls_ocsp_from_cert", "prefer", "How should we treat OCSP server described in certificate itself")
+	tlsOcspCheckWholeChain := flag.Bool("tls_ocsp_check_whole_chain", false, "Whether to check whole certificate chain, false = only end certificate")
```
vixentael, line 235: unclear, rewrite
```suggestion
	tlsOcspCheckWholeChain := flag.Bool("tls_ocsp_check_whole_chain", false, "Put 'true' to check the whole certificate chain, or 'false' to check only final/last certificate")
```

cmd/acra-connector/acra-connector.go
```diff
@@ -229,6 +229,14 @@ func main() {
 	tlsCert := flag.String("tls_cert", "", "Path to certificate")
 	tlsAcraserverSNI := flag.String("tls_acraserver_sni", "", "Expected Server Name (SNI) from AcraServer")
 	tlsAuthType := flag.Int("tls_auth", int(tls.RequireAndVerifyClientCert), "Set authentication mode that will be used in TLS connection with AcraServer/AcraTranslator. Values in range 0-4 that set auth type (https://golang.org/pkg/crypto/tls/#ClientAuthType). Default is tls.RequireAndVerifyClientCert")
+	tlsOcspURL := flag.String("tls_ocsp_url", "", "OCSP service URL")
+	tlsOcspRequired := flag.String("tls_ocsp_required", "denyUnknown", "Whether we need OCSP response in order to accept certificate")
+	tlsOcspFromCert := flag.String("tls_ocsp_from_cert", "prefer", "How should we treat OCSP server described in certificate itself")
```
vixentael, line 234: The parameter name is not clear for me, the description is not clear for me. I think we might add some verb here. Also, it make sense to list options and their meaning <use|trust|prefer|ignore>.

network/tls_wrapper.go
```diff
@@ -49,19 +49,39 @@ type TLSConnectionWrapper struct {
...
+	flag.StringVar(&tlsOcspURL, "tls_ocsp_url", "", "OCSP service URL")
+	flag.StringVar(&tlsOcspClientURL, "tls_ocsp_client_url", "", "OCSP service URL, for client certificates only")
+	flag.StringVar(&tlsOcspDbURL, "tls_ocsp_database_url", "", "OCSP service URL, for database certificates only")
+	flag.StringVar(&tlsOcspRequired, "tls_ocsp_required", "yes", "Whether we need OCSP response in order to accept certificate")
+	flag.StringVar(&tlsOcspFromCert, "tls_ocsp_from_cert", "prefer", "How should we threat OCSP server described in certificate itself")
+	flag.BoolVar(&tlsOcspCheckWholeChain, "tls_ocsp_check_whole_chain", false, "Whether to check whole certificate chain, false = only end certificate")
+	flag.StringVar(&tlsCrlURL, "tls_crl_url", "", "CRL URL")
+	flag.StringVar(&tlsCrlFromCert, "tls_crl_from_cert", "use", "How should we treat CRL URL described in certificate itself")
+	flag.IntVar(&tlsCrlCacheSize, "tls_crl_cache_size", 16, "Size of in-memory LRU cache for storing fetched CRLs, 0 = unlimited")
+	flag.IntVar(&tlsCrlCacheTime, "tls_crl_cache_time", 60, "How long cached CRL is considerend valid, seconds, max = 300")
```
ilammy, line 84: This is strangely familiar to the list from `acra-server` and friends 🤔  🤔  🤔

network/crl_test.go
```diff
@@ -0,0 +1,468 @@
...
+	// TestOCSPKeyFilename3     = "intermediate-ocsp-responder/intermediate-ocsp-responder.key"
+
+	// Make sure you use either of these groups and don't mix their components
+)
+
+type TestCertGroup struct {
+	prefix string
+
+	ca  string
+	ica string // may be empty (missing)
```
ilammy, line 63: IMO, `intermediateCA` will be more obvious when someone reads this code half a year from now.

network/crl_test.go
```diff
@@ -0,0 +1,468 @@
...
+	// TODO add certificates for this group, fix paths, uncomment related code
+	// A different group, uses intermediate CA
+	// TestCertPrefix3          = "../tests/ssl/group-with-intermediate"
+	// TestCACertFilename3      = "ca/ca.crt"
+	// TestICACertFilename3     = "intermediate-ca/intermediate-ca.crt"
+	// TestCertFilename3        = "acra-writer/acra-writer.crt"
+	// TestRevokedCertFilename3 = "acra-writer-revoked/acra-writer-revoked.crt"
+	// TestCRLFilename3         = "intermediate-ca/crl.pem"
+	// TestOCSPCertFilename3    = "intermediate-ocsp-responder/intermediate-ocsp-responder.crt"
+	// TestOCSPKeyFilename3     = "intermediate-ocsp-responder/intermediate-ocsp-responder.key"
```
ilammy, line 54: For mass commenting out of code it's more tidy to use `/* block comments */` since they give nicer diffs.

Another alternative is to move all disabled tests into a separate file `crl_with_intermediate_test.go` or something, and then disable its compilation with a line in the beginning of the file:
```go
// +build testIntermediateCA
```
Then at least `gofmt` will do its job, and the diff will be equally tiny.

Though, this has problems of its own, like what to do with all the shared test utilities. Or are they visible to all test files?

network/crl_test.go
```diff
@@ -0,0 +1,468 @@
...
+
+	httpServer := &http.Server{
+		Handler: mux,
+	}
+	go func() {
+		httpServer.Serve(listener)
+	}()
+
+	// Give the server time to start, otherwise this test will fail sometimes
+	time.Sleep(time.Millisecond * 100)
```
ilammy, line 185: Why is this necessary, given that the listener socket is already created? Any connections should be queued until the server goes up.

ilammy (commented):
Overall, I like it, despite all the nitty comments that follow.

Good Bash-fu you’ve got here! As well as good error-checking, logging, and testing discipline :+1: Much kudos to you for that, @iamnotacake.

Also, those are some nice, informative commit messages. They and the commit split make this PR easy to review.

As for the command-line options, @shadinua – see? that’s *the correct* Shad this time! – at some point has proposed to maintain CHANGELOG_DEV.md. All these new options seem like a good addition for that. Or, well, given the quality of the PR, it can be used as a reference on its own when writing documentation…

Finally, do I understand it correctly that this PR supersedes #431?

And finally finally, I think it will be okay to add more of more complex tests in a separate PR, just to avoid making this one a long-runner again (and to deal with potential merge conflicts earlier).

network/crl_test.go
```diff
@@ -0,0 +1,468 @@
...
+
+func getValidTestChain(t *testing.T) ([][]byte, [][]*x509.Certificate) {
+	return getTestChain(
+		t,
+		TestCertPrefix,
+		TestCertFilename,
+		TestCACertFilename)
+}
+
+// func getValidTestChain3(t *testing.T) ([][]byte, [][]*x509.Certificate) {
```
Lagovas, line 251: same about commented functions

network/crl_test.go
```diff
@@ -0,0 +1,468 @@
...
+	expectOk("http://127.0.0.1/main_crl.pem", "use", 1, 5)
+	expectOk("", "use", 1, 0)
+	expectOk("", "use", 1, 1)
+	expectOk("", "use", 1, 300)
+
+	expectErr("htt://invalid url", "use", 1, 5)
+	expectErr("", "IgNoRe", 1, 5)
+	expectErr("", "use", 1, -1)
+	expectErr("", "use", 1, -10)
+	expectErr("", "use", 1, 301)
```
Lagovas, line 307: will be good to have comments about what we test here and why it should fail

network/crl_test.go
```diff
@@ -0,0 +1,468 @@
...
+	expectErr := func(uri, fromCert string, cacheSize, cacheTime int) {
+		config, err := NewCRLConfig(uri, fromCert, cacheSize, cacheTime)
+		if config != nil || err == nil {
+			t.Logf("uri=%v, fromCert=%v, cacheSize=%v, cacheTime=%v\n", uri, fromCert, cacheSize, cacheTime)
+			t.Logf("config=%v, err=%v\n", config, err)
+			t.Fatal("Got unexpected result or `nil` error")
+		}
+	}
+
+	expectOk("", "use", 1, 5)
```
Lagovas, line 296: will be nice to use constants instead "use" to understand where we can find all acceptable values and which are not correct.

network/ocsp.go
```diff
@@ -0,0 +1,334 @@
...
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/crypto/ocsp"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+)
+
+// Errors returned by CRL verifier
+var (
+	ErrInvalidConfigOCSPRequired   = errors.New("Invalid `ocsp_required` value")
```
Lagovas, line 33: same, lowercase for error msgs

network/ocsp.go
```diff
@@ -0,0 +1,334 @@
...
+	ErrInvalidConfigAllRequiresURI = errors.New("When passing `--tls_ocsp_required=all`, URL is mandatory")
+	ErrOCSPRequiredAllButGotError  = errors.New("Cannot query OCSP server, but --tls_ocsp_required=all was passed")
+	ErrOCSPUnknownCertificate      = errors.New("OCSP server doesn't know about certificate")
+	ErrOCSPNoConfirms              = errors.New("None of OCSP servers confirmed the certificate")
+)
+
+// --tls_ocsp_required=<denyUnknown|allowUnknown|all>
+const (
+	// Deny certificates now known by OCSP server(s)
+	ocspRequiredYesStr = "denyUnknown"
```
Lagovas, line 44: same, better to make them public and use in descriptions and tests instead hardcoded values

network/ocsp.go
```diff
@@ -0,0 +1,334 @@
...
+	if uri != "" {
+		log.Debugf("OCSP: Using server '%s'", uri)
+
+		httpClient := &http.Client{}
+		_, err := httpClient.Head(uri)
+		if err != nil {
+			log.WithError(err).WithField("uri", uri).Warnln("OCSP: Cannot reach configured server")
+			// TODO return error after issues with failing tests are fixed;
+			//      the issue is the same as in NewCRLConfig()
+			// return nil, errors.New("Cannot reach configured OCSP server")
```
Lagovas, line 132: same here

network/ocsp.go
```diff
@@ -0,0 +1,334 @@
...
+		return nil, err
+	}
+	ocspURL, err := url.Parse(ocspServerURL)
+	if err != nil {
+		return nil, err
+	}
+	httpRequest.Header.Add("Content-Type", "application/ocsp-request")
+	httpRequest.Header.Add("Accept", "application/ocsp-response")
+	httpRequest.Header.Add("host", ocspURL.Host)
+	httpClient := &http.Client{}
```
Lagovas, line 186: why not initialize it once as a field of DefaultOCSPClient with configured timeouts?

network/ocsp.go
```diff
@@ -0,0 +1,334 @@
...
+		return ErrOCSPNoConfirms
+	}
+	return nil
+}
+
+// Verify ensures certificate is not revoked by querying configured OCSP servers
+func (v DefaultOCSPVerifier) Verify(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+	for _, chain := range verifiedChains {
+		if len(chain) == 1 {
+			// This one cert[0] must be trusted since it was allowed by more basic verifying routines.
```
Lagovas, line 312: hm, why it's trusted? it verified by expiration and signature, but where it checked for revocation status? not always clients sends own certificate with issuer. common case when only own certificate and it should be checked by CA cert which previously was distributed

tests/test.py
```diff
@@ -1036,6 +1066,59 @@ def fork_connector(self, connector_port: int, acraserver_port: int,
...
+        print('openssl ocsp args: {}'.format(' '.join(cli_args)))
+
+        process = self.fork(lambda: subprocess.Popen(['openssl', 'ocsp'] + cli_args))
+
+        check_cmd = f"openssl ocsp -CAfile {TEST_TLS_CA} -issuer {TEST_TLS_CA} -cert {TEST_TLS_CLIENT_CERT} -url {ocsp_server_connection}"
+
+        if check_connection:
+            print('check OCSP server connection {}'.format(ocsp_server_connection))
+            try:
+                wait_command_success(check_cmd)
```
Lagovas, line 1093: cool

tests/test.py
```diff
@@ -107,6 +111,8 @@
 NoClientCert, RequestClientCert, RequireAnyClientCert, VerifyClientCertIfGiven, RequireAndVerifyClientCert = range(5)
 if TEST_WITH_TLS:
     ACRA_TLS_AUTH = RequireAndVerifyClientCert  # verify if provided https://golang.org/pkg/crypto/tls/#ClientAuthType
+    # Re-generate CRL so it is up-to-date when used in tests
+    os.system("SKIP_EXISTING=1 ./tests/ssl/generate_tls_keys.sh")
```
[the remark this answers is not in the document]
iamnotacake: By default, `tests/ssl/generate_tls_keys.sh` would re-generate all certs (and a CRL that contains revoked ones). So with `SKIP_EXISTING=1` we only ask it to create a fresh CRL and don't touch certs.
I've thought about creating a separate script only for generating that CRL.

tests/test.py
```diff
@@ -389,6 +395,18 @@ def wait_connection(port, count=1000, sleep=0.001):
...
+    if exit code != 0 then sleep on and try again (<count> times)
+    if <count> times is failed than raise Exception
+    """
+    while count:
+        ret = os.system(command)
+        if ret == 0:
+            return
+        count -= 1
+        time.sleep(sleep)
+    raise Exception("can't wait command success")
```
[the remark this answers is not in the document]
iamnotacake: Will do. Though, after I discovered `-ignore_err` option for `openssl ocsp` server, the problem with "invalid request = server dead" disappeared and we can now switch to simple check "try open socket" like in other places (i.e. CRL HTTP server), also getting rid of this function

.circleci/check_gotest.sh
```diff
@@ -5,6 +5,9 @@
 # cd to code with go.mod file outside of GOPATH to work with module-based behaviour
 # https://github.com/golang/go/wiki/Modules#when-do-i-get-old-behavior-vs-new-module-based-behavior
 
+# Re-generate CRL so it is up-to-date when used in tests
+SKIP_EXISTING=1 ./tests/ssl/generate_tls_keys.sh
```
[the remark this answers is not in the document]
iamnotacake: This way it only creates fresh CRL
(answered above)

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+	crlCacheTimeMax = 300
+)
+
+// CRLConfig contains configuration related to certificate validation using CRL
+type CRLConfig struct {
+	uri       string
+	fromCert  int // crlFromCert*
+	cacheSize int
+	cacheTime time.Duration
+}
```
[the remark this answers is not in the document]
iamnotacake: Making them public is... kinda unsafe, I think, since not all possible values are valid. At least, in `NewCRLConfig()` we do the validation.

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+
+	if cacheTime < 0 || cacheTime > crlCacheTimeMax {
+		return nil, ErrInvalidConfigCRLCacheTime
+	}
+
+	if uri != "" {
+		// Since this is CRL configuration alone, we don't have access to cache yet;
+		// so let's just download the CRL and forget about it
+		crlClient := NewDefaultCRLClient()
+		_, err := crlClient.Fetch(uri)
```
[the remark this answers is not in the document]
iamnotacake: Makes sense. What I thought was: "let's first validate flag values, to make sure the config contains only valid ones, and then, perform more complex check"

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+type DefaultCRLVerifier struct {
+	Config CRLConfig
+	Client CRLClient
+	Cache  CRLCache
+}
+
+// Tries to find cached CRL, fetches using v.Client if not found, checks the signature of CRL using issuerCert
+func (v DefaultCRLVerifier) getCachedOrFetch(uri string, issuerCert *x509.Certificate) (*pkix.CertificateList, error) {
+	// Try v.Cache first, but only if caching is enabled (cache time > 0)
+	if v.Config.cacheTime > 0 {
```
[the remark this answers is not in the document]
iamnotacake: Currently, cache size is just passed to LRU cache constructor, and `0` there means "unlimited capacity" rather than "no capacity". I guess, we should disallow `0` and tell users "disable caching by setting time to `0`"? Or, handle zero cache size as "cache disabler", but then there's s question about what value should cache time have in this case.

tests/test.py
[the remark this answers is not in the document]
ilammy: I mean, I don't understand why tests need to regenerate CRL. Isn't it already pregenerated and checked into the repository along with the certificates?

network/crl.go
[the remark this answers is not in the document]
ilammy: You move Muhammad to the mountain instead 😉 That is, first validate flag values offline, only then do `url.Parse()`.

network/crl.go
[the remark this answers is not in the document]
ilammy: >  I guess, we should disallow 0 and tell users "disable caching by setting time to 0"? Or, handle zero cache size as "cache disabler", but then there's s question about what value should cache time have in this case.

I guess both work and are predictable enough. If either cache size or cache duration is zero then caching is disabled. If it is disabled due to one value being zero then the other value is more or less irrelevant, as long as it can be parsed, I guess.

tests/test.py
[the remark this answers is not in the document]
iamnotacake: The reason we need a "fresh" CRL because it contains date/time (called [Next Update](https://tools.ietf.org/html/rfc5280#section-5.1.2.5)) after which it is considered outdated, thus using such outdated CRL in tests leads to certificate denial and different errors, sometimes in unexpected places.
In `getCachedOrFetch()` there's a check `if crl.TBSCertList.NextUpdate.Before(time.Now()) { /* return error */ }` for this case.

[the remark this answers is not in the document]
ilammy: Oh... I did not know about that. Thank you for explanation!

I'm so obsessed with this because I like when the tests are more or less immutable and reproducible. Regenerating a part of them during CI runs might cause accidental breakage. Admittedly, certificates themselves might expire and thus stop working, but that is... um... circumvented by issuing them to be valid for the next 50 years.

I've looked around the `openssl ca` command line and it seems that `-crldays N` option allows to set the "next update" date for CRLs as well. I guess if this is set to 50 years in the future as well then it will not be necessary to update the CRL anytime sooner than the certificates it was generated from.

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+		if err != nil {
+			log.WithError(err).Debugf("CRL: Cannot get CRL from '%s'", v.Config.uri)
+			return err
+		}
+
+		for _, revokedCertificate := range crl.TBSCertList.RevokedCertificates {
+			if revokedCertificate.SerialNumber.Cmp(cert.SerialNumber) == 0 {
+				log.Warnf("CRL: Certificate %v was revoked at %v", cert.SerialNumber, revokedCertificate.RevocationTime)
+				return ErrCertWasRevoked
+			}
```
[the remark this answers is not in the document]
iamnotacake: Seems like [there is no `Equal()`](https://pkg.go.dev/math/big#Int), only [`Cmp()`](https://pkg.go.dev/math/big#Int.Cmp)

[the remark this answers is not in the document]
ilammy: I had [this `Equal()`](https://golang.org/pkg/crypto/x509/#Certificate.Equal) in mind 🤔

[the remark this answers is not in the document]
iamnotacake: Understood. The problem is, we are not comparing two `x509.Certificate`, but rather [`pkix.RevokedCertificate`](https://pkg.go.dev/crypto/x509/pkix#RevokedCertificate) and [`x509.Certificate`](https://golang.org/pkg/crypto/x509/#Certificate). CRLs do not contain whole certificates, but rather serial number and (optionally) some extensions. So I'll leave `Cmp()` as is. Also, seems like I missed that `Extensions` handling, while [RFC tells](https://tools.ietf.org/html/rfc5280#section-4.2) that we must process it, and deny certificate in case of any _critical_ extension cannot be processed (i.e. we dunno how to handle it).

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+
+// Verify ensures configured CRLs do not contain certificate from passed chain
+func (v DefaultCRLVerifier) Verify(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+	for _, chain := range verifiedChains {
+		if len(chain) == 1 {
+			// This one cert[0] must be trusted since it was allowed by more basic verifying routines.
+			// If we are at this point, we have nothing to do, and no CA means no OCSP.
+			log.Debugln("OCSP: Certificate chain contains one certificate, nothing to do")
+			return nil
+		}
```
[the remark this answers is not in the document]
iamnotacake: I guess we better invite @storojs72 to discuss this question.
My thought was that if there is only cert in chain, it must be in list of trusted certificates (like the list where root certs are stored)
```go
	// NewTLSConfig() in network/tls_wrapper.go
	return &tls.Config{
		RootCAs:               roots,
		ClientCAs:             roots,
```
otherwise it will be rejected even before OCSP/CRL verification is performed. See [TLS config docs](https://pkg.go.dev/crypto/tls#Config)
```
	// VerifyPeerCertificate, if not nil, is called after normal
	// certificate verification by either a TLS client or server. It
	// receives the raw ASN.1 certificates provided by the peer and also
	// any verified chains that normal processing found. (...)

	// If normal verification fails then the handshake will abort before
	// considering this callback. (...)
```
Simply rejecting single-cert-chains may also be a solution because why would one use such configuration for database encryption?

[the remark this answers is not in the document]
shadinua: > My thought was that if there is only cert in chain, it must be in list of trusted certificates (like the list where root certs are stored) ... otherwise it will be rejected even before OCSP/CRL verification is performed.
Generally I agree with such approach. But, it seems to me, it's worth considering the situation when the same certificate is also provided in `--tls_ca` argument. In such case, I think, we should also trust the transmitted certificate.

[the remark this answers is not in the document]
iamnotacake: > certificate is also provided in `--tls_ca`

But can we use CA certificate for data encryption? They are often (always?) made only to sign other certificates?
Or maybe that "basic TLS verification" will reject such certificates before attempting to check OCSP/CRL?

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+
+		if httpResponse.StatusCode != 200 {
+			log.WithField("status", httpResponse.Status).Warnln("Server returned non-OK status")
+			return nil, ErrHTTPServerReturnedError
+		}
+
+		return content, nil
+	} else if strings.HasPrefix(uri, "file://") {
+		content, err := ioutil.ReadFile(uri[7:])
```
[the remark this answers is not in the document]
iamnotacake: Done. Also, replaced (but not pushed yet, just noticed) redundant `url.Parse()` inside `http` case just to extract `Host`.
Maybe I'll make `Fetch()` take `url.Url` instead of `string` though it does not make a lot of sense since `http.NewRequest()` call inside `case http` takes `string` and converts it to `Url` inside (so we won't be able to avoid one more conversion)

network/crl_test.go
iamnotacake: When I was writing these tests, there were some issues (like client was unable to connect so I suspected server didn't have time to start), but now I don't observe this issue, so removed these lines. If the problem will appear again, we'll discuss it.

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+	ErrHTTPServerReturnedError     = errors.New("Server returned non-OK status")
+	ErrFetchCRLUnsupportedProtocol = errors.New("Cannot fetch CRL, unsupported protocol")
+	ErrCacheKeyNotFound            = errors.New("Cannot find cached CRL with given URI")
+	ErrOutdatedCRL                 = errors.New("Cannot find cached CRL with given URI")
+)
+
+// --tls_crl_from_cert=<use|ignore>
+const (
+	// If certificate contains CRL distribution point(s), use them
+	crlFromCertUseStr = "use"
```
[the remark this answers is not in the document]
iamnotacake: It doesn't seem like we are using constants to generate CLI flags description (from what I see in server and connector).
Isn't it better to agree about flag names, values and description and just hardcode possible values in place where many `flag.{String|Int|Bool}` are created?
If we want to use these constants in tests, they are already accessible (because of the same mod).

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+// NewCRLConfig creates new CRLConfig
+func NewCRLConfig(uri, fromCert string, cacheSize, cacheTime int) (*CRLConfig, error) {
+	if uri != "" {
+		_, err := url.Parse(uri)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	fromCertValValues := map[string]int{
```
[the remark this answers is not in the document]
iamnotacake: Moved to global scope, prefixed with `crl`

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+		crlClient := NewDefaultCRLClient()
+		_, err := crlClient.Fetch(uri)
+		if err != nil {
+			log.WithError(err).WithField("uri", uri).Warnln("CRL: Cannot fetch configured URI")
+			// TODO return error after issues with failing tests are fixed;
+			//      CRL HTTP server is starting, connection is checked, then Acra is starting
+			//      but somehow checking connection *here* fails thus failing the tests;
+			//      everything else seems working since real requests to configured server
+			//      are successful (when CRL verification is performed)
+			// return nil, errors.New("CRL: Cannot fetch configured URI")
```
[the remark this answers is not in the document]
iamnotacake: This is not like I wanna push this exact code to `master`, more like to show in review why there was no `return error`

network/crl.go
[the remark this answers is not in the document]
Lagovas: they will not accessible from `package main` packages which are `acra-connector`/`acra-server`. Or we should declare them in sub-packages

network/ocsp.go
```diff
@@ -0,0 +1,334 @@
...
+			// If we are at this point, we have nothing to do, and no CA means no OCSP.
+			log.Debugln("OCSP: Certificate chain contains one certificate, nothing to do")
+			return nil
+		}
```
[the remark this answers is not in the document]
storojs72: I'm sorry for joining this discussion a bit late. 

It seems to me that (following best practices) chain should contain at least two certificates - one is self-signed CA certificate and the second - is the certificate (signed by this CA) issued to the entity that we verify. We can trust the entity's certificate only if we trust to this CA. The situation when only one self-signed certificate (1 certificate in chain) is provided while handshake, is technically correct - it means that we trust to this entity directly (we can definitely run TLS handshake protocol). But such authentication is quite rare case. It means that we already have knowledge of entity's public key (e.g. during some prior installation phase). Alternatively, such authentication has no sense at al. In context of CRL/OCSP it means that we definitely can skip CRL/OCSP verification, since such entity is a Certificate Authority itself. And it will unlikely decide to revoke its own credential to TLS communication from itself.

network/crl.go
```diff
@@ -0,0 +1,389 @@
...
+			queriedCRLs[v.Config.url] = struct{}{}
+			log.Debugln("CRL: OK, not found in list of revoked certificates")
+		}
+	}
+
+	return nil
+}
+
+// Verify ensures configured CRLs do not contain certificate from passed chain
+func (v DefaultCRLVerifier) Verify(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
```
storojs72, line 370: I do not see `rawCerts` using in this function. Do you need it at al?

network/ocsp.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	if len(serversToCheck) > 0 && confirms == 0 {
+		return ErrOCSPNoConfirms
+	}
+	return nil
+}
+
+// Verify ensures certificate is not revoked by querying configured OCSP servers
+func (v DefaultOCSPVerifier) Verify(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
```
storojs72, line 319: Similarly to CRL, I do not see how `rawCerts` is used in this function. May be we can remote it from input parameters?

storojs72 (commented):
I like the progress in this direction. A lot of work is done here. But some implementation specifics should be addressed. And probably splitting this PR into logical sub-components would be good idea. For example one can implement CRL at first, make PR, collect a feedback, achieve it's approval and merge it in order to commit the work to some stage. Than next PR that deal with OCSP. Finally integrate this functionality to our services (one PR for AcraServer, one PR for AcraTranslator after deciding on how flags should be named, etc.). Splitting to little components and incremental adding new changes allows to reduce time on different agreements.

network/ocsp.go
iamnotacake: Yes, with current implementation we don't use `rawCerts`. I left it for possible future use. Here (as well as in `CertVerifier` interface) just to match signature of `VerifyPeerCertificate` callback used in TLS config to call our verifiers.
Do you think we better remove it unless it is really used?

Lagovas: @storojs72 , previously @iamnotacake removed it. but we decided to leave it to be compatible with tls standard library in golang. plus it allows using some custom extensions in tls certificates which will not be parsed and used by the standard library. probably it is why golang's tls library pass parsed and raw certificates - https://golang.org/pkg/crypto/tls/#Config

cmd/acra-connector/acra-connector.go
iamnotacake: `How long to keep CRLs cached, in seconds (use 0 to disable caching, maximum: 300 s)`
Updated description with this text string ^
Set default cache time to `0`

cmd/acra-connector/acra-connector.go
iamnotacake: updated this one with `How to treat certificates unknown to OCSP: <denyUnknown|allowUnknown|requireGood>`,
and renamed `all` -> `requireGood`

tests/test.py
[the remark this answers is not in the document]
iamnotacake: Done, now exception string will contain failed command

cmd/acra-connector/acra-connector.go
```diff
@@ -229,6 +229,14 @@ func main() {
...
+	tlsOcspCheckWholeChain := flag.Bool("tls_ocsp_check_whole_chain", false, "Whether to check whole certificate chain, false = only end certificate")
+	tlsCrlURL := flag.String("tls_crl_url", "", "CRL URL")
```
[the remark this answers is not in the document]
iamnotacake: Replaced all `CRL URL` -> `URL of the Certificate Revocation List (CRL) to use`,
though missed `How should we treat CRL URL described in certificate itself`, will change that too

cmd/acra-connector/acra-connector.go
```diff
@@ -229,6 +229,14 @@ func main() {
...
+	tlsCrlFromCert := flag.String("tls_crl_from_cert", "use", "How should we treat CRL URL described in certificate itself")
```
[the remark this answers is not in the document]
iamnotacake: Done

cmd/acra-connector/acra-connector.go
```diff
@@ -229,6 +229,14 @@ func main() {
...
+	tlsCrlCacheSize := flag.Int("tls_crl_cache_size", 16, "Size of in-memory LRU cache for storing fetched CRLs, 0 = unlimited")
```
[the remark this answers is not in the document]
iamnotacake: Thanks for suggestion. Done, but (currently) in a different way: `How many CRLs to cache in memory (use 0 to disable caching)`

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/golang/groupcache/lru"
+)
+
+// Errors returned by CRL verifier
+var (
+	ErrInvalidConfigCRLFromCert    = errors.New("Invalid `tls_crl_from_cert` value")
```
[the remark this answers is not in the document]
iamnotacake: Done

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+		if err != nil {
+			return nil, err
+		}
+		defer httpResponse.Body.Close()
+		content, err := ioutil.ReadAll(httpResponse.Body)
+		if err != nil {
+			return nil, err
+		}
+
+		if httpResponse.StatusCode != 200 {
```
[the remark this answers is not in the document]
iamnotacake: Switched to `http.StatusOK`

network/crl.go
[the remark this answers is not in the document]
iamnotacake: So, created
```go
func (c *CRLConfig) isCachingEnabled() bool {
	return c.cacheTime > 0 && c.cacheSize > 0
}
```
and using it here instead `if v.Config.cacheTime > 0`

cmd/acra-connector/acra-connector.go
vixentael: agree w/ @ilammy

cmd/acra-connector/acra-connector.go
```diff
@@ -397,7 +405,27 @@ func main() {
 	if connectorMode == connector_mode.AcraServerMode {
 		if *useTLS {
 			log.Infof("Selecting transport: use TLS transport wrapper")
-			tlsConfig, err := network.NewTLSConfig(network.SNIOrHostname(*tlsAcraserverSNI, *acraServerHost), *tlsCA, *tlsKey, *tlsCert, tls.ClientAuthType(*tlsAuthType))
+
+			ocspConfig, err := network.NewOCSPConfig(*tlsOcspURL, *tlsOcspRequired, *tlsOcspFromCert, *tlsOcspCheckWholeChain)
+			if err != nil {
+				log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorWrongConfiguration).
+					Errorln("Configuration error: invalid OCSP config")
+				os.Exit(1)
```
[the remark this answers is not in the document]
vixentael: I think it would interfere with EE version

cmd/acra-connector/acra-connector.go
```diff
@@ -397,7 +405,27 @@ func main() {
...
+			crlConfig, err := network.NewCRLConfig(*tlsCrlURL, *tlsCrlFromCert, *tlsCrlCacheSize, *tlsCrlCacheTime)
+			if err != nil {
+				log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorWrongConfiguration).
+					Errorln("Configuration error: invalid CRL config")
+				os.Exit(1)
+			}
+
+			certVerifier, err := network.NewCertVerifierFromConfigs(ocspConfig, crlConfig)
+			if err != nil {
+				log.WithError(err).Fatalln("Cannot create client certificate verifier")
```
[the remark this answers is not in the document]
vixentael: as @ilammy mentioned above, doesn't `osExit(1)` interfere with audit logs in EE?

network/crl_test.go
```diff
@@ -324,27 +321,27 @@ func TestDefaultCRLClientHTTP(t *testing.T) {
...
+	// Test with invalid URL (i.e. when server returns 404 not found or any other error)
 	//
-	uri = fmt.Sprintf("http://%s/wrong_filename.pem", addr)
+	url = fmt.Sprintf("http://%s/wrong_filename.pem", addr)
 
-	fetchedCRL, err = crlClient.Fetch(uri)
+	fetchedCRL, err = crlClient.Fetch(url)
 	if err == nil {
-		t.Fatal("Unexpected success during reading CRL from wrong URI\n")
+		t.Fatal("Unexpected success durlng reading CRL from wrong URL\n")
```
ilammy: You went a bit overboard with `s/uri/url/g` here and in some other log lines.

P.S. There's "word boundary" match, as in`/\buri\b/`.

cmd/acra-connector/acra-connector.go
```diff
@@ -230,13 +230,14 @@ func main() {
 	tlsAcraserverSNI := flag.String("tls_acraserver_sni", "", "Expected Server Name (SNI) from AcraServer")
 	tlsAuthType := flag.Int("tls_auth", int(tls.RequireAndVerifyClientCert), "Set authentication mode that will be used in TLS connection with AcraServer/AcraTranslator. Values in range 0-4 that set auth type (https://golang.org/pkg/crypto/tls/#ClientAuthType). Default is tls.RequireAndVerifyClientCert")
 	tlsOcspURL := flag.String("tls_ocsp_url", "", "OCSP service URL")
-	tlsOcspRequired := flag.String("tls_ocsp_required", "denyUnknown", "Whether we need OCSP response in order to accept certificate")
-	tlsOcspFromCert := flag.String("tls_ocsp_from_cert", "prefer", "How should we treat OCSP server described in certificate itself")
+	tlsOcspRequired := flag.String("tls_ocsp_required", "denyUnknown", "How to treat certificates unknown to OCSP: <denyUnknown|allowUnknown|requireGood>")
+	tlsOcspFromCert := flag.String("tls_ocsp_from_cert", "prefer", "How should we treat OCSP server described in certificate itself; <use|trust|prefer|ignore>")
```
ilammy, line 234: I believe this one should be a full colon `:`, not semicolon `;`

Ditto in AcraServer.

network/ocsp.go
```diff
@@ -30,37 +30,37 @@ import (
...
 	}
 )
 
 const (
-	ocspRequiredYes int = iota
-	ocspRequiredNo
-	ocspRequiredAll
+	ocspRequiredDenyUnknown int = iota
+	ocspRequiredAllowUnknown
+	ocspRequiredReuireGood
```
ilammy, line 63: Those names come from a department of redundancy department. (And have a "q" missing.)

ilammy (commented):
Well, I definitely like the direction where this PR is going 👍

network/ocsp.go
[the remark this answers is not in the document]
ilammy: Thanks @storojs72 for explaining this. Now that you've framed it like this:

> In context of CRL/OCSP it means that we definitely can skip CRL/OCSP verification, since such entity is a Certificate Authority itself. And it will unlikely decide to revoke its own credential to TLS communication from itself.

I believe that it would be okay to leave a small comment describing this and a log line to inform the users that Acra has skipped redundant CRL/OCSP verification for an ultimately trusted certificate.

network/crl.go
```diff
@@ -70,6 +71,8 @@ const (
 )
 
 const (
+	// Max int value, because cache implementation uses `int` for setting capacity
+	crlCacheSizeMax = 2147483647
```
ilammy, line 75: Uh...

1. If we are going to use `int` why read it as `uint` in the first place? And then bother with conversions here and there, sometimes even not checking ranges.
2. The spec says that `int` might be either 32 or 64-bit. Reliable way to get its maximum value is something like `int(^uint(0) >> 1)`.

If you'd like to avoid negative values, I believe that `size >= 0` is much simpler to write than `size <= crlCacheSizeMax` which you still have to do.

network/crl.go
```diff
@@ -41,7 +42,7 @@ var (
 	ErrUnknownCRLExtensionOID       = errors.New("unable to process unknown critical extension inside CRL")
 )
 
-// --tls_crl_from_cert=<use|ignore>
+// --tls_crl_from_cert=<use|truet|prefer|ignore>
```
ilammy: There's a typo:
```
truet
  ~~^~~
```

network/crl.go
iamnotacake: Making the flag itself `Uint` + converting to `int` is more obvious, it will detect error earlier and negative cache size does not make sense anyway. That's why I thought it was better than `Int` + validation it's `> 0`. Also, even max value of 32-bit int is overkill for cache size.

network/ocsp.go
iamnotacake: Renamed `ocspRequiredReuireGood` -> just `ocspRequiredGood`

network/ocsp.go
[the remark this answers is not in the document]
iamnotacake: Left the same comment, changed logging code to
```go
log.WithField("serial", chain[0].SerialNumber).WithField("subject", chain[0].Subject).
	Infoln("OCSP: Certificate chain consists of one already trusted certificate, nothing to do")
```
done the same thing to CRL, thought it would be useful to see in logs which certificate is used in this weird way

network/tls_wrapper.go
iamnotacake: This function `RegisterTLSBaseArgs()` (as well as `NewTLSConfigFromBaseArgs()`) is used in Acra EE, in acra-translator.
It makes sense there, no need to extend acra-translator's `main()`.
In Acra CE server & connector these functions are not called at all, so these two apps create TLS config in a different way.

network/crl.go
Lagovas: https://golang.org/pkg/math/#pkg-constants - math.MaxInt8|16|32|64

network/crl.go
```diff
@@ -199,6 +202,10 @@ func (c DefaultCRLClient) Fetch(url string) ([]byte, error) {
 
 		return content, nil
 	case "file":
+		if !allowLocal {
+			return nil, ErrFetchDeniedForLocalURL
+		}
```
ilammy: There should be a test for this.

network/crl.go
ilammy: Well, `math` constants exist but you still don't know whether `int` is `int32` or `int64`.

FWIW, we can say
```go
const crlCacheSizeMax = 1_000_000
```
or any other 'reasonable' number, for a certain definition thereof, as long as `crlCacheSizeMax <= math.MaxInt32`.

What I don't really want is to trade a single `if` in option/config parsing code for a whole bunch of range checks and random type casts after that. For example, slice length in Go is signed `int`. While arguably it should have been unsigned for the sake of domain correctness, the world didn't burn down because of that.

network/ocsp.go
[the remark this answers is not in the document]
iamnotacake: After last discussion, made this log message a warning, removed `subject` field and changed the message text. Everything else seems to be good enough.

network/crl.go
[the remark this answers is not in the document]
iamnotacake: These strings constants, possible values for `--tls_*_from_cert` and `--tls_ocsp_required` are now public.
`flag` description will be generated based on them.

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+
+// DefaultCRLClient is a default implementation of CRLClient
+// (as opposed to stub ones used in tests)
+type DefaultCRLClient struct {
+	httpClient *http.Client
+}
+
+// NewDefaultCRLClient creates new DefaultCRLClient
+func NewDefaultCRLClient() DefaultCRLClient {
+	return DefaultCRLClient{httpClient: &http.Client{}}
```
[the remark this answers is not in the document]
iamnotacake: Yeah, configured default timeouts for OCSPClient (15s) and CRLClient (20s)

network/crl_test.go
iamnotacake: Replaced string literals with constants in places like this

network/ocsp.go
iamnotacake: > them public and use in descriptions and tests instead hardcoded values

yes, done just that, thanks

network/ocsp.go
iamnotacake: Done, now `http.Client` is created and stored in OCSPClient, with configured timeout of 15s

network/crl.go
iamnotacake: Done, added one small test to ensure `Fetch("file://...", allowLocal = false)` returns `ErrFetchDeniedForLocalURL`

cmd/acra-connector/acra-connector.go
iamnotacake: Description now includes possible values (for tls_ocsp_required, tls_ocsp_from_cert, tls_crl_from_cert)

network/crl.go
iamnotacake: Thanks. Fixed

network/crl_test.go
iamnotacake: Didn't notice my regex replaced too much. Thanks, fixed those typos

network/crl.go
iamnotacake: As explained [here](https://github.com/cossacklabs/acra/pull/436#discussion_r539399797), we don't need it, left it for compatibility and future extensions

network/crl_test.go
```diff
@@ -0,0 +1,468 @@
...
+		ocspCert:              TestOCSPCertFilename,
+		ocspKey:               TestOCSPKeyFilename,
+		validRawCerts:         validRawCerts,
+		validVerifiedChains:   validVerifiedChains,
+		invalidRawCerts:       invalidRawCerts,
+		invalidVerifiedChains: invalidVerifiedChains,
+	}
+}
+
+// func getTestCertGroup3(t *testing.T) TestCertGroup {
```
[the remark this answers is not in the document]
iamnotacake: As the certificate chain with intermediate CA is added again, these lines are now uncommented

network/crl_test.go
iamnotacake: As the certificate chain with intermediate CA is added again, these lines are now uncommented

network/crl_test.go
```diff
@@ -0,0 +1,536 @@
...
+	expectOk("", CrlFromCertUseStr, true, 0, 1)
+	expectOk("", CrlFromCertUseStr, false, 1, 300)
+	expectOk("", CrlFromCertTrustStr, true, 1, 0)
+	expectOk("", CrlFromCertPreferStr, false, 1, 0)
+
+	expectErr("htt://invalid url", CrlFromCertUseStr, false, 1, 5)
+	expectErr("", "IgNoRe", false, 1, 5)
+	expectErr("", CrlFromCertUseStr, false, 2147483648, 5)
+	expectErr("", CrlFromCertUseStr, false, 1, 301)
+	expectErr("", CrlFromCertUseStr, false, 1, 9000)
```
ilammy, line 311: I'd add a comment at the end of each line with the reason why these values should be rejected.

network/crl.go
```diff
@@ -0,0 +1,503 @@
...
+				return ErrEmptyCertChain
+			}
+		}
+
+		if len(chain) == 1 {
+			// This one cert[0] must be trusted since it was allowed by more basic verifying routines.
+			// If we are at this point, we have nothing to do, and no CA means no CRL.
+			log.WithField("serial", chain[0].SerialNumber).
+				Warnln("CRL: Certificate chain consists of one already trusted certificate, nothing to do, it is recommended to use non-root certificates for TLS handshake")
+			return nil
```
ilammy, line 481: Similarly, do we still need to improve wording here?

network/ocsp.go
```diff
@@ -0,0 +1,384 @@
...
+			default: // tls.VerifyClientCertIfGiven, tls.RequireAndVerifyClientCert
+				return ErrEmptyCertChain
+			}
+		}
+
+		if len(chain) == 1 {
+			// This one cert[0] must be trusted since it was allowed by more basic verifying routines.
+			// If we are at this point, we have nothing to do, and no CA means no OCSP.
+			log.WithField("serial", chain[0].SerialNumber).
+				Warnln("OCSP: Certificate chain consists of one already trusted certificate, nothing to do, it is recommended to use non-root certificates for TLS handshake")
```
ilammy, line 361: Do we still need to improve wording here?

ilammy (approved):
Alrighty. Given that the build is green and my own nitpicks were mostly addressed, I'd be the first one to ja-jang a green thingie here! ✅  ✅  ✅

I also took the liberty to mark some comments made by other people as resolved, given that they were long resolved, in my opinion.

@Lagovas still has some grudges about implicit trust to root CA certificates [here](https://github.com/cossacklabs/acra/pull/436/files#r536944117), which I marked with my comments too.

I have a feeling that @iamnotacake has convinced me that if the user specifies a self-signed certificate then we can just trust it *per se* and avoid making OCSP checks. In that case I believe it's okay to just skip it with a log entry. (But maybe be more verbose in comments, explaining why does Acra trusts it: because it's a self-signed root certificate which is unlikely to point to an OCSP server that marks it as revoked, or an OCSP server configured from the command line would deem a self-signed certificate revoked.

Another part of me feels that this special case is a wart. If the configuration and/or certificate include an OCSP server, and if the configuration says that we should consult OCSP server, then we should consult it. Even if that's just one certificate. It's not that expensive, done only once, and we're not penalizing the users here too much, IMO. What if the users want to run their own OCSP and blacklist some known self-signed certificates because they know that a part of their infrastructure has been compromised, they rotated the certs, but are not sure that this has been rolled out everywhere?

You can s/OCSP/CRL/g in the above text, the point stands. This is one thing that I would like to see resolved somehow and documented in a top-level comment before this PR gets merged. If anything, I'd trust @storojs72 to be an arbiter here: should we trust self-signed certificates and skip OCSP, or should we still do the queries because we can? (Because the RFC does not seem to say anything about this.)

cmd/acra-connector/acra-connector.go
```diff
@@ -229,6 +229,19 @@ func main() {
...
+	tlsOcspFromCert := flag.String("tls_ocsp_from_cert", network.OcspFromCertPreferStr,
+		fmt.Sprintf("How to treat OCSP server described in certificate itself: <%s>", strings.Join(network.OcspFromCertValuesList, "|")))
+	tlsOcspCheckOnlyLeafCertificate := flag.Bool("tls_ocsp_check_only_leaf_certificate", false, "Put 'true' to check only final/last certificate, or 'false' to check the whole certificate chain using OCSP")
+	tlsCrlURL := flag.String("tls_crl_url", "", "URL of the Certificate Revocation List (CRL) to use")
+	tlsCrlFromCert := flag.String("tls_crl_from_cert", network.CrlFromCertPreferStr,
+		fmt.Sprintf("How to treat CRL URL described in certificate itself: <%s>", strings.Join(network.CrlFromCertValuesList, "|")))
+	tlsCrlCheckOnlyLeafCertificate := flag.Bool("tls_crl_check_only_leaf_certificate", false, "Put 'true' to check only final/last certificate, or 'false' to check the whole certificate chain using CRL")
+	tlsCrlCacheSize := flag.Uint("tls_crl_cache_size", 16, "How many CRLs to cache in memory (use 0 to disable caching)")
+	tlsCrlCacheTime := flag.Uint("tls_crl_cache_time", 0,
```
Lagovas, line 243: if 0 means no expiration then better to use some named constant for that. That how we did it for keystore cache - https://github.com/cossacklabs/acra/blob/master/keystore/cache.go#L21

cmd/acra-server/acra-server.go
```diff
@@ -127,6 +127,23 @@ func main() {
...
+	tlsOcspCheckOnlyLeafCertificate := flag.Bool("tls_ocsp_check_only_leaf_certificate", false, "Put 'true' to check only final/last certificate, or 'false' to check the whole certificate chain using OCSP")
+	tlsCrlURL := flag.String("tls_crl_url", "", "URL of the Certificate Revocation List (CRL) to use")
+	tlsCrlClientURL := flag.String("tls_crl_client_url", "", "URL of the Certificate Revocation List (CRL) to use, for client/connector certificates only")
+	tlsCrlDbURL := flag.String("tls_crl_database_url", "", "URL of the Certificate Revocation List (CRL) to use, for database certificates only")
+	tlsCrlFromCert := flag.String("tls_crl_from_cert", network.CrlFromCertPreferStr,
+		fmt.Sprintf("How to treat CRL URL described in certificate itself: <%s>", strings.Join(network.CrlFromCertValuesList, "|")))
+	tlsCrlCheckOnlyLeafCertificate := flag.Bool("tls_crl_check_only_leaf_certificate", false, "Put 'true' to check only final/last certificate, or 'false' to check the whole certificate chain using CRL")
+	tlsCrlCacheSize := flag.Uint("tls_crl_cache_size", 16, "How many CRLs to cache in memory (use 0 to disable caching)")
+	tlsCrlCacheTime := flag.Uint("tls_crl_cache_time", 0,
```
Lagovas, line 145: same

network/crl.go
```diff
@@ -0,0 +1,503 @@
...
+// DefaultCRLClient is a default implementation of CRLClient
+// (as opposed to stub ones used in tests)
+type DefaultCRLClient struct {
+	httpClient *http.Client
+}
+
+// NewDefaultCRLClient creates new DefaultCRLClient
+func NewDefaultCRLClient() DefaultCRLClient {
+	return DefaultCRLClient{httpClient: &http.Client{
+		Timeout: time.Second * time.Duration(20),
```
Lagovas, line 167: better to move this value `time.Second * time.Duration(20)` to constant value. it will allow to re-use it in tests to verify that timeout works as expected and allow to avoid copy-paste. Plus if some other part of acra will want to keep in mind such timeouts, it may access to this value if need (for example to sum or aggregate all timeouts in some chain of calls)

tests/ssl/generate_tls_keys.sh
```diff
@@ -17,33 +20,106 @@ openssl req -x509 \
...
+
+    # generate CRL (Certificate Revocation List) signed by corresponding CA
+    openssl ca \
+        -gencrl \
+        -config "$config" \
+        -crldays "${EXPIRE}" \
+        -out "${OUT_DIR}/$signer/crl.pem"
+}
+
+declare -a names=("mysql" "postgresql" "acra-writer" "acra-writer-2" "acra-writer-revoked" "acra-server" "ocsp-responder" "intermediate-ca")
```
Lagovas, line 111: thanks for acra-writer-2, I missed to add)

tests/test.py
```diff
@@ -1012,6 +1031,10 @@ def get_connector_tls_params(self):
             'tls_ca': TEST_TLS_CA,
             'tls_key': TEST_TLS_CLIENT_KEY,
             'tls_cert': TEST_TLS_CLIENT_CERT,
+            'tls_ocsp_url': 'http://127.0.0.1:{}'.format(self.OCSP_SERVER_PORT),
```
Lagovas, line 1034: better to use `localhost` instead 127.0.0.1 if will be used TLS in a future and certificates will use localhost in SAN

tests/test.py
```diff
@@ -1106,6 +1188,16 @@ def get_connector_api_connection_string(self, port=None):
             port = self.CONNECTOR_API_PORT_1
         return get_connector_connection_string(port)
 
+    def get_ocsp_server_connection_string(self, port=None):
+        if not port:
+            port = self.OCSP_SERVER_PORT
+        return 'http://127.0.0.1:{}'.format(port)
```
Lagovas, line 1194: same here, better to prefer localhost instead 127.0.0.1

Lagovas (commented):
actually looks good. after final small fixes we can merge it

storojs72:
@ilammy, I believe that single ultimate CRL/OCSP cant be used for the case that you mentioned. We can perform blacklisting on per-certificate basis, which means that if some itermediate CA is compromised, we should blacklist each certificate issued by this CA. Each CA ideally should have own correspondent CRL/OCSP instance, because parent (root) CA has no clue about certificates issued by child (intermediate) CA.

I found very good explanation of recursive algorithm described in RFC 5280 (tools.ietf.org/html/rfc5280#page-71) that should be applied to verification of end-user certificate issued by intermediate CA (which has parent  root cert) on stackoverflow: https://stackoverflow.com/questions/19962069/how-does-certificate-revocation-work-with-intermediate-cas

tests/test.py
iamnotacake: Done

network/crl_test.go
iamnotacake: Done

network/crl.go
iamnotacake: Extracted HTTP client default timeouts to constants (`OcspHttpClientDefaultTimeout` and `CrlHttpClientDefaultTimeout`)

tests/test.py
iamnotacake: Done

network/crl.go
iamnotacake: Ah, took `1_000_000` as max value. Nobody would ever need such a big cache anyway (where every cache item corresponds to unique URL, so old CRLs won't take space)

network/ocsp.go
iamnotacake: As [crypto.tls.ConnectionState](https://pkg.go.dev/crypto/tls#ConnectionState) struct description says:
> // VerifiedChains is a list of one or more chains where the first element is
> // PeerCertificates[0] and the last element is from Config.RootCAs (on the
> // client side) or Config.ClientCAs (on the server side).
> // ...
> VerifiedChains [][]*x509.Certificate

and this `VerifiedChains` stored in `tls.ConnectionState` is essentially the same as we get in this function. But since we got only one certificate in chain, that means it must be one of those `Config.RootCAs (on the client side) or Config.ClientCAs (on the server side)`, that's why I consider it "trusted"

If other side sent only one certificate, not the whole chain, TLS implementation should build whole chain to verify its path to one of root certs. If it's not possible (i.e. there exists some intermediate CA, other side didn't send its cert and we also don't have it), then we wouldn't be able to get to this point at all.

Also, [crypto.tls.Config](https://pkg.go.dev/crypto/tls#Config) says about its field `VerifyPeerCertificate`:
> // ...
> // It receives the raw ASN.1 certificates provided by the peer and also
> // any verified chains that normal processing found.
> // ...
> // If normal verification fails then the handshake will abort before considering this callback.
> // If normal verification is disabled by
> // setting InsecureSkipVerify, or (for a server) when ClientAuth is
> // RequestClientCert or RequireAnyClientCert, then this callback will
> // be considered but the verifiedChains argument will always be nil

So with `normal verification` meaning "check that leaf cert is signed by CA and CA is trusted (root) or signed by another CA (and so on)", `verifiedChains` in this function will either contain really full chains ending with ultimately trusted certificates, or be empty (which also means one doesn't care about certificate validation) and OCSP/CRL verifiers will do nothing.

I hope all this text makes at least some sense %)

ilammy: > If other side sent only one certificate, not the whole chain, TLS implementation should build whole chain to verify its path to one of root certs.

But what if the whole chain consists of just one certificate? That is, the other side is using a self-signed certificate (which is still in our trusted root store). Like, you're just using a public key and all the PKI stuff is there just for compatibility with TLS.

The fact that a certificate is in trusted root store does not mean that it's valid. Expired certificates can be placed there, it does not mean that we should trust them past their validity date (or any certificates issued outside of the validity date). Similarly, a root certificate could have been 'self-revoked', with a revocation being included into CRLs, with notices distributed to OCSPs. While this does not make much sense from the PKI perspective, where CA is supposed to revoke certificates they publish, not commit sudoku. However, it seems to be *technically* possible.

Hence, even if this a rare use case – I do realize that this should not really happen in the wild, not with proper PKI – the new code here has a branch to handle this special case. That's why I'm haggling so much about the expected behavior. If we do put something in there it has to be sensible.

iamnotacake: Well, I guess you're right that root CA can generate CRL with itself being in list of revoked. But I gotta check this quirky scenario to be sure (or find this in RFC).
And speaking about OCSP, is there any way CA can respond about itself? Or its dedicated OCSP-responder certificate can tell "my CA is revoked"? `(X) Doubt`
I guess it's better to dig into RFC again than think how it could work.

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+
+// Get tries to get CRL from cache, returns error if failed
+func (c *LRUCRLCache) Get(key string) (*CRLCacheItem, error) {
+	log.Debugf("CRL: LRU cache: loading '%s'", key)
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+
+	value, ok := c.cache.Get(key)
+	if ok {
+		value, _ := value.(*CRLCacheItem)
```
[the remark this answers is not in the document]
iamnotacake: Do we really have to do this if `CRLCache` interface is strictly limited to only allow insertion of this same type?
And how can we do a compile-time check if `lru.Cache` works with `interface{}`, and not some exact type?

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+	Client CRLClient
+	Cache  CRLCache
+}
+
+// Tries to find cached CRL, fetches using v.Client if not found, checks the signature of CRL using issuerCert
+func (v DefaultCRLVerifier) getCachedOrFetch(uri string, issuerCert *x509.Certificate) (*pkix.CertificateList, error) {
+	// Try v.Cache first, but only if caching is enabled (cache time > 0)
+	if v.Config.cacheTime > 0 {
+		cacheItem, err := v.Cache.Get(v.Config.uri)
+		if cacheItem != nil {
```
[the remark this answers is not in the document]
iamnotacake: It means there was no item we needed in cache, so we just fall through and do the usual fetch-verify-storeInCache-return actions.

I've thought about returning `(T, bool)` instead of `(T, error)` in cache's methog `Get()` where `bool` is `true` if item found and `T` is some value, `false` if not found and we should not access `T`. Just like standard map does. Because there is actually no need in different error types when just getting items from cache. Exists or not, that's all. Though, this could be done in next PR that updates/extends behavior of certificate verifiers.

network/crl.go
```diff
@@ -0,0 +1,370 @@
...
+
+func (v DefaultCRLVerifier) verifyCertWithIssuer(cert, issuer *x509.Certificate) error {
+	log.Debugf("CRL: Verifying '%s'", cert.Subject.String())
+
+	if len(v.Config.uri) == 0 && v.Config.fromCert == crlFromCertIgnore {
+		log.Debugln("CRL: Skipping check since no config URI specified and we were told to ignore URIs from certificate")
+		return nil
+	}
+
+	for _, crlDistributionPoint := range cert.CRLDistributionPoints {
```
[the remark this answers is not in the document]
iamnotacake: Not-so-unnecessary since using using CRL URLs from cert may be disabled in config, but in this loop we'll still be able to see debug log with contained URLs. No more than for debugging purposes

cmd/acra-connector/acra-connector.go
iamnotacake: Done. Created const `CrlDisableCacheTime = 0`, put as default value for `tls_crl_cache_time`.
Also, `CrlDefaultCacheSize = 16` as default value for `tls_crl_cache_size`

vixentael (approved):

network/ocsp.go
Lagovas: CA may respond about itself and OCSP response will be signed by expired cert and we should deny such connections without success check. So it correctly maps to our current logic.

network/crl.go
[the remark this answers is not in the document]
Lagovas: hm, you are right. it's not necessary.

Lagovas (approved):
lgtm
don't forget to close parallel PR)

PR #437, opened 2020-12-22
Merge identifier extractor and convertor

Lagovas:
To encapsulate calls order of identifier extractor and convertor into one component, I added TLSClientIDExtractor which aggregate and encapsulate that logic and simplify re-usage in parts which don't need to wrap connections to TLS, only extract info from already established connection and TLS state (acra-translator), plus it will simplify code re-usage in acra-ee


ilammy (approved):
Sure, why not 🤷

iamnotacake (approved):
Looks good, two things that are always used together are now more unified

PR #438, opened 2020-12-23
Handle certificate validation flags in single place

iamnotacake:
Create a function `RegisterCertVerifierArgs()` which will now be responsible for handling OCSP- and CRL-related command line flags.
In acra-server and acra-connector, call it instead of describing all these flags (which are mostly duplicated).

Then, call one of
* `NewCertVerifierFromArgs()`
* `NewClientCertVerifierFromArgs()`
* `NewDatabaseCertVerifierFromArgs()`

which will use values generated by `RegisterCertVerifierArgs()` and create the CertVerifier under the hood.

All these changes will allow editing/adding/removing flags much easier since they will be in one place.
Also, adding cert verification functionality to different binaries in `cmd/` will require writing less code:
* one call to register flags,
* another one to get the verifier we can use in TLS wrapper.

### DIFF
--- a/network/cert_verifier.go
+++ b/network/cert_verifier.go
@@ -17,9 +17,13 @@ limitations under the License.
 package network
 
 import (
+	"crypto/tls"
 	"crypto/x509"
 	"errors"
+	"flag"
+	"fmt"
 	log "github.com/sirupsen/logrus"
+	"strings"
 )
 
 // Errors common for OCSP and CRL verifiers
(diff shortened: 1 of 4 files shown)


Lagovas (approved):
lgtm, but let someone else review too before merging

network/cert_verifier.go
```diff
@@ -28,16 +32,139 @@ var (
...
+	tlsCrlClientURL                 string
+	tlsCrlDbURL                     string
+	tlsCrlFromCert                  string
+	tlsCrlCheckOnlyLeafCertificate  bool
+	tlsCrlCacheSize                 uint
+	tlsCrlCacheTime                 uint
+)
+
+// RegisterCertVerifierArgs register CLI args tls_ocsp_url|tls_ocsp_client_url|tls_ocsp_database_url|tls_ocsp_required|tls_ocsp_from_cert|tls_ocsp_check_only_leaf_certificate|tls_crl_url|tls_crl_client_url|tls_crl_database_url|tls_crl_from_cert|tls_crl_check_only_leaf_certificate|tls_crl_cache_size|tls_crl_cache_time which allow to get CertVerifier by NewCertVerifierFromArgs|NewClientCertVerifierFromArgs|NewDatabaseCertVerifierFromArgs functions
+func RegisterCertVerifierArgs(separate_client_db_urls bool) {
```
ilammy, line 52: I have a point to make and I'm going to hit it every time.

It's hard to understand why AcraConnector needs
```go
network.RegisterCertVerifierArgs(false)
```
and AcraServer uses
```go
network.RegisterCertVerifierArgs(true)
```
Yes, I lead a crusade over boolean flags.

For example,
```go
func RegisterCertVerifierArgs() {
        registerCertVerifierArgs(false)
}

func RegisterCertVerifierArgsWithSeparateClientAndDatabase() {
        registerCertVerifierArgs(true)
}

func registerCertVerifierArgs(separate_client_db_urls bool) {
        // whatever we have now
}
```
makes for more readable code at the use site, despite looking like it came from the Java land.

network/cert_verifier.go
```diff
@@ -28,16 +32,139 @@ var (
...
-	// - (for OCSP) the certificate is not known by OCSP server and we requested tls_ocsp_required == "yes" or "all"
-	// - (for OCSP) if we were unable to contact OCSP server(s) but we really need the response, tls_ocsp_required == "all"
+	// - (for OCSP) the certificate is not known by OCSP server and we requested tls_ocsp_required == "denyUnknown" or "requireGood"
+	// - (for OCSP) if we were unable to contact OCSP server(s) but we really need the response, tls_ocsp_required == "requireGood"
 	Verify(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error
 }
 
+// NewCertVerifierFromArgs creates a CertVerifier based on passed OCSP and CRL command line flags.
+// Ignores `--tls_{ocsp,crl}_{client,database}_url` flags, only uses `--tls_{ocsp,crl}_url` as URL source.
+func NewCertVerifierFromArgs() (CertVerifier, error) {
```
ilammy, line 90: IMO, you can drop `FromArgs` from the names.

`NewThingFromOtherThing` constructors typically take that other thing as an argument, which is not the case here. It is possible to rewrite the code to use a dedicated struct with parsed values instead of keeping them all as global variables, then pass a reference to that struct. But it's not a complexity that it's needed right now.

network/cert_verifier.go
iamnotacake: OK, I don't think two more tiny functions will make it too complex

network/cert_verifier.go
iamnotacake: OK, will rename (this and two other `New*CertVerifierFromArgs`).
Maybe we'll have another constructor for `CertVerifier` some day, then we may need to rename constructor funcs to distinguish between them.

Lagovas:
@iamnotacake implemented @ilammy suggestions. Due to the late time for @ilammy and to avoid extra PR to merge it to acra-ee (because it will cause to update acra-ee one more time after merge), I think we can merge it now.

PR #439, opened 2020-12-23
Use absolute path for files in tests

Lagovas:
When we run acra-ce tests with binaries from acra-ee, working directory used from acra-ee environment, where not exists `tests/ssl/*` folders and files. So to avoid errors, we should rely on abs paths in acra-ce tests and calculate them relative to files in acra-ce.

p.s. also fixed some naming issues to pass linter's checks


iamnotacake (approved):
Good. And thanks for fixing golint issues

vixentael (approved):

PR #440, opened 2020-12-29
Additional unit tests for certificate verifier

iamnotacake:
Additional unit tests for better coverage of OCSP client & verifier behavior.

Small bugfix where OCSPClient did not check whether response was related to asked certificate, now it will return error if mismatched.

### DIFF
--- a/network/crl.go
+++ b/network/crl.go
@@ -42,6 +42,7 @@ var (
 	ErrCacheKeyNotFound             = errors.New("cannot find cached CRL with given URL")
 	ErrOutdatedCRL                  = errors.New("fetched CRLs NextUpdate is behind current time")
 	ErrUnknownCRLExtensionOID       = errors.New("unable to process unknown critical extension inside CRL")
+	ErrUnimplementedCRLExtension    = errors.New("handling of CRL extension is not yet implemented")
 )
 
 // --tls_crl_from_cert=<use|trust|prefer|ignore>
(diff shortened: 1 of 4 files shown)


network/crl.go
```diff
@@ -346,6 +347,35 @@ func (v DefaultCRLVerifier) getCachedOrFetch(url string, allowLocal bool, issuer
...
+		// This means we also have to handle these extensions.
+		switch extension.Id.String() {
+		case "2.5.29.35":
+			// section 5.2.1 (4.2.1.1), id-ce-authorityKeyIdentifier
+
+		case "2.5.29.18":
+			// section 5.2.2 (4.2.1.7), id-ce-issuerAltName
+
+		case "2.5.29.20":
+			// section 5.2.3, id-ce-cRLNumber
```
ilammy, line 362: Don't we need to do something here? At least a log line should be nice. Or are those extensions handled by Go's x509?

iamnotacake: > Or are those extensions handled by Go's x509?

Nope, these are not handled. All the stuff related to CRL is like "if you need it, take these simple structs from crypto/x509/pkix and implement needed functions by yourself"

> At least a log line should be nice.

okay, gonna add some logs for all these `case`s. Debug or something more "serious"?

ilammy: Thanks for clarifications.

I'm just not sure why this function is necessary. It seems to be filtering out CRLs that Acra 'does not handle' (because it does not recognize the extensions). And yet, those extensions that it does recognize don't seem to be used in any way. The only side effect here are those debug-level logs about unrecognized extensions, and a possible failure if there is some critical extension which is not listed here. And if it were listed here, the code would just do nothing.

So I wonder, what's the point here. Discourage people from putting weird extensions?

iamnotacake: The point was to create some skeleton for further extension of cert verifier using CRL. I guess you wanna way "better to not have these "empty switches" at all or have them doing what they have to do"?

ilammy: Well, yeah, something like that. Though, I did not realize that you intend them as a scaffolding for future development. I'm sorry. In that case I guess a TODO comment will be better than logs. Just to leave a message to whoever will be reading this code.

Lagovas (commented):
lgtm

shadinua (approved):

Lagovas (approved):

PR #441, opened 2021-01-06
Update RDBMS docker versions with TLS certs actualization

ZhmakAS:
update RDBMS docker version to new stable ones, regenerate tls cert with SAN for running in docker


Lagovas:
How you checked that images working with new certificates and with new dbs?

iamnotacake:
I see these new DB certificates contain OCSP & CRL URIs. Well, that's expected if you generated them with `tests/ssl/generate_tls_keys.sh`. There is no problem if they are only used in integration tests since `tests/test.py` will launch corresponding services if testing with TLS enabled.
However, keep in mind that using such certificates (with TLS enabled in Acra) when there are no services running (OCSP responder at `localhost:8888` and HTTP server being able to reply with valid CRL at `localhost:8889/crl.pem`), Acra will give certificate errors. Because it will try to validate certificates using known URIs by default. Of course, this behavior can be changed with CLI flags (or YAML configuration?).

Also, as far as I can see, you didn't regenerate root CA cert. Okay. But if you do this, remember to look at the bottom of generation script, there's one more thing to be done in such case.

ZhmakAS:
> How you checked that images working with new certificates and with new dbs?

Jus

> How you checked that images working with new certificates and with new dbs?

To check that everything works fine, I regenerated certs for corresponding services, rebuild RDBMS docker images to apply certs changes, run docker-composes files with ssl on db side and ssl between acra-connector and acra-server and verified by running golang app from example folder. In result, i managed to execute golang script successfully for both DBs. Did I miss something?

shadinua:
As @iamnotacake correctly noted, the new certificates contain both OSCP and CRL URIs, but actually we don't have services listen there.

To prevent unexpected behaviour in the future, I see 2 possible ways:
* regenerate certificates without OSCP and CRL (currently we do not use it in demos)
* explicitly adapt the demos (adding appropriate arguments) to work with current certificates, assuming that further we will build demos with OSCP and CRL services, which will require these defined URIs

The second way looks less attractive to me, because it may look a bit dirty to specify URIs which will be ignored by arguments and never used in bounds of the demos.

So far I'd suggest to regenerate certificates without OSCP and CRL records. I think it would be nice to modify the `generate_tls_keys.sh` script slightly and add an option to generate certificates without OSCP and CRL (for example, `TLS_NO_OSCPCRL` environment variable may be checked).

shadinua (approved):
Thanks, I see the regenerated certificates without OSCP and CRL. However, as far as I understand, it was done ad-hoc without improving the `generate_tls_keys.sh`. Well, probably it looks like the necessary minimum for this case.

iamnotacake (approved):
At first, I didn't notice you were changing certificates in `docker/ssl/`, not in `tests/ssl/`. And then you removed OCSP & CRL URLs which will prevent errors when Acra images will be updated and will include cert validation feature. If these changes do not affect docker images used in integration tests (those are built using `tests/docker/*.dockerfile` if I'm not wrong), I see no problems. But better to wait for someone else's review.

PR #451, opened 2021-11-10
Refactor vault integration tests

Lagovas:
- Now tests related to vault run using docker container instead of sdk's instance due to problems with cross-module dependencies between `github.com/hashicorp/vault/sdk`, `github.com/hashicorp/vault/http`, `github.com/hashicorp/vault/api`, `github.com/hashicorp/vault` packages. Plus, we already have flaky tests with this module due to background initialization problems. 
- Plus all tests start only after checking that Vault already started and accepts connections.
- Additionally, we didn't check tests that depend on redis and should run with tags. Now for these tests will be started Redis container and tests called with all required tags.
- Updated golang's versions that we check and now it is the last 3 versions.
- Removed most of indirect dependencies from `go.mod` that are not imported directly from our code.

P.S. I updated our image for tests `cossacklabs/ci-py-go-themis` due to changed go versions list. So, please pull new image for local tests (or it will do `circleci` tool automatically)

## Checklist

  with new changes


.circleci/check_redis_connection.sh
```diff
@@ -0,0 +1,21 @@
...
+
+  # empty reply status
+  if [ "$?" -eq "52" ]; then
+    exit 0
+  else
+    sleep ${DELAY}
+  fi
+done
+
+echo "Cannot reach the Redis at $TEST_TEST_VAULT_HOST"
```
ZhmakAS, line 20: I think it might be a little typo here. TEST_REDIS_HOST?

.circleci/check_vault_connection.sh
```diff
@@ -0,0 +1,23 @@
...
+    -XGET "https://${TEST_VAULT_HOST}:${TEST_VAULT_PORT}/"
+
+  if [ "$?" -eq "0" ]; then
+    exit 0
+  else
+    sleep ${DELAY}
+  fi
+done
+
+echo "Cannot reach the HashiCorp Vault at $TEST_TEST_VAULT_HOST"
```
ZhmakAS, line 22: Is it correct env variable? From the first lines of code I see it like

`TEST_VAULT_HOST` but you have `TEST_TEST_VAULT_HOST`

ZhmakAS (changes requested):

.circleci/check_redis_connection.sh
Lagovas: nice catch, fixed

ZhmakAS (approved):
Thanks!

PR #454, opened 2021-11-11
Updated keymaker behaviour

ZhmakAS:
Updated `keymaker` tool to be able to generate keys that are not using `client_id` with its empty definition.
`client_id` will be validated only for generation of keys with required `client_id`

## Checklist

- [-] Public API has proper documentation in the [Acra documentation] site or has PR on [documentation repository] 
  with new changes
- [ -] CHANGELOG.md is updated (in case of notable or breaking changes)


tests/test.py
```diff
@@ -1001,6 +1001,27 @@ def random_keys(size):
...
+                 '--generate_log_key',
+                 '--generate_acrawebconfig_keys',
+                 '--keys_public_output_dir={}'.format(folder)])
+
+        with tempfile.TemporaryDirectory() as folder:
+            with self.assertRaises(subprocess.CalledProcessError) as exc:
+                subprocess.check_output(
+                    ['./acra-keymaker', '--keystore={}'.format(KEYSTORE_VERSION),
+                     '--keys_output_dir={}'.format(folder),
+                     '--client_id=''',
```
Lagovas, line 1021: what will print next code:
```
a =  '--client_id='''
print(a)
```
?

I suggest to rewrite this literal to `'--client_id'` or `"--client_id=''"` to be more readable what the actual value will be used. Because now we check `client_id` validation, not empty value)

ZhmakAS: If you provide just `--client_id` it wont be parsed at all and default value will be used. 
So agree to use `"--client_id=''"`

Lagovas (approved):
lgtm if CI tests pass

tests/test.py
ZhmakAS: Actually, this code will produce just `-client_id=` which is totally fine for our scenario, but in case of simply passing `--client_id` it wont start because you are passing param but not specified its value:)

```
a =  '--client_id='''
print(a)
```

PR #455, opened 2021-11-11
Split vault unit tests

ZhmakAS:
We have continually have some phantom vault unit tests failures because of internal vault switching between secret engine V1 and V2. Vault needs some time to properly switch on new secret engine and our test may fail during this time.

So we decided to split tests to test it on different vault instances to escape any secrets engine changing. So this is draft PR for now, to test hot it will work if we are running them directly on CI.

## Checklist

  with new changes
- [-] CHANGELOG.md is updated (in case of notable or breaking changes)
- [-] Benchmark results are attached (if applicable)


.circleci/check_vault_connection.sh
```diff
@@ -7,12 +7,10 @@ NUM_PINGS="${NUM_PINGS:-15}"
...
-    --cacert 'tests/ssl/ca/ca.crt' \
-    -XGET "https://${TEST_VAULT_HOST}:${TEST_VAULT_PORT}/"
+  curl -XGET "https://${TEST_VAULT_HOST}:${TEST_VAULT_PORT}/"
 
-  if [ "$?" -eq "0" ]; then
+    #60 code is SSL certificate problem - means that we can reach host
+  if [ "$?" -eq "60" ]; then
     exit 0
   else
     sleep ${DELAY}
```
shadinua, line 18: It seems a bit unstable for me. For example, when for some reason we will trust that certificate, here we will get unexpected behavior.

This code is just about checking TCP connectivity (so far I understand) so nothing more should affect behavior changes.

I'd suggest this approach:
```
curl -s -k "https://${TEST_VAULT_HOST}:${TEST_VAULT_PORT}" && exit 0 || sleep "$DELAY"
```

ZhmakAS: Thanks @shadinua, but even with two separate vault instances we had such issue, so I decided restructure tests a little bit to mount v1/v2 paths only one time at the start of the tests so I redo these changes and keep using one instance of Vault.

Lagovas (approved):
lgtm. As I see, tests not failed in previous starts

shadinua (approved):

PR #459, opened 2021-11-16
zhars/fix_acra_keys_with_redis

ZhmakAS:
Fixed `acra-keys` read command to Redis communication. Previously it was used global var redisOption from keystore package to check passed redis parameters, but all `acra-keys` commands have its own RedisOption object, locally stored inside struct, so new function `IsKeyDirectory` was introduced which get RedisOptions from KeyStoreParameters interface.

## Checklist

  with new changes

### DIFF
--- a/cmd/acra-keys/keys/keystore.go
+++ b/cmd/acra-keys/keys/keystore.go
@@ -19,6 +19,7 @@ package keys
 import (
 	"errors"
 	"flag"
+	filesystemV2 "github.com/cossacklabs/acra/keystore/v2/keystore/filesystem"
 
 	"github.com/cossacklabs/acra/cmd"
 	"github.com/cossacklabs/acra/keystore"
@@ -27,7 +28,6 @@ import (
 	"github.com/cossacklabs/acra/keystore/keyloader/hashicorp"
 	keystoreV2 "github.com/cossacklabs/acra/keystore/v2/keystore"
 	"github.com/cossacklabs/acra/keystore/v2/keystore/api"
-	filesystemV2 "github.com/cossacklabs/acra/keystore/v2/keystore/filesystem"
 	filesystemBackendV2 "github.com/cossacklabs/acra/keystore/v2/keystore/filesystem/backend"
 
 	"github.com/go-redis/redis/v7"
@@ -119,7 +119,7 @@ func OpenKeyStoreForReading(params KeyStoreParameters) (keystore.ServerKeyStore,
 		return nil, err
 	}
 
-	if filesystemV2.IsKeyDirectory(params.KeyDir()) {
+	if IsKeyDirectory(params) {
 		return openKeyStoreV2(params, keyLoader)
 	}
 	return openKeyStoreV1(params, keyLoader)
@@ -132,7 +132,7 @@ func OpenKeyStoreForWriting(params KeyStoreParameters) (keyStore keystore.KeyMak
 		return nil, err
 	}
 
-	if filesystemV2.IsKeyDirectory(params.KeyDir()) {
+	if IsKeyDirectory(params) {
 		return openKeyStoreV2(params, keyLoader)
 	}
 	return openKeyStoreV1(params, keyLoader)
@@ -145,7 +145,7 @@ func OpenKeyStoreForExport(params KeyStoreParameters) (api.KeyStore, error) {
 		return nil, err
 	}
 
-	if filesystemV2.IsKeyDirectory(params.KeyDir()) {
+	if IsKeyDirectory(params) {
 		return openKeyStoreV2(params, keyLoader)
 	}
 	// Export from keystore v1 is not supported right now
@@ -159,7 +159,7 @@ func OpenKeyStoreForImport(params KeyStoreParameters) (api.MutableKeyStore, erro
 		return nil, err
 	}
 
-	if filesystemV2.IsKeyDirectory(params.KeyDir()) {
+	if IsKeyDirectory(params) {
 		return openKeyStoreV2(params, keyLoader)
 	}
 	// Export from keystore v1 is not supported right now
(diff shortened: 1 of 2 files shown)


cmd/acra-keys/keys/keystore.go
```diff
@@ -245,3 +245,27 @@ func openKeyStoreV2(params KeyStoreParameters, loader keyloader.MasterKeyLoader)
 	}
 	return keystoreV2.NewServerKeyStore(keyDirectory), nil
 }
+
+// IsKeyDirectory checks if the directory contains a keystore version 2 from KeyStoreParameters
+func IsKeyDirectory(params KeyStoreParameters) bool {
```
Lagovas, line 250: what about to change this legacy naming from IsKeyDirectory to `CheckV2KeyStore` or `IsKeyStoreV2`?
because this function checks is it keystore v2 and doesn't check is it keystore over some folder or FileSystem.

cmd/acra-keys/keys/read-keys_test.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = store.GenerateDataEncryptionKeys(clientID)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	readCmd.Execute()
```
Lagovas, line 73: how we check that command executed correctly?
and what about to add integration test too?

cmd/acra-keys/keys/keystore.go
ZhmakAS: Agree, will rename it

cmd/acra-keys/keys/read-keys_test.go
ZhmakAS: Read command will exit with exit 1 because of Fatal on any error so that tests will fail too.

cmd/acra-keys/keys/keystore.go
```diff
@@ -245,3 +245,27 @@ func openKeyStoreV2(params KeyStoreParameters, loader keyloader.MasterKeyLoader)
...
+		redisClient, err := filesystemBackendV2.OpenRedisBackend(&filesystemBackendV2.RedisConfig{
+			RootDir: params.KeyDir(),
+			Options: &redis.Options{
+				Addr:     redisOption.Addr,
+				Password: redisOption.Password,
+				DB:       redisOption.DB,
+			},
+		})
+		if err != nil {
+			log.WithError(err).Debug("Failed to find keystore v2 in Redis")
```
Lagovas, line 262:
```suggestion
			log.WithError(err).Debugln("Failed to find keystore v2 in Redis")
```
better to use appropriate Debug**ln** method for logging with new line at the end. because logrus may change API for such methods according to std `log` library that doesn't add \n at the end. Plus, we try to be compatible as much as possible  with standard library for case if we want to change logging backend.

cmd/redis.go
```diff
@@ -39,13 +43,24 @@ const (
...
 
 // RegisterRedisKeyStoreParameters registers CLI parameters for Redis (keystore).
 func RegisterRedisKeyStoreParameters() {
 	redisOptions.RegisterKeyStoreParametersWithPrefix("", "")
 }
 
+// ValidateRedisCLIOptions validate Redis CLI options.
+func ValidateRedisCLIOptions() {
+	if err := redisOptions.validateOptions(); err != nil {
+		log.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorWrongParam).Error(
```
Lagovas, line 58:
```suggestion
		log.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorWrongParam).Errorln(
```
same here

tests/test.py
```diff
@@ -3511,6 +3511,47 @@ def create_key_store_with_client_id_from_cert(self, extractor, dir_name):
             timeout=PROCESS_CALL_TIMEOUT)
 
 
+class TestAcraKeysWithRedis(unittest.TestCase):
+    def setUp(self):
+        self.master_key = get_master_key()
+        self.client_id = 'keypair1'
+
+    def test_read_command_keystore(self):
+        with tempfile.TemporaryDirectory() as keys_folder:
```
Lagovas, line 3520: as I understand, we don't need FS folder for redis storage. we just pass some folder path as parameter that will be used as key in redis. so we don't need to create any real directory

tests/test.py
```diff
@@ -3511,6 +3511,47 @@ def create_key_store_with_client_id_from_cert(self, extractor, dir_name):
...
+            subprocess.check_call(
+                ['./acra-keymaker',
+                 '--client_id={}'.format(self.client_id),
+                 '--generate_acrawriter_keys',
+                 '--redis_host_port=localhost:6379',
+                 '--keys_output_dir={}'.format(keys_folder),
```
Lagovas, line 3526: same here, we can omit it and leave default `.acrakeys` value

tests/test.py
ZhmakAS: If I am not mistaken actually it is used for creating path to keys inside Redis, i suggested to use such approach to save keys to redis uniquely so we dont need to flush all keys inside redis after test finished.

So in such case, the actual path to some keys inside redis will be:

/tmp/tmpmhsq9m28/keypair1_storage.pub

But without key specifying:

.acrakeys/keypair1_storage.pub

tests/test.py
```diff
@@ -112,6 +113,9 @@
 
 TEST_WITH_TRACING = os.environ.get('TEST_TRACE', 'off').lower() == 'on'
 TEST_TRACE_TO_JAEGER = os.environ.get('TEST_TRACE_JAEGER', 'off').lower() == 'on'
+TEST_REDIS_KEYS_DB = 0
+TEST_REDIS_TOKEN_DB = 1
```
Lagovas, line 117: can we move these variable to RedisMixin class level to not pollute global scope and allow to overwrite them in the inherited classes? because if some subclass will need change some of these values, it will change global variable for all test Classes, not only for itself. But if they assigned in the class level scope then it may be overrides in the subclass without touching global scope or parent class.

ZhmakAS: Done:)

Lagovas (approved):
lgtm but merge after successful tests. I am updating test container now. I will notice you when I finish and then restart test workflows please

Lagovas (approved):

PR #461, opened 2021-11-22
Refactor AT services

Lagovas:
There are nothing changed in features or public API. Only internal structures and implementations.
All shared components have moved to `AcraTranslatorData` structure that used to initialize services (HTTP, gRPC) and passed to callbacks that called on server initialization. 

Refactored initialization dependencies like poison record callbacks, tokenizer and moved them from redundant factories to program startup to initialize them at start and share.

Added registries that store callbacks for events from gRPC/HTTP servers. It helps extend implementation with build flags. Also, updates integration tests with new env variable that accept and pass build flags for compilation commands.

Updated `go_package` option in .proto file to full path to fix import in other .proto files

And renamed `TestPoisonRecordWholeCellStatusOff` (that exists in 2 examples) with proper suffix.

## Checklist

  with new changes


cmd/acra-translator/grpc_api/server_service_registry.go
```diff
@@ -0,0 +1,32 @@
...
+func AddgRPCServerSubscriber(subscriber GRPCServerSubscriber) {
+	lock.Lock()
+	grpcServerSubscribers = append(grpcServerSubscribers, subscriber)
+	lock.Unlock()
+}
+
+// OngRPCServerInit call all registered callbacks on gRPC server initialization
+func OngRPCServerInit(server *grpc.Server, data *common.TranslatorData, service DecryptService) {
+	lock.Lock()
+	defer lock.Unlock()
```
ZhmakAS, line 28: Do we really need to use `defer` here? If I am not mistaken the `OnServerInit` function does not return any error so here no possible `early-exit` scenarios, right? Could we place it after the loop statement?

ZhmakAS (approved):
Looks good to me

cmd/acra-translator/grpc_api/server_service_registry.go
Lagovas: Defer has used because of subscribers may panic. Yes, the program will interrupt and exit if nothing catches this panic. But for a future better to handle it from the start correctly.

ZhmakAS (approved):

PR #465, opened 2021-11-25
[T2316] fix flaky acra-censor's tests

Lagovas:
* Added `file.Sync` calls in tests before reading files to be sure that all data dumped
* A bit refactored `QueryWriter`'s time constants and changed them to variables. Now we can change them in our tests to reduce time of waiting (decreased from 6s to ~1s)
* Removed some cache deletion in our scripts. I checked that golang take care of invalidation result binaries with different golang versions. It detects that binary has been built by another compiler and rebuild it if need. 
* Added downloading golang sources of dependencies to our base docker image to reduce time of tests on CircleCi (and local too)

## Checklist

  with new changes

### DIFF
--- a/acra-censor/common/logging_logic.go
+++ b/acra-censor/common/logging_logic.go
@@ -10,15 +10,20 @@ import (
 	"os/signal"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 )
 
+const defaultSerializationTimeout = time.Second
+
 // DefaultSerializationTimeout defines a default ticker' timeout
-const DefaultSerializationTimeout = time.Second
+var DefaultSerializationTimeout = defaultSerializationTimeout
+
+const defaultWriteQueryChannelSize = 500
 
 // DefaultWriteQueryChannelSize defines size of channel used for writing input queries
-const DefaultWriteQueryChannelSize = 50
+var DefaultWriteQueryChannelSize = defaultWriteQueryChannelSize
 
 // QueryInfo defines format of exporting query into file
 type QueryInfo struct {
@@ -47,6 +52,8 @@ type QueryWriter struct {
 	serializationTimeout time.Duration
 	serializationTicker  *time.Ticker
 	logger               *log.Entry
+	// may be used for metrics and useful for tests
+	skippedQueryCount uint64
 }
 
 // NewFileQueryWriter creates QueryWriter instance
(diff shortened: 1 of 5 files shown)


vixentael (commented):
👀 👍

iamnotacake (approved):
It's great that some tests are now faster because of more parallelism.

And I guess from now on it is better to rebuild CI docker image every time the dependencies are changed for faster builds on CI.
Caching compilation artifacts would yield even better speedup, but it definitely shouldn't be inside a docker container, rather in CI cache, if the time difference shows any visible advantage at all. Plus, that cache would be specific for a certain Go version and other issues may appear...

acra-censor/common/logging_logic.go
```diff
@@ -155,7 +162,9 @@ func (queryWriter *QueryWriter) WriteQuery(query string) {
 	case queryWriter.signalWriteQuery <- query:
```
iamnotacake, line 162: (very minor thing) there's a redundant space in line above, couldn't comment that exact line since it was not part of the committed diff

Lagovas: hm... I don't understand which tab you mean. Plus all code passed through gofmt that removes extra tabs and tests verify that gofmt used for code.

Lagovas (commented):
It's great that some tests are now faster because of more parallelism.

> And I guess from now on it is better to rebuild CI docker image every time the dependencies are changed for faster builds on CI.
Yes, maybe we will automate it in a future. Because for that we should write infrastructure code that will verify dependency changes and force re-building.

> Caching compilation artifacts would yield even better speedup, but it definitely shouldn't be inside a docker container, rather in CI cache, if the time difference shows any visible advantage at all. Plus, that cache would be specific for a certain Go version and other issues may appear...

golang compile care about version specific changes and recompile if detect something like that. I tested it before applying these changes. I tried to compile same package with different golang versions but same GOPATH. And compiler skips recompilation if it has built with same version previously and hasn't changed. And rebuild if something changed or used another version.
But better to explicitly separate it between version. One of the future possible path to re-use cache and images, it is build dependency packages in base image with all used versions to separate GOPATH to avoid their building/rebuilding on a test run. We will use golang compiler building process when it compiles as separate compiled files and link/compile them together in the next step of building.

acra-censor/common/logging_logic.go
iamnotacake: Seems like it was rendered differently for some reason on my side

PR #473, opened 2021-12-15
Deprecate acra connector

Lagovas:
In this PR was deprecated acra-connector related CLI parameters in tools, updated configs. 
Additionally updated a few descriptions where mentioned only AcraConnector and added mentioning application because in a future applications will depend on these parameters.

Documentation hasn't been updated yet. After finishing the update, I will update and signal in this PR and mark as done appropriate list item. Only after that this PR will be merged. But it is ready to review

Docs update - https://github.com/cossacklabs/product-docs/pull/215

## Checklist

- [X] Public API has proper documentation in the [Acra documentation] site or has PR on [documentation repository] 
  with new changes


CHANGELOG_DEV.md
```diff
@@ -1,3 +1,11 @@
+## 0.91.0 - 2021-12-15
+### Deprecated
+- `acra-connector` - deprecated and will be removed in next releases. All related flags in
+  `acra-server`/`acra-translator`/`acra-keymaker` deprecated too and will be ignored in next releases.
```
shadinua, line 4: Do we really want to _ignore_ unknown/removed flags? It looks like we may get some security issues here. I'm not sure about exactly that flags, but the whole approach looks a bit danger.

I'd suggest to raise exception and do not start at all in such cases.

vixentael: > I'd suggest to raise exception and do not start at all in such cases.

i think the exception should be raised after these flags will be removed. right now the users can use these flags and AC, we just note that they are deprecated. on next release, when AC is removed, we shouldn't ignore, but will raise an exception

Lagovas: hm... you are right in case of `acra-keymaker` (looks like it safely to do like this here). In case of `acra-[server|translator]` we discussed about warnings. Because previously we didn't remove deprecated flags, just ignored them if they do nothing.
In case of `acra-connector`, application will fail and couldn't connect to `acra-server`, if they will expect `connector` on other side. But maybe really, better to exit with non-zero status and error log...
How do you suggest note about future changes? "Will be deprecated and will cause service shutdown with non-zero status"?

shadinua: > i think the exception should be raised after these flags will be removed
Sure. "Deprecated" status means we say we will remove them in future, but currently they still work with warnings in logs.

> How do you suggest note about future changes?
I'd say simply "they will be removed".

Lagovas:
Changed `ignored` to `removed`, updated documentation - https://github.com/cossacklabs/product-docs/pull/215

shadinua (approved):

PR #479, opened 2021-12-29
Handle null::text casts in postgresql dialect, fix panics

Lagovas:
Due to using sqlparser that targeted to MySQL, it didn't use to support postgresql's type castings, especially `null::text`.
Previously we extended SQLVal to support `::<type>` type casts, to avoid a lot of refactorings and re-use existing code that depends on SQLVal (for example our transparent encryption). But it's simplest way to support such type casts. Harder is to refactor `sql.y` and used SQL's grammar, and propagate casts to all kinds of lexical rules related to SQL values (`select expr`, `functions`, etc). 
But Acra works with simple SQL value literals and don't support complicated SQL values like functions or `(select 1)::text`, so for now we don't need to do it and can leave it for later (probably we will switch to another parser earlier than meet real case where it need:)

So here all complicated SQL expressions will be stored in `unknown` field of `SQLVal` to reuse existing code and just serialize/deserialize when it need as is.

## Checklist

- [X] Change is covered by automated tests
- [X] The [coding guidelines] are followed
- [X] Public API has proper documentation in the [Acra documentation] site or has PR on [documentation repository] 
  with new changes
- [X] CHANGELOG.md is updated (in case of notable or breaking changes)
- [X] CHANGELOG_DEV.md is updated
- [X] Benchmark results are attached (if applicable)
- [X] [Example projects and code samples] are up-to-date (in case of API changes) 

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -8762,15 +8762,6 @@ class TestInvalidCryptoEnvelope(unittest.TestCase):
         self.assertTrue(found, "Not found any expected exception")
 
 
-class TestPanicRecover(AcraCatchLogsMixin, BaseTestCase):
-    def testRecovering(self):
-        with self.assertRaises(DatabaseError):
-            # call unsupported query by sqlparser
-            self.engine1.execute('select null::text from dual;')
-
-        self.assertIn('panic in connection processing, close connection', self.read_log(self.acra).lower())
-
-
 class TestRegressionInvalidOctalEncoding(BaseTokenizationWithBinaryPostgreSQL):
     def testOctalIntegerValue(self):
         default_client_id_table = sa.Table(
--- a/CHANGELOG_DEV.md
+++ b/CHANGELOG_DEV.md
@@ -1,3 +1,6 @@
+## 0.92.0 - 2021-12-29
+- Improve sqlparser, support `null::text` type casts and avoid panics.
+
 ## 0.92.0 - 2021-12-22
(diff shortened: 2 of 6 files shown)


tests/test.py
```diff
@@ -8762,15 +8762,6 @@ def test_invalid_specified_values(self):
         self.assertTrue(found, "Not found any expected exception")
 
 
-class TestPanicRecover(AcraCatchLogsMixin, BaseTestCase):
```
vixentael, line 8765: no more test? 🥺

vixentael (commented):
👀

tests/test.py
Lagovas: it was one known case that caused panics. I don't know other cases to put here *( We can leave this test empty, but IMHO, we can write it one more time when we will need it.

vixentael: i thought this PR means that panics won't happen for null:text cast, so the test should always pass now.

Lagovas: And after this fix it doesn't panic anymore )

Lagovas: > i thought this PR means that panics won't happen for null:text cast, so the test should always pass now.

oh, you wrote it before me)) 10 sec difference) Yes, it fixes and doesn't panic anymore

vixentael: okay, thank you for the explanation! great job!

ZhmakaAS (approved):

PR #481, opened 2022-01-12
Remove acra-connector from acra-server/acra-translator

ZhmakaAS:
Initial removing acra-connector from acra-server and acra-translator cmds. 

## Checklist

  with new changes
- [X] CHANGELOG_DEV.md is updated


cmd/acra-translator/grpc_api/service_test.go
```diff
@@ -669,35 +651,6 @@ func testgRPCServiceFlow(ctx context.Context, expectedClientID []byte, conn *grp
...
-
-	keystorage.On("GetPrivateKey", mock.Anything).Return(testKeypair.Private, nil)
-	keystorage.On("GetPeerPublicKey", mock.Anything).Return(testKeypair.Public, nil)
-	conn := server.NewConnection(clientOpts, t)
-	defer conn.Close()
-
-	testgRPCServiceFlow(ctx, expectedClientID, conn, keystorage, t)
-}
-
 func TestNewFactoryWithClientIDFromSecureSessionConnectionInvalidAuthInfo(t *testing.T) {
```
Lagovas, line 654: this one we can remove too. but only if we have `<Something else>InvalidAuthInfo` check, to be sure that we handle case with incorrect auth info. But if not, then better to replace SS wrapper with TLS/something else and remove SS usage too

tests/ee_acrablock_config.yaml
```diff
@@ -16,7 +16,7 @@ schemas:
       - masked_prefix
     encrypted:
       - column: specified_client_id
-        client_id: keypair1
+        client_id: 12824c1c541a615f428a740770291374a8504f84a0682ab4015bc8e792b1bc8240022c5c9faa8c2111e0076b3b929148f4b801414413edaa800cb42492c20cf7
```
Lagovas, line 19: please, add comments that it's hash of tls certificate with relative path to it. it will simplify understanding how to reproduce or what we should use to re-generate if certs will be changed.

tests/test.py
```diff
@@ -154,6 +154,8 @@
 TEST_VAULT_TLS_CA = abs_path(os.environ.get('TEST_VAULT_TLS_CA', 'tests/ssl/ca/ca.crt'))
 VAULT_KV_ENGINE_VERSION=os.environ.get('VAULT_KV_ENGINE_VERSION', 'v1')
 CRYPTO_ENVELOPE_HEADER = b'%%%'
+TLS_CERT_CLIENT_ID_1 = None
```
Lagovas, line 157: please, add comment where this variables should be initialized (`setupModule` function). now I used ctrl+f to understand is it temporal stub during refactoring or new variables that used somewhere)

tests/test.py
```diff
@@ -844,11 +859,14 @@ class AsyncpgExecutor(QueryExecutor):
     BinaryFormat = 'binary'
 
     def _connect(self, loop):
+        ssl_context = ssl.create_default_context(cafile=self.connection_args.ssl_ca)
+        ssl_context.load_cert_chain(self.connection_args.ssl_cert, self.connection_args.ssl_key)
+        ssl_context.check_hostname = False
```
Lagovas, line 865: it doesn't work if we will check hostname? all our certs generated for `localhost` and we connect to `localhost` in tests. did you try?

tests/test.py
```diff
@@ -1035,7 +1053,7 @@ def get_acraserver_connection_string(self, port=None):
         """use similar to connector unix socket connection string to allow connect directory to acra by db driver"""
         if not port:
             port = self.ACRASERVER_PORT
-        return self.get_connector_connection_string(port)
+        return get_acraserver_unix_connection_string(port)
```
Lagovas, line 1057: hm... is it really use TLS after that? because PSQL doesn't establish TLS over unix sockets

tests/test.py
```diff
@@ -1394,17 +1302,11 @@ def _fork_acra(self, acra_kwargs, popen_kwargs):
...
             args['incoming_connection_prometheus_metrics_string'] = self.get_prometheus_address(
                 self.ACRASERVER_PROMETHEUS_PORT)
-        if self.CONNECTOR_TLS_TRANSPORT:
-            args['acraconnector_tls_transport_enable'] = 'true'
         if self.with_tls():
             args['tls_key'] = TEST_TLS_SERVER_KEY
             args['tls_cert'] = TEST_TLS_SERVER_CERT
             args['tls_ca'] = TEST_TLS_CA
             args['tls_auth'] = ACRA_TLS_AUTH
-            args['tls_ocsp_url'] = 'http://localhost:{}'.format(self.OCSP_SERVER_PORT)
```
Lagovas, line 1404: why we don't use it now? if it worked before with `self.with_tls`, what problem do we have now?

tests/test.py
```diff
@@ -1455,6 +1357,7 @@ def fork_translator(self, translator_kwargs, popen_kwargs=None):
 
         cli_args = ['--{}={}'.format(k, v) for k, v in default_config.items()]
 
+        print(cli_args)
```
Lagovas, line 1361: don't forget to remove it) or use `logger`/`logging` for that, because it configurable

tests/test.py
```diff
@@ -1701,19 +1607,19 @@ def checkSkip(self):
     def setUp(self):
         super().setUp()
 
-        def executor_with_port(port):
+        def executor_with_port(ssl_key, ssl_cert):
```
Lagovas, line 1611: if we changed signature and function meaning then we should rename the function to not confuse other readers.

tests/test.py
```diff
@@ -3488,24 +3247,6 @@ def test_read_keys_symmetric_zone(self):
...
-                '--acratranslator_transport_key',
-                '--poison_record_keys',
-                '--audit_log_symmetric_key',
-                '--keys_dir={}'.format(self.dir_with_distinguished_name_client_id.name),
-                '--keys_dir_public={}'.format(self.dir_with_distinguished_name_client_id.name),
-                '--keystore={}'.format(KEYSTORE_VERSION),
-            ],
-                env={ACRA_MASTER_KEY_VAR_NAME: self.master_key},
-                stderr=subprocess.STDOUT)
-        self.assertIn("--client_id or --tls_cert is required to generate keys".lower(), exc.exception.output.decode('utf8').lower())
```
Lagovas, line 3506: hm, why you removed this test? as I understand we test there requirement to pass `client_id` or `tls_cert` arguments

tests/test.py
```diff
@@ -4529,310 +4242,61 @@ class TestZoneIDDecryptionWithVaultMasterKeyLoader(HashiCorpVaultMasterKeyLoader
 
 
 class AcraTranslatorTest(AcraTranslatorMixin, BaseTestCase):
+    def get_http_schema(self):
+        return 'https'
 
-    def checkSkip(self):
-        return
+    def get_grpc_channel(self, port):
```
Lagovas, line 4249: why we can't re-use same method from `AcraTranslatorMixin` that configure using same keys?

tests/test.py
```diff
@@ -4529,310 +4242,61 @@ class TestZoneIDDecryptionWithVaultMasterKeyLoader(HashiCorpVaultMasterKeyLoader
...
-                response = request_func(connector_port, correct_client_id, None, data)
-                decrypted = deserialize_and_decrypt_acrastruct(response, client_id_private_key, 'keypair1')
-                self.assertEqual(data, decrypted)
-                # test with correct zone id
-                print("encrypt with zone {}".format(zone['id']))
-                response = request_func(
-                    connector_port, client_id, zone['id'], data)
-                print("decrypt with zone {}".format(zone['id']))
-                decrypted = deserialize_and_decrypt_acrastruct(response, zone_private_key, zone_id=zone['id'].encode('ascii'))
+        key_folder = tempfile.TemporaryDirectory()
```
Lagovas, line 4273: hm, why we don't use same test pattern with existing keys? we have already generated keys for tests from certs. why we re-generate them in temp folder?

tests/test.py
```diff
@@ -4529,310 +4242,61 @@ class TestZoneIDDecryptionWithVaultMasterKeyLoader(HashiCorpVaultMasterKeyLoader
...
+        try:
+            client_id = self.get_valid_certificate_identifier()
+            self.assertEqual(create_client_keypair(name=client_id, keys_dir=key_folder.name), 0)
```
Lagovas, line 4276: plus better to generate keys from certificate instead of extracted client_id, as it will do users) and we have function for that

tests/test.py
```diff
@@ -4529,310 +4242,61 @@ class TestZoneIDDecryptionWithVaultMasterKeyLoader(HashiCorpVaultMasterKeyLoader
...
-                    connector_port, client_id, incorrect_zone['id'],
-                    acrastruct_with_zone)
-                self.assertNotEqual(data, response)
-
-            # wait decryption error with incorrect client id
-            with ProcessContextManager(self.fork_connector_for_translator(connector_port2, translator_port, incorrect_client_id)):
-                response = request_func(connector_port2, incorrect_client_id, None, acrastruct)
-                self.assertNotEqual(data, response)
-
-    def testHTTPApiResponses(self):
```
Lagovas, line 4625: here too. we can test it with `https://` and without connector

tests/ee_acrablock_config.yaml
ZhmakaAS: Sure, do you want to add comments in all corresponded  config files?

tests/test.py
ZhmakaAS: Ok, will do that

tests/test.py
ZhmakaAS: I didn't try, but re-test all with  `check_hostname=True` and its working well.

tests/test.py
ZhmakaAS: In case of TLS we have tcp connection string

```
def get_acraserver_unix_connection_string(port):
    if TEST_MYSQL:
        return get_tcp_connection_string(port)
    else:
        if TEST_WITH_TLS:
            return get_tcp_connection_string(port)
        else:
            return 'unix://{}/.s.PGSQL.{}'.format(PG_UNIX_HOST, port)

```

tests/test.py
ZhmakaAS: Returned it back, It works, i think kind of typo, thanks

tests/test.py
ZhmakaAS: Sure:)

tests/test.py
ZhmakaAS: You are right!

tests/test.py
ZhmakaAS: Not exactly, you are right that this test was used to check requirement to pass `client_id` or `tls_cert`, but now threre are no                 '--acratranslator_transport_key' and `--acraserver_transport_key` cli args for `acra-keys generate` so there in no sense for this test in such case because `poison_record_keys` args doesnt require `clien_id` or `tls_cert`, am I right?

tests/test.py
ZhmakaAS: As I understood, we are generating it in separate temp folder to able to test different `certificate_identifier` like distinguished_name or serial_number in such case this tests looks generic and with help of `TLSAuthenticationBySerialNumberMixin` or `TLSAuthenticationByDistinguishedNameMixin` we can change the logic.

But with the statically declared keys_dir and client_id we will miss such flexibility? What do you think?

tests/test.py
ZhmakaAS: Will return this test

tests/test.py
ZhmakaAS: I put it here because here - https://github.com/cossacklabs/acra/blob/master/tests/test.py#L5588 we are using native test call and we should expect here secure grpc channel by default - I think we should delete duplication of this method from `AcraTranslatorMixin`

ZhmakaAS: I think, you are right, we can omit that function here

tests/test.py
ZhmakaAS: Actually we do generate keys from cert but not extract, probably we should change the actual function name.
Here is the source of get_valid_certificate_identifier:

```
    def get_valid_certificate_identifier(self, certificate_file=None):
        if certificate_file is None:
            certificate_file = TEST_TLS_CLIENT_CERT
        # converted data from certificate "CN=Test leaf certificate (acra-writer),OU=IT,O=Global Security,L=London,ST=London,C=GB"
        with open(certificate_file, 'rb') as f:
            pem = f.read()
        cert = x509.load_pem_x509_certificate(pem, default_backend())
        sha512 = hashlib.sha512()
        sha512.update(cert.subject.rfc4514_string().encode('utf-8'))
        return sha512.hexdigest().lower()
```

tests/test.py
ZhmakaAS: But we already have such test

https://github.com/cossacklabs/acra/blob/master/tests/test.py#L3421

Here is the souce test we have, without passing transport keys we will be able to test such requirement.

tests/ee_acrablock_config.yaml
Lagovas: don't forget to add a comment. because now it's omitted for all of files.

cmd/acra-server/acra-server.go
```diff
@@ -150,25 +148,23 @@ func realMain() error {
...
 	tlsDbSNI := flag.String("tls_database_sni", "", "Expected Server Name (SNI) from database")
 	tlsDbSNIOld := flag.String("tls_db_sni", "", "Expected Server Name (SNI) from database (deprecated, use \"tls_database_sni\" instead)")
 	tlsDbCert := flag.String("tls_database_cert", "", "Path to client TLS certificate shown to database during TLS handshake (overrides \"tls_cert\")")
 	tlsDbKey := flag.String("tls_database_key", "", "Path to private key of the TLS certificate used to connect to database (see \"tls_database_cert\")")
-	tlsUseClientIDFromCertificate := flag.Bool("tls_client_id_from_cert", false, "Extract clientID from TLS certificate. Take TLS certificate from application/AcraConnector's connection if acraconnector_tls_transport_enable is TRUE; otherwise take TLS certificate from application's connection if acraconnector_transport_encryption_disable is TRUE. Can't be used with --tls_client_auth=0 or --tls_auth=0")
+	tlsUseClientIDFromCertificate := flag.Bool("tls_client_id_from_cert", false, "Extract clientID from TLS certificate from application connection. Can't be used with --tls_client_auth=0 or --tls_auth=0")
```
Lagovas, line 165: now this parameter should be `true` as default mode for acra-server

tests/test.py
```diff
@@ -309,7 +309,7 @@ def get_pregenerated_random_data():
         return f.read()
 
 
-def create_acrastruct_with_client_id(data, client_id='keypair1'):
+def create_acrastruct_with_client_id(data, client_id=TLS_CERT_CLIENT_ID_1):
```
Lagovas: hm... it's strange that it works if some code relied on default `client_id`) because now at function declaration time client_id will be None everytime because of TLS_CERT_CLIENT_ID_1 re-initialized after interpreting and compiling function. What about to remove default value to force explicit declaring (now all code specify client_id explicitly) to not confuse in the future if something will be broken due to invalid value?

tests/test.py
```diff
@@ -690,6 +642,18 @@ def setUpModule():
     drop_tables()
 
 
+def extract_client_id_from_cert(tls_cert, extractor='distinguished_name'):
```
Lagovas: because of found several usage of these constant value `distinguished_name`, can we create some global value for it? Or constant class value like:
```
class TLSClientIDSource:
  DN = 'distinguished_name'
  SERIAL = 'serial_number'
```
and use it instead of hardcoded values?

tests/test.py
```diff
@@ -1378,6 +1225,9 @@ def _fork_acra(self, acra_kwargs, popen_kwargs):
             # we doesn't need in tests waiting closing connections
             'incoming_connection_close_timeout': 0,
             self.ACRA_BYTEA: 'true',
+            'tls_client_id_from_cert': 'true',
```
Lagovas: better to remove it from base class to test acra with default parameters as base test case.

tests/test.py
Lagovas: hm... yes, we should change it)) because this logic may work differently from how it works in golang code (that is main reason why we provided `extract-client-id` command for `acra-keys`). So... it I left a time bomb *(

But my original question was related to flow `get client_id` and `call create_client_keypair` instead of `create_client_keypair_from_certificate`. Why we get client_id from somewhere and pass it to `create_client_keypair` instead of calling `acra-keymaker` passing path to certificate?

tests/test.py
```diff
@@ -6243,59 +5664,6 @@ class TestTLSAuthenticationWithConnectorByDistinguishedName(TLSAuthenticationByD
     def setUp(self):
         if not TEST_WITH_TLS:
             self.skipTest("Test works only with TLS support on db side")
-        self.acra_writer_id = self.get_valid_certificate_identifier()
```
Lagovas: why these tests remove and it named `TestTLSAuthenticationWithConnectorByDistinguishedName` (with connector:)

Lagovas (commented):
and general petition - make ctrl+f "connector" and remove it where it possible if it doesn't take a lot of time, please

tests/ee_acrablock_config.yaml
ZhmakaAS: Sure, thanks for reminder

cmd/acra-server/acra-server.go
ZhmakaAS: Yes, as we discussed

tests/test.py
ZhmakaAS: Lets create corresponded global variables

tests/test.py
```diff
@@ -2904,22 +2593,23 @@ def setUp(self):
...
-                acraserver_port=self.ACRASERVER_PORT,
-                client_id=self.client_id,
-                api_port=self.CONNECTOR_API_PORT_1,
-                zone_mode=True,
-                keys_dir=self.connector_keys_dir)
             if not self.EXTERNAL_ACRA:
                 self.acra = self.fork_acra(
                     zonemode_enable='true',
                     http_api_enable='true',
+                    tls_ocsp_from_cert='ignore',
```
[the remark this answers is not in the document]
ZhmakaAS: Same here

tests/test.py
```diff
@@ -2157,10 +1989,13 @@ def close(self):
     def setUp(self):
         self.checkSkip()
         try:
-            self.connector_1 = self.fork_connector(
-                self.CONNECTOR_PORT_1, self.ACRASERVER_PORT, 'keypair1')
             if not self.EXTERNAL_ACRA:
-                self.acra = self.fork_acra()
+                self.acra = self.fork_acra(
+                    tls_ocsp_from_cert='ignore',
```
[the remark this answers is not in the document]
ZhmakaAS: Discussed to left it as is for now.

Lagovas (approved):
great job

PR #482, opened 2022-01-18
CRL validation improvements

iamnotacake:
CRL contains revoked certificates as a simple list, which means validating certificates is O(N).
Most certificates will be valid which means validator would have to search through the whole list each time.
This improvement is about making verification O(1) operations thanks to `map` being used instead of plain array.

## Checklist

  with new changes~

### DIFF
--- a/network/crl.go
+++ b/network/crl.go
@@ -22,7 +22,6 @@ import (
 	"crypto/x509/pkix"
 	"errors"
 	"io/ioutil"
-	"math/big"
 	"net/http"
 	url_ "net/url"
 	"sync"
@@ -234,9 +233,9 @@ func (c DefaultCRLClient) Fetch(url string, allowLocal bool) ([]byte, error) {
 
 // CRLCacheItem is combination of fetched+parsed+verified CRL with fetch time
 type CRLCacheItem struct {
-	Fetched             time.Time                            // When this CRL was fetched and cached
-	CRL                 pkix.CertificateList                 // Parsed CRL itself
-	RevokedCertificates map[*big.Int]pkix.RevokedCertificate // Copy of CRL.TBSCertList.RevokedCertificates with SerialNumber as key
+	Fetched             time.Time                           // When this CRL was fetched and cached
+	CRL                 pkix.CertificateList                // Parsed CRL itself
+	RevokedCertificates map[string]*pkix.RevokedCertificate // Copy of CRL.TBSCertList.RevokedCertificates with SerialNumber as key
 }
 
 // CRLCache is used to store fetched CRLs to avoid downloading the same URL more than once,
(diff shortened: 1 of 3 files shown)


network/crl.go
```diff
@@ -331,23 +334,36 @@ func (v DefaultCRLVerifier) getCachedOrFetch(url string, allowLocal bool, issuer
...
+		revokedCertificates[cert.SerialNumber] = cert
+	}
+
 	if crl.TBSCertList.NextUpdate.Before(time.Now()) {
 		log.Warnf("CRL: CRL at %s is outdated", url)
 		return nil, ErrOutdatedCRL
 	}
 
+	// We don't need this array anymore as all revoked certificates are now inside the `map`
+	crl.TBSCertList.RevokedCertificates = make([]pkix.RevokedCertificate, 0)
```
ZhmakaAS, line 348: Do we really need to initialize a new empty slice. I think you can just set it to nil or you are going to reuse it somewhere else?

network/crl_test.go
```diff
@@ -538,9 +545,9 @@ func testDefaultCRLVerifierWithGroup(t *testing.T, certGroup TestCertGroup) {
 }
 
 func TestDefaultCRLVerifier(t *testing.T) {
-	testDefaultCRLVerifierWithGroup(t, getTestCertGroup(t))
-	testDefaultCRLVerifierWithGroup(t, getTestCertGroup3(t))
-	testDefaultCRLVerifierWithGroup(t, getTestCertGroupOnlyRoot(t))
+	// testDefaultCRLVerifierWithGroup(t, getTestCertGroup(t))
```
vixentael, line 548: 👀

network/crl.go
iamnotacake: Tried `= []`, forgot about `= nil`, thanks, will fix

network/crl.go
```diff
@@ -232,8 +234,9 @@ func (c DefaultCRLClient) Fetch(url string, allowLocal bool) ([]byte, error) {
 
 // CRLCacheItem is combination of fetched+parsed+verified CRL with fetch time
 type CRLCacheItem struct {
-	Fetched time.Time // When this CRL was fetched and cached
-	CRL     pkix.CertificateList
+	Fetched             time.Time                            // When this CRL was fetched and cached
+	CRL                 pkix.CertificateList                 // Parsed CRL itself
```
Lagovas, line 238: hm, why we store it as value and make several copy operations if `v.Client.Fetch(url, allowLocal)` returns pointer? can we store just pointer and avoid extra copies?

Lagovas: plus, what about to encapsulate all this fields and use CRLCacheItem as interface, and initialize it with constructor?
because previously we initialized with `cacheItem := &CRLCacheItem{Fetched: time.Now(), CRL: *crl}`
now we prepare map for it, set nil for a field of CRL result and only after that create struct value with:
```
cacheItem := &CRLCacheItem{
		Fetched:             time.Now(),
		CRL:                 *crl,
		RevokedCertificates: revokedCertificates,
	}
```
But it's not `DefaultCRLVerifier` responsibility to know how CacheItem works. It should initialize it with CRL result and `CacheItem` prepare itself and his map as it wants. And just provide simple API how to create item and how to get info about revoked cert: `CacheItem.Get(cert *x509.Certificate)(revokedCert, ok)`

And it is encapsulates that it uses SerialNumber of certificate for map, not some hash or any other cert fields.

network/crl.go
```diff
@@ -331,23 +334,36 @@ func (v DefaultCRLVerifier) getCachedOrFetch(url string, allowLocal bool, issuer
...
 		log.Warnf("CRL: CRL at %s is outdated", url)
 		return nil, ErrOutdatedCRL
 	}
 
+	// We don't need this array anymore as all revoked certificates are now inside the `map`
+	crl.TBSCertList.RevokedCertificates = make([]pkix.RevokedCertificate, 0)
+
+	cacheItem := &CRLCacheItem{
+		Fetched:             time.Now(),
+		CRL:                 *crl,
```
Lagovas, line 352: do we need to do copy of whole list instead of copying pointer?

network/crl.go
```diff
@@ -395,57 +395,57 @@ func checkCertWithCRL(cert *x509.Certificate, cacheItem *CRLCacheItem) error {
...
-			case "2.5.29.33":
-				// section 4.2.1.5, id-ce-policyMappings
-
-			default:
-				if extension.Critical {
-					log.WithField("oid", extension.Id.String()).Warnln("CRL: Unable to process critical extension with unknown Object ID")
-					return ErrUnknownCRLExtensionOID
-				}
-				log.WithField("oid", extension.Id.String()).Debugln("CRL: Unable to process non-critical extension with unknown Object ID")
+	revokedCertificate, ok := cacheItem.RevokedCertificates[cert.SerialNumber.Text(16)]
```
Lagovas, line 398: if we use same integer in several places, lets use constant for that to avoid remembering in future changes

network/crl.go
iamnotacake: Changed struct to contain pointer instead, and put `CRL: crl` here

network/crl.go
iamnotacake: BTW, you also proposed using base62 for this to have smaller strings. But seems like `.Text()` has [special implementation for power-of-two bases](https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/math/big/natconv.go;l=282;drc=refs%2Ftags%2Fgo1.17.6), which makes me think this should work faster with no division involved. Then I will make the constant and set it to 32. A bit longer that base62, but hashing should be faster.

network/crl.go
iamnotacake: Let's do it in next CRL-related PR? Gonna mention your comment in the task.

vixentael (approved):
👍  performance ++

PR #495, opened 2022-02-14
Add docker-compose for benchmarks with collection profiling data

Lagovas:
Here added new `run_transparent.sh` script in `benchmarks/` folder that runs simple commands that generate data in simple table and then read from database directly and through acra-server. That helps compare gap between DB and DB + Acra.
It is simplified existing `benchmark/run.sh` script, with calling same logic as for direct database, but through acra-server in transparent mode too.
Additionally added fetching profiling data from http debug server to analyze it with `go tool pprof` after that.

For now it's only for postgresql because used existing code that worked only with this db.

## Checklist

  with new changes


vixentael (commented):
👀

shadinua (approved):

benchmarks/common/consts.go
```diff
@@ -12,15 +12,15 @@
...
-	RowCount = 10000
+	RowCount = 6000
 	// RequestCount num of requests that will be done in each read benchmark
-	RequestCount = 10000
+	RequestCount = 80000
 	// ZoneCount num of zones which will be generated and used
 	ZoneCount = 100
 	// MaxDataLength size of test random data that will be generated and inserted to db (before encrypting)
-	MaxDataLength = 100 * 1024 // 100 kb
+	MaxDataLength = 1 * 1024 // 100 kb
```
iamnotacake, line 25:
```suggestion
	MaxDataLength = 1 * 1024 // 1 KiB
```

benchmarks/common/db.go
```diff
@@ -18,12 +18,13 @@ package common
...
 	"fmt"
-	"github.com/cossacklabs/acra/benchmarks/config"
+	_ "github.com/jackc/pgx/v4/stdlib"
+	"github.com/sirupsen/logrus"
 	"os"
 )
 
 func connect(connectionString string) *sql.DB {
-	db, err := sql.Open("postgres", connectionString)
+	db, err := sql.Open("pgx", connectionString)
```
iamnotacake, line 27: So we switch to a different PgSQL driver? It is better in terms of performance?

benchmarks/common/db.go
```diff
@@ -86,20 +87,21 @@ func DropCreateRaw(db *sql.DB) {
...
-			fmt.Printf("Error: on sql - %v\n", script)
+			logrus.Debugf("Error: on sql - %v\n", script)
 			panic(err)
 		}
 	}
 }
 
 // IsExistsData checks is exists table with name <tablename>
 func IsExistsData(tablename string, db *sql.DB) bool {
+	return false
```
iamnotacake, line 101: Sure we need this line?

benchmarks/README.md
```diff
@@ -0,0 +1,86 @@
...
+## Prepare environment 
+
+```
+export OUTPUT_DIR=${OUTPUT_DIR:-/tmp}
+export BENCHMARK_NAME=${BENCHMARK_NAME:-default}
+export READ_TIMEOUT=${READ_TIMEOUT:-25}
+export WRITE_TIMEOUT=${WRITE_TIMEOUT:-25}
+```
+
+* `OUTPUT_DIR` - here will be saved profiling data
```
iamnotacake, line 18:
```suggestion
* `OUTPUT_DIR` - where to store profiling data
```

iamnotacake:
One thing that came to my mind is that we may want to store PostgreSQL's data (which by default lives in `/var/lib/postgres/data` or somewhere there) in `tmpfs` to ensure that the time needed to write/read all the columns data (directly, or through AcraServer) will be equal

benchmarks/run_transparent.sh
```diff
@@ -0,0 +1,49 @@
...
+export WRITE_TIMEOUT=${WRITE_TIMEOUT:-25}
+
+
+go build -o read.bin ./benchmarks/cmd/read/direct/direct.go
+go build -o write.bin ./benchmarks/cmd/write/raw/raw.go
+
+echo "run write scripts"
+# direct to database
+export PG_CONNECTION_STRING=${DIRECT_CONNECTION_STRING}
+script="./write.bin 2> >(tee -a ${BENCHMARK_NAME}.direct.txt)"
```
iamnotacake, line 18: Weird construction. Just curious, does it mean the same as this?
`script="./write.bin 2>&1 | tee -a ${BENCHMARK_NAME}.direct.txt" # append both stdout and stderr to that file`

benchmarks/common/db.go
Lagovas: there are occurred errors with the previous driver with inserting binary data. plus this driver is stopped support and live only in maintenance mode. Pgx is pure golang driver that suggested by https://github.com/lib/pq driver ("postgres").

benchmarks/run_transparent.sh
Lagovas: updated to your variant

benchmarks/README.md
Lagovas: updated

iamnotacake (approved):

benchmarks/common/db.go
Lagovas: removed this function because better to always re-generate, because we use also in transparent mode and to be sure that data generated with fresh encryptor_config, new keys, etc. With such checks it should be manually checked and more likely to forget and test with wrong data.

PR #498, opened 2022-02-16
Cache fetching rotated key names

Lagovas:
* added caching results from `store.getCachedHistoricalPrivateKeyFilenames` call that reads folder every time when decryption keys are requested.
* due to `[]byte` type used in our in-memory cache with interface `Get(id[]byte)[]byte`/`Set([]byte, []byte)`, to cache results used msgpack as [more efficient](https://github.com/alecthomas/go_serialization_benchmarks) way to serialize/deserialize than `gob` or `split/join` strings.
* struct for msgpack placed in `internal` package because msgpack can serialize only public structs. And this struct shouldn't be public and exported from `keystore/filesystem` package. So to hide and encapsulate this struct, it placed to `internal` that accessible from `keystore/filesystem`, but not accessed from top-level packages.
* added benchmarks that prove efficient serialization in comparison with simple strings operations
* because of checked profiling data that this function removed from call graph, also updated docker-compose and added tmpfs usage for postgresql that decreased time of tests and works more predictable. It cannot be used for keystore because docker doesn't support tmpfs volumes shared between containers.
* also in prifling data found that we have extra allocations in `getSymmetricKeys` due to small start allocation `keys := make([][]byte, 0, 1)` when usually may be at least 2 keys (current and 1 rotated). So increased to 4. It is not increase memory usage because it's allocates +3 structs with 3 `int` fields (SliceHeader with ptr + length + capacity), so it's just 3 * 24 bytes (72) instead of 24.

## Checklist

  with new changes


keystore/filesystem/internal/msgp_structs.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2016, Cossack Labs Limited
```
iamnotacake, line 1: Shouldn't we put current year here?
```suggestion
// Copyright 2022, Cossack Labs Limited
```

keystore/filesystem/server_keystore.go
```diff
@@ -387,12 +390,55 @@ func (store *KeyStore) GetPublicKeyFilePath(filename string) string {
 	return fmt.Sprintf("%s%s%s", store.publicKeyDirectory, string(os.PathSeparator), filename)
 }
 
+// use key started with "." (dot) because it's invalid character for clientID that generally stored in cache and
+// it will not intersect with other keys
+const cacheKeySuffix = ".historical."
```
iamnotacake, line 395: Based on how you use it, it's more like a `cacheKeyPrefix` :thinking:

iamnotacake:
We probably _could_ make keystore cache interface to be `[]byte <-> interface{}` instead of `[]byte <-> []byte`, but I guess existing scenario (caching bytes array) is more common in practice, and using interface would need type casting on each access, so probably some little overhead will be involved

keystore/filesystem/server_keystore.go
```diff
@@ -387,12 +390,55 @@ func (store *KeyStore) GetPublicKeyFilePath(filename string) string {
...
+
 // GetHistoricalPrivateKeyFilenames return filenames for current and rotated keys
 func (store *KeyStore) GetHistoricalPrivateKeyFilenames(filename string) ([]string, error) {
 	// getHistoricalFilePaths() expects a path, not a name, but we must return names.
 	// Add private key directory path and then remove it to avoid directory switching.
 	fullPath := filepath.Join(store.privateKeyDirectory, filename)
-	paths, err := getHistoricalFilePaths(fullPath, store.fs)
+	paths, err := store.getCachedHistoricalPrivateKeyFilenames(fullPath)
+	if err == nil {
+		return paths, nil
```
iamnotacake, line 431: Guess I'm missing something, still not finished reviewing changes, but how can we be sure that there were no changes in keystore (in filesystem or Redis or whereever) when returning cached (possibly outdated) version?

keystore/filesystem/server_keystore_test.go
```diff
@@ -768,3 +976,58 @@ func testHistoricalKeyAccess(storage Storage, t *testing.T) {
 		}
 	}
 }
+
+// BenchmarkHistoricalPathsSerialization compare simple way of serialization []string into []byte with strings.[Split|Join] and msgpack
+// Example of benchmarks to see comparison how msgpack faster simple join/split operations with strings
+// goos: linux
+// goarch: amd64
+// pkg: github.com/cossacklabs/acra/keystore/filesystem
+// cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
```
iamnotacake, line 985: Did you lock CPU frequency at this specific (or some other) value when benchmarking?
BTW, to get more precise results, instead of closing other apps, sending them SIGSTOP will also do the trick :+1:

keystore/filesystem/server_keystore.go
Lagovas: as we discussed previously, for now we guess and it is user's responsibility to take care about it and probably use HTTP API to reset cache after adding new keys. and we have task T2479 to add this feature and track keystore changes

keystore/filesystem/server_keystore_test.go
Lagovas: for what we need here such precise benchmarks?  there are 10^10 difference between approaches. you think such improvements may change general picture of that and we should spent time for that?)

Lagovas: it's just simple benchmark to show other developers who will read code and think: why you used this serialization algo instead of some more simple from std. and to allow everyone reproduce by one command on their machine to be sure

keystore/filesystem/server_keystore.go
Lagovas: yeah, renamed

Lagovas:
> We probably could make keystore cache interface to be []byte <-> interface{} instead of []byte <-> []byte, but I guess existing scenario (caching bytes array) is more common in practice, and using interface would need type casting on each access, so probably some little overhead will be involved

I though about it... but I don't want to use reflection here because it more type checks, more code like:
```
value, ok := cache.Get(key)
if !ok{...}
typedValue, ok := value.([]string}
if !ok {...}
```

for now, it is enough to convert everything to bytes, imho. you can add more arguments why we should change this approach.

keystore/filesystem/server_keystore_test.go
iamnotacake: Just saying what comes in mind. Here we clearly see that `msgpack` is much better, that's enough.

BTW, if you wanted to try something like this, I got a shell function to set min/max frequencies, at least it worked for some Intel laptop CPU
```
set-cpu-freq-min-max() {
	[[ $1 > 0 && $2 > 0 ]] || { echo "$0 <min_freq_KHz> <max_freq_KHz>"; return }
	echo $1 | sudo tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_min_freq
	echo $2 | sudo tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_max_freq
}
```

iamnotacake (approved):

PR #503, opened 2022-02-18
Update go version up to 1.17.7

shadinua:
* Add checksums for 3 latest major versions
* Update the default version which we use to build


docker/_scripts/acra-build/install_go.sh
```diff
@@ -51,7 +51,7 @@ if [ ! -f "$CSUMS_FILE" ]; then
   exit 1
 fi
 
-GO_VERSIONS="${GO_VERSIONS:-1.15.2}"
+GO_VERSIONS="${GO_VERSIONS:-1.17.7}"
```
Lagovas, line 54: We test with least supportable version to be sure, that we support 2 previous before latest golang version

shadinua: Hmmm. Sounds like a workaround. Probably, I don't understand the whole idea, but:
* If we rely on CI smoke test results, there are we perform tests with all supported (3 latest major) versions. Thus, this value doesn't matter in this case.
* If we do rely on building and test with _the only_ version specified here, then how could we know that it works with other Go major versions?

Lagovas: We have one fast tests that if works with 1.15, it will work with 1.17. and we test complicated integration tests with all tests that includes all dependencies. It.s just decreasing time of unit-tests

PR #504, opened 2022-02-21
use Get*SymmetricKey API instead of Get*SymmetricKeys

Lagovas:
* Use the optimized new method for fetching symmetric encryption key instead of set of keys
* Update tests that verify that new method used
* adapt integration test script for python3.6 that used in Centos 7/8. Remove `default` keyword for namedtuple that introduced in python3.7 and `--directory` parameter for `http.server` and specify the working directory for fork.

## Checklist

  with new changes

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -594,11 +594,11 @@ def fork_crl_http_server(port: int, check_connection: bool=True):
     logging.info("fork HTTP server with port {}".format(port))
 
     http_server_connection = get_crl_http_server_connection_string(port)
-
-    cli_args = ['--bind', '127.0.0.1', '--directory', TEST_TLS_CRL_PATH, str(port)]
+    # use cwd= parameter for Popen instead of --directory parameter to support 3.6 that doesn't accept --directory
+    cli_args = ['--bind', '127.0.0.1', str(port)]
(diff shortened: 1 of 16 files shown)


tests/test.py
```diff
@@ -807,11 +807,26 @@ def tearDownModule():
     kill_certificate_validation_services()
 
 
-ConnectionArgs = collections.namedtuple("ConnectionArgs",
-    field_names=["user", "password", "host", "port", "dbname",
-                 "ssl_ca", "ssl_key", "ssl_cert", "format"],
-    # 'format' is optional, other fields are required.
-    defaults=[None])
+if sys.version_info[1] > 6:
```
iamnotacake, line 810: So modern Python `>=3.7` requires this new way of creating database connection args, but to make sure it works on 3.6 as well you added this `class` thing?

iamnotacake (approved):
Looks good, many places where new API methods are used, more tests added as well

tests/test.py
Lagovas: this type is our and not related to python or DBs actually. Namedtuple same as class, but easier to declare. But <=3.6 doesn't support default values. So we use more verbose type description than namedtuple

PR #510, opened 2022-03-11
Ignore legacy keys when caching keys on startup

iamnotacake:
When AcraServer/AcraTranslator starts and loads all the keys into cache (if the option is enabled), ignore old keys related to AcraConnector.
Add new key kind "legacy" and ignore this key if found inside keystore.

## Checklist

  with new changes~
- [x] ~CHANGELOG_DEV.md is updated~

### DIFF
--- a/keystore/filesystem/server_keystore.go
+++ b/keystore/filesystem/server_keystore.go
@@ -809,6 +809,10 @@ func (store *KeyStore) describeDir(dirName string) ([]keystore.KeyDescription, e
 		if err != nil {
 			return nil, err
 		}
+		if description.Purpose == PurposeLegacy {
+			log.WithField("ID", description.ID).Debug("Ignoring legacy key")
+			continue
+		}
 		keys = append(keys, *description)
 	}
 	return keys, nil
@@ -836,6 +840,19 @@ func (store *KeyStore) DescribeKeyFile(fileInfo os.FileInfo) (*keystore.KeyDescr
 	}
 
 	components := strings.Split(fileInfo.Name(), "_")
+
+	if len(components) == 1 {
+		id := fileInfo.Name()
+		if strings.HasSuffix(id, ".pub") {
+			id = id[:len(id)-4]
+		}
+
+		return &keystore.KeyDescription{
+			ID:       id,
+			Purpose:  PurposeLegacy,
+			ClientID: []byte(components[0]),
+		}, nil
+	}
 
 	//in case of one split result slice will have more than one element
 	if len(components) < 2 {
@@ -906,6 +923,14 @@ func (store *KeyStore) DescribeKeyFile(fileInfo os.FileInfo) (*keystore.KeyDescr
 			ID:       fileInfo.Name(),
 			Purpose:  PurposeAuditLog,
 			ClientID: []byte(strings.Join(components[:len(components)-2], "_")),
+		}, nil
+	}
+
+	if lastKeyPart == "server" || lastKeyPart == "server.pub" || lastKeyPart == "translator" || lastKeyPart == "translator.pub" {
+		return &keystore.KeyDescription{
+			ID:       fileInfo.Name(),
+			Purpose:  PurposeLegacy,
+			ClientID: []byte(components[0]),
 		}, nil
 	}
 
--- a/keystore/filesystem/server_keystore.go
+++ b/keystore/filesystem/server_keystore.go
@@ -810,7 +810,7 @@ func (store *KeyStore) describeDir(dirName string) ([]keystore.KeyDescription, e
 			return nil, err
 		}
 		if description.Purpose == PurposeLegacy {
-			log.WithField("ID", description.ID).Debug("Ignoring legacy key")
+			log.WithField("ID", description.ID).Warn("Ignoring legacy key")
 			continue
 		}
 		keys = append(keys, *description)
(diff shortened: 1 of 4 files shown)


Lagovas:
TODO: 
* add integration tests that checks that acra works correctly with old keys and logs about existing legacy keys.

keystore/filesystem/server_keystore.go
```diff
@@ -837,6 +841,19 @@ func (store *KeyStore) DescribeKeyFile(fileInfo os.FileInfo) (*keystore.KeyDescr
 
 	components := strings.Split(fileInfo.Name(), "_")
 
+	if len(components) == 1 {
+		id := fileInfo.Name()
+		if strings.HasSuffix(id, ".pub") {
+			id = id[:len(id)-4]
```
ZhmakaAS, line 847: Is it save enough operation? Do not we need additional id length checking?

keystore/filesystem/server_keystore.go
```diff
@@ -909,6 +926,14 @@ func (store *KeyStore) DescribeKeyFile(fileInfo os.FileInfo) (*keystore.KeyDescr
 		}, nil
 	}
 
+	if lastKeyPart == "server" || lastKeyPart == "server.pub" || lastKeyPart == "translator" || lastKeyPart == "translator.pub" {
```
ZhmakaAS, line 929: Correct me if I am wrong. 

Here, you are checking that if fileInfo contains `.pub` suffix and you truncate it, right?
https://github.com/cossacklabs/acra/pull/510/files#diff-ca8937b8ce39cdc821c3d9b826aecd30d40cc5beb2aec26ea11497f39e13658fR847

So why are you checking scenarios that `lastKeyPart == "server.pub"` or `lastKeyPart == "translator.pub"`. Is it not enough to just check ` lastKeyPart == "server" || lastKeyPart == "translator"`

keystore/filesystem/server_keystore.go
ZhmakaAS, line 847: I think you can `strings.TrimSuffix()` here even without checking on `.pub` existence.

keystore/filesystem/server_keystore.go
iamnotacake: See these lines above
```golang
components := strings.Split(fileInfo.Name(), "_") // note the separator
// ...
lastKeyPart := components[len(components)-1]
```
And checked while debugging that `lastKeyPart` really contains `server.pub` or `translator.pub` when corresponding files exist.

tests/test.py
```diff
@@ -6173,6 +6165,41 @@ def testServerRead(self):
...
+
+    def checkSkip(self):
+        if KEYSTORE_VERSION != 'v1':
+            self.skipTest("test only for keystore v1")
+
+    def setUp(self):
+        super().setUp()
+
+        def create_key_file(name):
+            open(f"{KEYS_FOLDER.name}/{name}", "w").close()
```
ZhmakaAS, line 6182: Do we really want to create some test oriented key files inside general `KEYS_FOLDER` directory. Probably it would be better to create some temporary directory and drop it when test is finished?

tests/test.py
```diff
@@ -6173,6 +6165,41 @@ def testServerRead(self):
...
+
+        try:
+            create_key_file('testclientid')
```
Lagovas, line 6185: with such test's template we test only `testclientid` key because of acra will fail on first error.
But we should be sure that we can work with each of these keys. So let's move all these logic from setUp to test function and iterate over all these types:
```
def setUp(self):
 # do nothing, we don't need acra process, engines and all other stuff
  pass

def testKeysCachedSuccessfully(self):
  with tempfile.TemporaryDirectory() as folder:
     for key in ('testclientid.pub', testclientid_server', ...):
        with self.assertRaises(Exception) as exc:
          create_key_file(key)
          log_file = tempfile.TemporaryFile()
          args = {'keystore_cache_size': 0, 'keystore_cache_on_start_enable': 'true', 'log_to_file': log_file.name, 'keys_dir': folder}
          self.fork_acra(**args)

      # use correct sub-field for msg and exception, it's only example
      self.assertEqual(exc.exception.msg, 'can't wait connection')
      with open(log_file.name, 'r') as f:
        logs = f.read()
      self.assertIn("Cached keystore on start successfully".lower(), logs)
      # remove invalid key to not raise same error at next iteration
      os.remove(key)
```

With such test's template we check all these key types and verify them separately.

Lagovas: oops, my fault. misunderstood that we check correct startup with all of these keys... but my comment about opposite case) nevermind

tests/test.py
```diff
@@ -6173,6 +6165,58 @@ def testServerRead(self):
...
+        try:
+            for key_file in self.legacy_key_files:
+                open(f"{KEYS_FOLDER.name}/{key_file}", "w").close()
+        except:
+            self.tearDown()
+            raise
+
+    def tearDown(self):
+        for key_file in self.legacy_key_files:
+            os.remove(f"{KEYS_FOLDER.name}/{key_file}")
```
Lagovas, line 6202:
```
$ python3 -c 'import os; os.remove("/tlqkeqwleqwlekqweq");'
  File "<string>", line 1, in <module>
FileNotFoundError: [Errno 2] No such file or directory: '/tlqkeqwleqwlekqweq'
```

we should catch exceptions here too or we may fail on first file and leave all remaining files

iamnotacake: Sure, wrapped with `try: ... except: pass`

Lagovas (approved):
lgtm

PR #511, opened 2022-03-11
Interpret uknown startup message as an error

G1gg1L3s:
Postgres has two types of messages:
- Most of them begin with a one-byte id and 4-byte length. We call them _general messages_.
- But there are couple, which begins with a 4-byte length and 4-byte magic number. We call them _startup messages_ because they should come first after the connection is established.

This PR fixes a couple of issues with postgres packet parser, including:
- Now, the `PacketHandler` registers, if the connection is new, so it can distinguish witch packet it expects: startup or general.
- If a startup message has unknown format, the handler raises an error.
- If a general message has unknown format, it is parsed anyway.
- Couple of bugs with the parser are fixed.
- New type of startup packet is added: [GSSENCRequest](https://www.postgresql.org/docs/current/protocol-message-formats.html).

## Checklist

  with new changes~


G1gg1L3s:
Also, I just found another way to break the parser:
The `StartupMessage` packet of postgres protocol is a little special: it's the only packet that has variable length size. It's structure is:
```
[4-byte length] [4-byte id] [payload...]
```

But, what if in some case the length would begin with `0x44` or `'D'` in ascii table. It would match against `DataRow` packet, not the `StartupMessage`.

[Here is](https://www.postgresql.org/docs/current/protocol-message-formats.html) the documentation that describes the protocol.

But, after spending more than an hour, I found that postgres doesn't allow `StartupMessage` lengths more than [10000 bytes](https://github.com/postgres/postgres/blob/641f3dffcdf1c7378cfb94c98b6642793181d6db/src/include/libpq/pqcomm.h#L146), so I guess it's ok.

Or we can change the parser to first check for special messages and then for one-byte headers.

G1gg1L3s:
By the way, a lot of packets have length field, which is 4-byte integer. Acra doesn't limit this field, so in theory, if someone can send packets to acra, he can craft messages with a huge value in length field, which will force Acra to make large allocations (up to 4GB per packet), which can result in denial of service attack.

Maybe we assume that the clients, that have access to the Acra, are trusted, so this attack may not be the case.

Lagovas:
> But, after spending more than an hour, I found that postgres doesn't allow StartupMessage lengths more than [10000 bytes](https://github.com/postgres/postgres/blob/641f3dffcdf1c7378cfb94c98b6642793181d6db/src/include/libpq/pqcomm.h#L146), so I guess it's ok.

Great searching) I think we should limit startup package in same manner.

> Or we can change the parser to first check for special messages and then for one-byte headers.

yes, it's main goal of the task and we should filter at least  first invalid packets until recognizing any of start packets like startup/ssl.

> By the way, a lot of packets have length field, which is 4-byte integer. Acra doesn't limit this field, so in theory, if someone can send packets to acra, he can craft messages with a huge value in length field, which will force Acra to make large allocations (up to 4GB per packet), which can result in denial of service attack.

in theory yes, but it's a database that can be legally used to fetch/upload 4GB data and we can't check it on our side. If we will limit it on our side then we will limit DBs features. 
In the future we can try to catch or even request DBs configuration to understand how destination DB is configured and how long packets it accepts from clients.  But it's not for now.

decryptor/postgresql/packet_handler.go
```diff
@@ -447,15 +458,45 @@ func (packet *PacketHandler) ReadClientPacket() error {
 		We read first 5 bytes, check is there known message types. If not then we try to read 3 more bytes and recognize 3 special messages
 		by their values. If we don't recognize, then process it as general message. We may have error if it's unknown message with message length < 8 bytes when
 		we not recognize, try to read +3 bytes and will block on system call read to read more when message may have only 1 bytes of MessageType and minimal MessageLength = 4 bytes (itself)
+		Overall, as of today (PostgreSQL 14), the protocol supports following packets, that can be
```
Lagovas, line 461: will be great to update whole comment block) because above we mention 3 packets)

decryptor/postgresql/packet_handler.go
```diff
@@ -427,100 +444,159 @@ const WithoutMessageType = 0
...
+	n, err := io.ReadFull(packet.reader, packetBuf)
+	if err := base.CheckReadWrite(n, 8, err); err != nil {
+		packet.logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorDecryptorReadPacket).WithError(err).Debugln("Can't read 8 bytes of Startup/SSLRequest packet")
+		return err
+	}
+
+	switch {
+	case bytes.Equal(StartupRequest, packetBuf[4:8]):
+	case bytes.Equal(SSLRequestHeader, packetBuf[:8]):
+	case bytes.Equal(CancelRequestHeader, packetBuf[:8]):
```
Lagovas, line 533: actually Cancel request is not as general, but it cannot be first. And we should parse it in general process flow.

decryptor/postgresql/packet_handler.go
```diff
@@ -427,100 +444,159 @@ const WithoutMessageType = 0
...
+	// we read 4 bytes before. decrease before call readData because it read exactly as dataLength
+	packet.dataLength -= 4
+
+	if err := packet.readData(false); err != nil {
+		return err
+	}
+	// restore correct value
+	packet.dataLength += 4
+
+	packet.started = true
```
Lagovas, line 557: https://github.com/cossacklabs/acra/blob/0.92.0/decryptor/postgresql/pg_decryptor.go#L560

we should change this flag here too. When app connects without TLS, asks TLS and postgresql accepts it, then they establish TLS connection and sends startup message again. 
hm, maybe we should add integration test that will connect like that. When tries without TLS but works with TLS too. 
to do it app should use `sslmode=allow`

Lagovas: also you can check manually that it fails. try to connect with psql and such sslmode through acra to postgresql that configured to use ssl. you can use our image used in postgresql-job in `.circleci/config.yml` and certificates in `tests/ssl/acra-client/*` and `tests/ssl/ca/ca.crt`

decryptor/postgresql/packet_handler.go
G1gg1L3s: From the [docs](https://www.postgresql.org/docs/current/protocol-flow.html#id-1.10.5.7.9):
> During the processing of a query, the frontend might request cancellation of the query. The cancel request is not sent directly on the open connection to the backend for reasons of implementation efficiency: we don't want to have the backend constantly checking for new input from the frontend during query processing. Cancel requests should be relatively infrequent, so we make them slightly cumbersome in order to avoid a penalty in the normal case.
To issue a cancel request, the frontend opens a new connection to the server and sends a CancelRequest message, rather than the StartupMessage message that would ordinarily be sent across a new connection. The server will process this request and then close the connection. For security reasons, no direct reply is made to the cancel request message.

So, I thought it's one of the _startup_ messages. That also explains why `CancelRequest` doesn't have general format and uses the same format as the `StartupMessage`.

Lagovas: yes, but it is during processing a query. But query they can process only after Startup/SSL and authentication

decryptor/postgresql/packet_handler.go
Lagovas: yeah, you are right. missed that packetHandler responsible for parsing and processing first packet now. Last note, can you remove redundant `firstByte` variable? and it will be enough)

PR #515, opened 2022-03-21
Transparent decryption with metadata replacing

Lagovas:
* Added processing `DataRowDescription` and `ParameterDescriptionData`. In `DataRowDescription` now we replace `OID` values that tell drivers which data types will be returned. In `ParameterDescriptionData` we replace `OID`s for parameters in prepared statements. To tell apps that they should pass string value instead of binary (that will be changed on Acra side)
* Extend encryptor config with two new fields: `default_data_value` and `data_type`. Last values generated via protobuf in similar way as token type. 
* Added support of text format values in binary protocols (prepared statements). Previously was supported only binary formatted values.
* Refactored `QueryDataEncryptor` to re-use functions related to parsing placeholders and saving encryptor's settings in context/client session values
* Extended `ClientSession` interface to provide KV interface and store session related data for client/db packet handlers
* Removed legacy `acra-log-verifier` config file from configs folder
* Added updating context (`ctx`) with flag after successful decryption (`base.[MarkDecryptedContext|IsDecryptedFromContext]`). Needs for replacing with default values binary data types that hard to determine when they was untouched
* Removed legacy interface method `SubscribeOnColumnDecryption` and all related methods of proxies due to not using
* Refactored `PgSQLDataEncoderProcessor`. Now it used not only for tokenization, but for decoding all input data from database and encoding before sending to client. Also removed this logic from `PostgresqlDecryptor` which now responsible only for database packet parsing. All other data manipulations moved to OnQuery/OnBind/OnColumn subscribers
* Additionally removed `DecodedData` wrapper for `Column` structs used by decryptor

It's **ready to review**. Probably last changes will be related to encryptor's config validation and updating changelogs (after that I will change it from status `Draft`). So welcome to review.

## Checklist

  with new changes
- [-] CHANGELOG.md is updated (in case of notable or breaking changes)
- ~~[ ] Benchmark results are attached (if applicable)~~
- ~~[ ] [Example projects and code samples] are up-to-date (in case of API changes)~~


decryptor/postgresql/data_encoder.go
```diff
@@ -0,0 +1,255 @@
...
+	return "PgSQLDataEncoderProcessor"
+}
+
+// here we process encryption/tokenization results before send it to a client
+// acra decrypts or de-tokenize SQL literals, so we should convert string SQL literals to binary format
+// if client expects int, then parse INT literals and convert to binary 4/8 byte format
+// if expects bytes, then pass as is
+// if expects string, then leave as is if it is valid string or encode to hex
+// if it is encrypted data then we return default values or as is if applicable (binary data)
+func (p *PgSQLDataEncoderProcessor) encodeBinary(ctx context.Context, data []byte, setting config.ColumnEncryptionSetting, columnInfo base.ColumnInfo, logger *logrus.Entry) (context.Context, []byte, error) {
```
ZhmakaAS, line 58: Seems to be `columnInfo` is unused field here.

ZhmakaAS, line 58: Seems to be `columnInfo` is unused field here.

decryptor/postgresql/data_encoder.go
```diff
@@ -0,0 +1,255 @@
...
+			logger.WithField("size", size).Warningln("Unsupported int value size")
+			return ctx, data, ErrInvalidIntValueBinarySize
+		}
+		return ctx, newData, nil
+	}
+
+	return ctx, data, nil
+}
+
+func (p *PgSQLDataEncoderProcessor) decodeBinary(ctx context.Context, data []byte, setting config.ColumnEncryptionSetting, columnInfo base.ColumnInfo, logger *logrus.Entry) (context.Context, []byte, error) {
```
ZhmakaAS, line 121: And here the same I think. Correct me if I am wrong

ZhmakaAS, line 121: And here the same I think. Correct me if I am wrong

decryptor/postgresql/data_encoder.go
```diff
@@ -0,0 +1,255 @@
...
+	case common2.EncryptedType_Int32, common2.EncryptedType_Int64:
+		size := 8
+		if setting.GetEncryptedDataType() == common2.EncryptedType_Int32 {
+			size = 4
+		}
+		// convert back from text to binary
+		value, err := strconv.ParseInt(string(data), 10, 64)
+		// we don't return error to not cause connection drop on Acra side and pass it to app to deal with it
+		if err != nil {
+			if setting.GetDefaultDataValue() != nil {
```
ZhmakaAS, line 92:
```suggestion
			if setting.GetDefaultDataValue() != nil {
```
```suggestion
			if newVal := setting.GetDefaultDataValue(); newVal != nil {
```

ZhmakaAS, line 92:
```suggestion
			if setting.GetDefaultDataValue() != nil {
```
```suggestion
			if newVal := setting.GetDefaultDataValue(); newVal != nil {
```

ZhmakaAS, line 92:
```suggestion
			if newVal := setting.GetDefaultDataValue(); newVal != nil {
```

ZhmakaAS, line 92:
```suggestion
			if newVal := setting.GetDefaultDataValue(); newVal != nil {
```

decryptor/postgresql/data_encoder.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+		newData := make([]byte, size)
+		switch size {
+		case 4:
+			binary.BigEndian.PutUint32(newData, uint32(value))
+			break
+		case 8:
+			binary.BigEndian.PutUint64(newData, uint64(value))
+			break
+		default:
```
ZhmakaAS, line 111: I think, there is no need of default block because size is our local variable that could be only 4 or 8, right?

ZhmakaAS, line 111: I think, there is no need of default block because size is our local variable that could be only 4 or 8, right?

decryptor/postgresql/data_encoder.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	logger := logging.GetLoggerFromContext(ctx)
+	columnInfo, ok := base.ColumnInfoFromContext(ctx)
+	if !ok {
+		logger.WithField("processor", "PgSQLDataEncoderProcessor").Warningln("No column info in ctx")
+		// we can't do anything
+		return ctx, data, nil
+	}
+
+	if p.mode == DataEncoderModeEncode {
```
ZhmakaAS, line 242: Do you think to split this logic into two different parts?

I mean something like this.

```

type BasePgSQLDataEncoderProcessor struct {}

func (p *BasePgSQLDataEncoderProcessor) encodeBinary()

func (p *BasePgSQLDataEncoderProcessor) decodeBinary()

func (p *BasePgSQLDataEncoderProcessor) encodeText()

func (p *BasePgSQLDataEncoderProcessor) decodeText()





type DecodePgSQLDataEncoderProcessor struct {
     BasePgSQLDataEncoderProcessor
}
// with it own decode logic
func (p *DecodePgSQLDataEncoderProcessor) OnColumn()



type EncodePgSQLDataEncoderProcessor struct {
     BasePgSQLDataEncoderProcessor
}

//with it own encode logic
func (p *EncodePgSQLDataEncoderProcessor) OnColumn()
```

We can decouple this logic from one place which can be easier to support in future. What do you think?

ZhmakaAS, line 242: Do you think to split this logic into two different parts?

I mean something like this.

```

type BasePgSQLDataEncoderProcessor struct {}

func (p *BasePgSQLDataEncoderProcessor) encodeBinary()

func (p *BasePgSQLDataEncoderProcessor) decodeBinary()

func (p *BasePgSQLDataEncoderProcessor) encodeText()

func (p *BasePgSQLDataEncoderProcessor) decodeText()





type DecodePgSQLDataEncoderProcessor struct {
     BasePgSQLDataEncoderProcessor
}
// with it own decode logic
func (p *DecodePgSQLDataEncoderProcessor) OnColumn()



type EncodePgSQLDataEncoderProcessor struct {
     BasePgSQLDataEncoderProcessor
}

//with it own encode logic
func (p *EncodePgSQLDataEncoderProcessor) OnColumn()
```

We can decouple this logic from one place which can be easier to support in future. What do you think?

decryptor/postgresql/data_encoder.go
Lagovas: yes. but contains metadata about columns and left because may be left for future use. It wasn't added in this PR, it already exists. I suggest to review need of this struct at finish of this feature. And if it will be redundant then remove.

decryptor/postgresql/data_encoder.go
Lagovas: separated into 2 components

decryptor/postgresql/data_encoder.go
Lagovas: removed

decryptor/postgresql/data_encoder.go
Lagovas: agree, updated

tests/random_utils.py
```diff
@@ -21,6 +21,9 @@ def random_int64():
 
 
 def random_bytes(n=100):
+    # if default size then return binary and printable data
+    if n == 100:
+        return bytes([i for i in range(n)])
```
G1gg1L3s, line 26: Interesting decision :) Maybe extract it into a separate function?

decryptor/postgresql/proxy.go
```diff
@@ -65,7 +65,7 @@ func (factory *proxyFactory) New(clientID []byte, clientSession base.ClientSessi
 	schemaStore := factory.setting.TableSchemaStore()
 	storeMask := schemaStore.GetGlobalSettingsMask()
 	// register only if masking/tokenization/searching will be used
-	if storeMask&(config.SettingSearchFlag|config.SettingMaskingFlag|config.SettingTokenizationFlag) > 0 {
+	if storeMask&(config.SettingSearchFlag|config.SettingMaskingFlag|config.SettingTokenizationFlag|config.SettingDefaultDataValueFlag|config.SettingDataTypeFlag) > 0 {
```
G1gg1L3s, line 68: I would say that it's time to extract this `|` chain into a variable :)

encryptor/queryDataEncryptor.go
```diff
@@ -95,8 +115,8 @@ func (encryptor *QueryDataEncryptor) encryptInsertQuery(ctx context.Context, ins
 			for _, valTuple := range rows {
 				// collect values per column
 				for j, value := range valTuple {
-					columnName := columnsName[j]
-					if changedValue, err := encryptor.encryptExpression(ctx, value, schema, columnName); err != nil {
+					columnName := columnsName[j%len(columnsName)]
```
G1gg1L3s, line 118: Why is there a modulo operation?

G1gg1L3s: In the vscode plugin that I use there is a support of additional diagnostics about gc-escapes and small optimizations that guarantee to apply. And when I change this to just `j`, this what the vscode shows:
image

That means the compiler was able to deduce that `j` is in bounds of the `columnsName` slice.

G1gg1L3s (commented):
And a few comments and wishes about git history: I think it would be better to split this PR into multiple ones. For example, each item of changes in the description could easily get its own PR. I don't mind if these small PRs would contain only change of a single line of code, because it would greatly improve the quality and speed of reviews: it's easier to separate changes and think why they are important on their own. Also, I like reading not only the final changes, but all commits one by one, to understand the logic about how and reasons why something is changing. But it was hard, because the commits often contain multiple changes at once, or are the parts of other commits. Very often they, beside some changes, they contain formatting fixes or commented code. For formatting I would recommend setup something like `format on save` or `format on commit` or something similar. Also I recommend reading about best practices about commit messages https://cbea.ms/git-commit

P.S. Sorry if some comments are outdated. They were written during the review of commits one by one, so some changes could happen in the next commits, that weren't reviewed yet.

tests/random_utils.py
Lagovas: when I tested locally I found that tests failed only when data contained `\` symbol with value `92` in ascii table. So I decided to include it for all binary values used by default (`random_bytes` function without parameters). Additionally, it includes printable [32; 127] values and not printable [0, 31]. If move it into a separate function then it requires updating all functions that use `random_bytes` function and decide which one needs it and which don't. But better to test such cases by default, imho.

utils/utils.go
```diff
@@ -224,3 +226,15 @@ func WaitWithTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
...
+// BytesToString converts data to string with re-using same allocated memory
+// Warning: data shouldn't be changed after that because it will cause runtime error due to strings are immutable
+// Only for read/iterate operations
+func BytesToString(data []byte) string {
+	if len(data) == 0 {
+		return ""
+	}
+	sliceHeader := (*reflect.SliceHeader)(unsafe.Pointer(&data))
+	stringHeader := reflect.StringHeader{Data: sliceHeader.Data, Len: sliceHeader.Len}
+	return *(*string)(unsafe.Pointer(&stringHeader))
```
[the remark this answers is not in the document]
Lagovas: thanks for the post, I will change it into fasthttp format.
Anyway, all these functions should be used carefully and only when writer knows what he do. 
At this moment it was used only for the short operations and in context when the reference owner leaves longer than temporary casted value. But yes, better to refactor it into the format when we will be in safe from GC.

utils/utils.go
```diff
@@ -224,3 +226,15 @@ func WaitWithTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
...
+}
```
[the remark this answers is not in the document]
Lagovas: yes. because strings are immutable and it's easier to copy them as separate immutable value.

Lagovas (commented):
> And a few comments and wishes about git history: I think it would be better to split this PR into multiple ones. For example, each item of changes in the description could easily get its own PR. I don't mind if these small PRs would contain only change of a single line of code, because it would greatly improve the quality and speed of reviews: it's easier to separate changes and think why they are important on their own. Also, I like reading not only the final changes, but all commits one by one, to understand the logic about how and reasons why something is changing. But it was hard, because the commits often contain multiple changes at once, or are the parts of other commits. Very often they, beside some changes, they contain formatting fixes or commented code. For formatting I would recommend setup something like format on save or format on commit or something similar. Also I recommend reading about best practices about commit messages https://cbea.ms/git-commit
> 
> P.S. Sorry if some comments are outdated. They were written during the review of commits one by one, so some changes could happen in the next commits, that weren't reviewed yet.

agree with you, will try it in next my PRs. This feature wasn't clear at the start and a lot of things were changed during the development due to existing of a lot of components linked with each other. but looking at now I see that there a lot of pieces that could be splitted into separate PRs easily *(
you prefer that I will fix current comments, commit, wait approval and after that split it into the separate PRs and review one more time or we can finish it as is and keep it in mind in the future?)
p.s. about auto-format. I already have it. But it has a small sleep before applying (to not format on every new character) and sometimes when I switch into the next tab/window after changes my IDE fails to format code) So, I will add it to pre-commit hooks to avoid such problems.

encryptor/config/encryptionSettings.go
```diff
@@ -153,6 +153,15 @@ type BasicColumnEncryptionSetting struct {
 	settingMask              SettingMask
 }
 
+// IsBinaryDataOperation return true if setting related to operation over binary data
+func IsBinaryDataOperation(setting ColumnEncryptionSetting) bool {
+	// tokenization for binary data or encryption/masking of binary data (not text)
+	hasBinaryOperation := setting.GetTokenType() == common.TokenType_Bytes
+	hasBinaryOperation = hasBinaryOperation || setting.OnlyEncryption() || setting.IsSearchable()
+	hasBinaryOperation = hasBinaryOperation || len(setting.GetMaskingPattern()) != 0
```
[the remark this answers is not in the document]
G1gg1L3s: О.о

G1gg1L3s:
> you prefer that I will fix current comments, commit, wait approval and after that split it into the separate PRs and review one more time or we can finish it as is and keep it in mind in the future?)

I think we can left this PR as is :)

decryptor/postgresql/data_encoder.go
```diff
@@ -0,0 +1,249 @@
...
+		// convert back from text to binary
+		value, err := strconv.ParseInt(string(data), 10, 64)
+		// we don't return error to not cause connection drop on Acra side and pass it to app to deal with it
+		if err != nil {
+			if newVal := setting.GetDefaultDataValue(); newVal != nil {
+				value, err = strconv.ParseInt(*newVal, 10, 64)
+				if err != nil {
+					logger.WithError(err).Errorln("Can't parse default integer value")
+					return ctx, data, err
+				}
```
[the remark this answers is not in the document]
Lagovas: hm, reviewed these statements. I will add such validations but anyway value stored as string and we should parse it or change API for setting. For example return `interface{}` instead of string and make type casts. But anyway to be in safe we should check every cast even if we validate the config at the start. Or we can provide several methods like `GetDefaultValueString`/`GetDefaultValueInt32`/etc and internally parse once at validation stage and make such type casts internally. And extend current base inteface for `ColumnEncryptionSetting`. For now I'm not sure that it will be useful and will not complicate future extensions. I prefer to leave it as is for now and back again after several iterations and better understanding how it can be improved or what new types we want to support.

encryptor/queryDataEncryptor.go
Lagovas: we use module because there should be not more than declared columns count:
`insert into table1(col1, col2, col3) values (val1, val2, val3), (val4, val5, val6);`
There is we have 3 columns and 3 tuples. Every tuple can have 3 values. If more than it's invalid behavior and we can skip it and pass as is to database. Because database will deny such query. But in valid case we are safe and map value to correct column.

encryptor/config/encryptionSettings.go
```diff
@@ -319,3 +412,17 @@ func (s *BasicColumnEncryptionSetting) applyDefaults(defaults defaultValues) {
 		}
 	}
 }
+
+// HasTypeAwareSupport return true if setting configured for decryption with type awareness
+func HasTypeAwareSupport(setting ColumnEncryptionSetting) bool {
+	maskingSupport := setting.GetMaskingPattern() != ""
+	switch setting.GetEncryptedDataType() {
+	//case common2.EncryptedType_String, common2.EncryptedType_Bytes:
```
G1gg1L3s, line 420: Something is left :)

masking/common/patterns.go
```diff
@@ -45,5 +49,13 @@ func ValidateMaskingParams(pattern string, plaintextLength int, plaintextSide Pl
 	if plaintextSide != PlainTextSideRight && plaintextSide != PlainTextSideLeft {
 		return ErrInvalidPlaintextSide
 	}
+	switch dataType {
+	//case common2.EncryptedType_String, common2.EncryptedType_Bytes:
```
G1gg1L3s, line 53: And here too

encryptor/queryDataEncryptor.go
G1gg1L3s: Ok, this is reasonable. But in my opinion, it would be better to just check that if `j > len(columnNames)` then just skip this iteration with `continue`. At least we will not spend time encrypting extra values.

G1gg1L3s (approved):
Beside one comment, everything is cool!

decryptor/postgresql/data_encoder.go
```diff
@@ -0,0 +1,249 @@
...
+			}
+		}
+		return ctx, data, nil
+	case common2.EncryptedType_Int32, common2.EncryptedType_Int64:
+		size := 8
+		if setting.GetEncryptedDataType() == common2.EncryptedType_Int32 {
+			size = 4
+		}
+		// convert back from text to binary
+		value, err := strconv.ParseInt(string(data), 10, 64)
```
G1gg1L3s, line 65: Small question: in all other cases we first check that value is indeed decrypted (by calling `base.IsDecryptedFromContext(ctx)`), and if not, we fallback to returning default values. But in this case we check decryption by parsing the number. Does it have some special meaning or reason?

Lagovas: the main reason is that any encrypted blob has size > len('9223372036854775807') = 19. Our acrablock has 90+ bytes of overhead. Acrastructs even more. So no one supported ciphertext cannot be valid integer string literal. So there is enough to check that it's valid valid integer string literal.

PR #516, opened 2022-03-23
Don't generate poison keys during the query

G1gg1L3s:
Right now every call to the `GetPoison*` creates keys, if they don't exist. These results in a superfluous side effect, unexpected behaviour and delays in decryption (because we try to check records with fresh keys, which obviously do nothing useful).

Change this, by:
- Introducing new interface, which is responsible for poison key generation
- Removing creation of keys in `Get` methods

Also, keep this behaviour for poisonrecordmaker, but do this explicitly.

## Checklist

  with new changes: [here](https://github.com/cossacklabs/product-docs/pull/239).


Lagovas (approved):
lgtm

PR #517, opened 2022-03-29
MySQL encoding data types

ZhmakaAS:
Basic implementation and code refactoring regarding data type encoding in MySQL based on @Lagovas PR for PostgreSQL.

## Checklist

  with new changes


decryptor/mysql/data_encoder.go
```diff
@@ -0,0 +1,344 @@
...
+		encoded := make([]byte, 4)
+		intValue, err := strconv.ParseInt(string(data), 10, 32)
+		if err != nil {
+			if newVal := setting.GetDefaultDataValue(); newVal != nil {
+				intValue, err = strconv.ParseInt(*newVal, 10, 64)
+				if err != nil {
+					return data, false, err
+				}
+			}
+			// TODO: ask Lagovas what to do in case of error casting to defined type.
```
Lagovas, line 186: do nothing. return value inappropriate to type as it would be with encrypted binary data in case of expected integer.
at the future we will return DB specific error using wire protocol error notifications.

decryptor/mysql/response_proxy.go
```diff
@@ -531,8 +525,8 @@ func (handler *Handler) processTextDataRow(ctx context.Context, rowData []byte,
 
 func (handler *Handler) processBinaryDataRow(ctx context.Context, rowData []byte, fields []*ColumnDescription) ([]byte, error) {
 	pos := 0
-	var n int
-	var err error
+	//var n int
+	//var err error
```
Lagovas, line 529: dont' forget to remove

decryptor/mysql/prepared_statements.go
```diff
@@ -187,7 +187,11 @@ func (m *mysqlBoundValue) Copy() base.BoundValue {
 
 // SetData set new value to BoundValue using ColumnEncryptionSetting if provided
 func (m *mysqlBoundValue) SetData(newData []byte, setting config.ColumnEncryptionSetting) error {
-	m.textData = newData
+	// means that we set encrypted data
+	if !bytes.Equal(m.textData, newData) {
+		m.paramType = TypeBlob
+		m.textData = newData
```
Lagovas, line 193: hm, so now it's not only `textData` but binary too. maybe we should rename `textData` to `data`?

decryptor/mysql/prepared_statements.go
```diff
@@ -187,7 +187,11 @@ func (m *mysqlBoundValue) Copy() base.BoundValue {
 
 // SetData set new value to BoundValue using ColumnEncryptionSetting if provided
 func (m *mysqlBoundValue) SetData(newData []byte, setting config.ColumnEncryptionSetting) error {
-	m.textData = newData
+	// means that we set encrypted data
```
Lagovas, line 190: hm... for now it is used by QueryEncryptor uses it after encryption/tokenization. but what about if some else component will use it and there will be not only encrypted? can we somehow check that it is encrypted and we should change type?

decryptor/mysql/response_proxy.go
```diff
@@ -560,251 +554,65 @@ func (handler *Handler) processBinaryDataRow(ctx context.Context, rowData []byte
...
+}
 
-	// Now show the value to the subscribers. Note that they might change it.
-	value, err = handler.onColumnDecryption(ctx, columnIndex, value)
-	if err != nil {
-		handler.logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorGeneral).
-			WithError(err).WithField("field_index", columnIndex).
-			Errorln("Failed to process column data")
-		return nil, err
+func (handler *Handler) extractProcessData(pos int, rowData []byte, field *ColumnDescription) ([]byte, int, error) {
```
Lagovas, line 577: maybe name it just `extractData`? because two verbs a bit discourage

decryptor/mysql/response_proxy.go
```diff
@@ -835,18 +641,18 @@ func (handler *Handler) QueryResponseHandler(ctx context.Context, packet *Packet
...
 						handler.logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorProtocolProcessing).Errorln("EOF and field count != current row packet count")
 						return ErrMalformPacket
 					}
+					output = append(output, fieldPacket)
 					break
 				}
 			}
 			handler.logger.WithField("column_index", i).Debugln("Parse field")
-			field, err := ParseResultField(fieldPacket.GetData())
+			field, err := ParseResultField(fieldPacket, handler.setting.TableSchemaStore())
```
Lagovas, line 655: what about to split Parsing and updating type of field? or rename to something obvious that show the fact that field's type will be changed?

decryptor/mysql/data_encoder.go
```diff
@@ -0,0 +1,364 @@
...
+	"github.com/cossacklabs/acra/encryptor/config/common"
+	"github.com/cossacklabs/acra/logging"
+	"github.com/sirupsen/logrus"
+)
+
+// ErrConvertToDataType error that indicates if data type conversion was failed
+var ErrConvertToDataType = errors.New("error on converting to data type")
+
+// BaseMySQLDataEncoderProcessor implements processor and encode/decode binary intX values to text format which acceptable by Tokenizer
+type BaseMySQLDataEncoderProcessor struct{}
```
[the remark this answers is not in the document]
ZhmakaAS: Agree!

decryptor/mysql/data_encoder.go
ZhmakaAS: Discussed to implement rollaback type logic

decryptor/mysql/response_proxy.go
```diff
@@ -515,13 +502,13 @@ func (handler *Handler) processTextDataRow(ctx context.Context, rowData []byte,
 		if err != nil {
 			return nil, err
 		}
-		value, err = handler.onColumnDecryption(ctx, i, value)
+		_, value, err = handler.onColumnDecryption(ctx, i, value, false, fields[i])
```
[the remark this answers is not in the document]
ZhmakaAS: The same question - we are not rolling back types in case of conversion failure for text protocol

decryptor/mysql/response_proxy.go
ZhmakaAS: Agree to rename. It looks like confusion

decryptor/mysql/data_encoder.go
```diff
@@ -0,0 +1,388 @@
...
+func (p *BaseMySQLDataProcessor) encodeBinary(ctx context.Context, data []byte, setting config.ColumnEncryptionSetting, columnInfo base.ColumnInfo, logger *logrus.Entry) (context.Context, []byte, error) {
+	logger = logger.WithField("column", setting.ColumnName()).WithField("decrypted", base.IsDecryptedFromContext(ctx))
+	logger.Debugln("Encode binary")
+
+	if len(data) == 0 {
+		// we still need to encode result data as it might be null field in db
+		return ctx, PutLengthEncodedString(data), nil
+	}
+
+	var columnType = columnInfo.DataBinaryType()
```
Lagovas, line 74: let's assign this variable a bit below close to the usage block? because read this row and tried to find where it used

decryptor/mysql/data_encoder.go
```diff
@@ -0,0 +1,388 @@
...
+	// The result must have exact same format as it had. Overflows are unacceptable.
+	switch Type(columnType) {
+	case TypeNull:
+		if data != nil {
+			return nil, nil, errors.New("NULL not kept NULL")
+		}
+
+	case TypeTiny:
+		encoded = make([]byte, 1)
+		intValue, err := strconv.ParseInt(string(data), 10, 8)
```
Lagovas, line 103: what about to use [BytesToString](https://github.com/cossacklabs/acra/blob/master/utils/utils.go#L236)?) 
looks like safe place for accessing data without concurrent access during parsing
same below for other int types

decryptor/mysql/data_encoder.go
```diff
@@ -0,0 +1,388 @@
...
+	case TypeNull:
+		// do nothing
+
+	case TypeTiny:
+		var numericValue int8
+		err := binary.Read(bytes.NewReader(encoded), binary.LittleEndian, &numericValue)
+		if err != nil {
+			break
+		}
+		return ctx, []byte(strconv.FormatInt(int64(numericValue), 10)), nil
```
Lagovas, line 303: what about to use [strconv.AppendInt](https://pkg.go.dev/strconv#AppendInt)? It helps avoid extra conversion from `string` to `bytes`:
`strconv.FormatInt(nil, int64(numericValue), 10)`
same below

encryptor/dbDataCoder.go
```diff
@@ -62,15 +63,32 @@ func (*MysqlDBDataCoder) Decode(expr sqlparser.Expr) ([]byte, error) {
...
+	}
+
 	switch val := expr.(type) {
 	case *sqlparser.SQLVal:
 		switch val.Type {
-		case sqlparser.IntVal, sqlparser.StrVal:
-			return data, nil
+		case sqlparser.IntVal:
+			// if data was just tokenized, so we return it as is because it is valid int literal
+			if _, err := strconv.Atoi(string(data)); err == nil {
```
Lagovas, line 79: looks like `BytesToString`

Lagovas:
in general looks great except question about returning error in binary encoding. improvements related to type conversions are questionable)

decryptor/mysql/data_encoder.go
ZhmakaAS: Agree, makes sense.

decryptor/mysql/response_proxy.go
```diff
@@ -835,27 +649,31 @@ func (handler *Handler) QueryResponseHandler(ctx context.Context, packet *Packet
...
 				}
 			}
 			handler.logger.WithField("column_index", i).Debugln("Parse field")
-			field, err := ParseResultField(fieldPacket.GetData())
+			field, err := ParseResultField(fieldPacket)
 			if err != nil {
 				handler.logger.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorProtocolProcessing).WithError(err).Errorln("Can't parse result field")
 				return err
 			}
+			// updating filed type according to DataType provided in schemaStore
```
Lagovas, line 668:
```suggestion
			// updating field type according to DataType provided in schemaStore
```

decryptor/mysql/data_encoder.go
ZhmakaAS: Nice idea!

decryptor/mysql/data_encoder.go
ZhmakaAS: Cool, didnt know about this methods before

PR #519, opened 2022-04-05
Convert 'bytes' data_type to BYTEA oid for postgresql

Lagovas:
Found that `bytes` `data_type` not handled properly for postgresql

## Checklist

  with new changes


Lagovas:
actually not, because by default all data stored in the database as bytea... and DB returns it as bytea anyway. so this change just to be consistent. `bytes` type we support to allow configure `default_data_value` for cases where we want return something meaningful instead of acrablock/acrastruct from app side

G1gg1L3s (approved):
Cool!

ZhmakaAS (approved):
Thanks @Lagovas. Looks good

PR #520, opened 2022-04-05
Set default data type to unknown

Lagovas:
* Use `Unknown` as default `data_type`
* Add unit-tests that verifies that cannot be defined `token_type` together with `data_type`
* move conversation from TokenType to EncryptedType from pseudonymization package to encryptor/config/common that responsible for EncryptedType

## Checklist

  with new changes

### DIFF
--- a/encryptor/config/common/encryptedTypes.go
+++ b/encryptor/config/common/encryptedTypes.go
@@ -53,7 +53,7 @@ var (
 )
 
 // ValidateEncryptedType return true if value is supported EncryptedType
-func ValidateEncryptedType(value EncryptedType, tokenType common.TokenType) error {
+func ValidateEncryptedType(value EncryptedType) error {
 	supported, ok := supportedEncryptedTypes[value]
 	if !ok {
 		return ErrUnknownEncryptedType
--- a/encryptor/config/common/encryptedTypes.go
+++ b/encryptor/config/common/encryptedTypes.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"errors"
+	"github.com/cossacklabs/acra/pseudonymization/common"
 	"strconv"
 	"unicode/utf8"
 )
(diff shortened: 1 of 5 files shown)


encryptor/config/common/encryptedTypes.go
```diff
@@ -52,7 +53,7 @@ var (
 )
 
 // ValidateEncryptedType return true if value is supported EncryptedType
-func ValidateEncryptedType(value EncryptedType) error {
+func ValidateEncryptedType(value EncryptedType, tokenType common.TokenType) error {
```
G1gg1L3s, line 56: `TokenType` is not used. Or is it for the future?

encryptor/config/encryptionSettings.go
```diff
@@ -230,38 +230,43 @@ func (s *BasicColumnEncryptionSetting) Init() (err error) {
 	if s.ReEncryptToAcraBlock != nil && *s.ReEncryptToAcraBlock {
 		s.settingMask |= SettingReEncryptionFlag
 	}
+	var tokenType common.TokenType
+	var ok bool
 	if s.TokenType != "" || s.Tokenized {
-		tokenType, ok := tokenTypeNames[s.TokenType]
+		tokenType, ok = tokenTypeNames[s.TokenType]
```
G1gg1L3s, line 236: Here we define `ok` in scope of the entire function, though it is used only in scope of this `if` statement. And them, the `ok` is overwritten on 297th line. For the reader this is confusing, because my first assumption was that the result of this variable is used somewhere below.

G1gg1L3s, line 236: Here we define `ok` in scope of the entire function, though it is used only in scope of this `if` statement. And then, the `ok` is overwritten on the 297th line. For the reader this is confusing, because my first assumption was that the result of this variable is used somewhere below, and I scroll up and down a couple of times to see where it is used.

Lagovas: there is only `tokenType` used in the block below. But without pre-declaring `ok` variable `tokenType` in the `if` block will be different and don't  re-use tokenType from outside block.

encryptor/config/common/encryptedTypes.go
Lagovas: it's my fault. I tried to validate EncryptedType in pair with TokenType. But then found that we don't allow define token_type with data_type explicitly and this check is redundant. I removed it

ZhmakaAS (approved):
Lgtm

G1gg1L3s (approved):
Cool!

PR #526, opened 2022-04-14
Allow plain connection after ssl deny

G1gg1L3s:
Right now, if some clients receive `SSL deny` from the server, they can try to start plain connection by sending plain `Startup message`. The last one would then be served not as a startup, but as a general one, which will hang a connection.

This PR fixes this by reloading client thread after `ssl deny` to expect startup message again.

There are a couple of `TODO`s, which I hope you will help me to resolve.

## Checklist

- [x] ~Public API has proper documentation in the [Acra documentation] site or has PR on [documentation repository]~
  with new changes

### DIFF
--- a/decryptor/postgresql/pg_decryptor.go
+++ b/decryptor/postgresql/pg_decryptor.go
@@ -431,7 +431,11 @@ func (proxy *PgProxy) stopProxyClientConnection(logger *log.Entry) error {
 		return errors.New("can't stop background goroutine")
 	}
 
-	// TODO: why twice?
+	// Reset the deadline
+	// From the https://pkg.go.dev/net#Conn:
+	//
+	//   A zero value for t means I/O operations will not time out.
+	//
 	if err := proxy.clientConnection.SetDeadline(time.Time{}); err != nil {
 		logger.
 			WithError(err).
--- a/decryptor/postgresql/pg_decryptor.go
+++ b/decryptor/postgresql/pg_decryptor.go
@@ -598,6 +598,14 @@ func (proxy *PgProxy) ProxyDatabaseConnection(ctx context.Context, errCh chan<- 
 			switch {
 			case packetHandler.IsSSLRequestDeny():
 				logger.Debugln("Deny ssl request")
+				// In case of deny ssl, the client can send plain startup message
+				// again. To handle this, we reload client thread to reset the state
+				if err := proxy.stopClientThread(logger); err != nil {
+					errCh <- base.NewDBProxyError(err)
+					return
+				}
+				go proxy.ProxyClientConnection(ctx, errCh)
+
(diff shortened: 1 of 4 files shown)


tests/test.py
```diff
@@ -9400,6 +9400,54 @@ def testClientIDRead(self):
...
+        self.checkSkip()
+        try:
+            if not self.EXTERNAL_ACRA:
+                self.acra = self.fork_acra()
+
+        except:
+            self.tearDown()
+            raise
+
+    def testPlainConnectionAfterDeny(self):
```
Lagovas, line 9429: I suggest to rewrite this function:
```
def testPlainConnectionAfterDeny():
    # declare fully async function to call run_until_complete once. better to separate async and sync code
    async def _testPlainConnectionAfterDeny(self):
      conn = await async.connect(...)
      conn.fetch(...)

    loop = asyncio.new_event_loop()  # create new to avoid concurrent usage of the loop in the current thread and allow parallel execution in the future
    loop.run_until_complete(_testPlainConnectionAfterDeny())

```

decryptor/postgresql/pg_decryptor.go
```diff
@@ -408,6 +408,41 @@ func (proxy *PgProxy) sendClientError(msg string, logger *log.Entry) error {
...
+	case <-proxy.ClientStopResponse:
+	case <-time.NewTimer(ClientStopTimeout).C:
+		logger.
+			// TODO: which event code
+			WithField(logging.FieldKeyEventCode, logging.EventCodeErrorDecryptorCantInitializeTLS).
+			Errorln("Can't stop background goroutine")
+		return errors.New("can't stop background goroutine")
+	}
+
+	// TODO: why twice?
```
Lagovas, line 434: actually I don't remember) maybe to be sure)
or you can remove it and run several tests to check is it break something or not.

decryptor/postgresql/pg_decryptor.go
```diff
@@ -408,6 +408,41 @@ func (proxy *PgProxy) sendClientError(msg string, logger *log.Entry) error {
 	return nil
 }
 
+// stopClientThread sends a signal to a client thread to stop. Returns error in
+// case of an error or timeout. Is used to stop and reload client with TLS
+func (proxy *PgProxy) stopClientThread(logger *log.Entry) error {
```
Lagovas, line 413: maybe lets name `stopClientGoroutine` or `stopProxyClientConnection` as opposite to `ProxyClientConnection`?

decryptor/postgresql/pg_decryptor.go
G1gg1L3s: Ohh, I guess this resets the timeout, because it uses `time.Time{}` instead of `time.Now()`

Lagovas (approved):

PR #527, opened 2022-04-14
Deprecate `tokenize` option

G1gg1L3s:
This PR deprecates `tokenize` from encryptor config. It is still present, but it's role is taken by `token_type` - if it's not empty, the tokenization is enabled.

Also, forbids the `tokenize: false and token_type: <something>` pair.

## Checklist

  with new changes

### DIFF
--- a/encryptor/config/encryptionSettings.go
+++ b/encryptor/config/encryptionSettings.go
@@ -125,15 +125,22 @@ var validSettings = map[SettingMask]struct{}{
 	// TOKENIZATION
 	// default clientID
 	SettingTokenizationFlag | SettingTokenTypeFlag | SettingReEncryptionFlag | SettingAcraBlockEncryptionFlag: {},
+	SettingTokenTypeFlag | SettingReEncryptionFlag | SettingAcraBlockEncryptionFlag:                           {},
 
 	SettingTokenizationFlag | SettingTokenTypeFlag | SettingConsistentTokenizationFlag | SettingReEncryptionFlag:                                  {},
 	SettingTokenizationFlag | SettingTokenTypeFlag | SettingConsistentTokenizationFlag | SettingReEncryptionFlag | SettingAcraBlockEncryptionFlag: {},
+	SettingTokenTypeFlag | SettingConsistentTokenizationFlag | SettingReEncryptionFlag:                                                            {},
+	SettingTokenTypeFlag | SettingConsistentTokenizationFlag | SettingReEncryptionFlag | SettingAcraBlockEncryptionFlag:                           {},
 	// specified clientID
 	SettingTokenizationFlag | SettingTokenTypeFlag | SettingClientIDFlag | SettingReEncryptionFlag | SettingAcraBlockEncryptionFlag:                                     {},
 	SettingTokenizationFlag | SettingTokenTypeFlag | SettingConsistentTokenizationFlag | SettingClientIDFlag | SettingReEncryptionFlag | SettingAcraBlockEncryptionFlag: {},
+	SettingTokenTypeFlag | SettingClientIDFlag | SettingReEncryptionFlag | SettingAcraBlockEncryptionFlag:                                                               {},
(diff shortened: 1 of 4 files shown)


encryptor/config/encryptionSettings.go
```diff
@@ -243,9 +254,15 @@ func (s *BasicColumnEncryptionSetting) Init() (err error) {
...
+
+	tokenizeEnabled := (s.Tokenized != nil && *s.Tokenized)
+	if s.Tokenized != nil && !*s.Tokenized && s.TokenType != "" {
+		return errors.New("`tokenized` is disabled, but `token_type` is provided")
+	}
+
 	var tokenType common.TokenType
 	var ok bool
-	if s.TokenType != "" || s.Tokenized {
+	if s.TokenType != "" || tokenizeEnabled {
```
Lagovas, line 265: why not just `if s.TokenType != ""`? Becase now only TokenType is required.

encryptor/config/encryptionSettings.go
```diff
@@ -243,9 +254,15 @@ func (s *BasicColumnEncryptionSetting) Init() (err error) {
 	if s.ReEncryptToAcraBlock != nil && *s.ReEncryptToAcraBlock {
 		s.settingMask |= SettingReEncryptionFlag
 	}
+
+	tokenizeEnabled := (s.Tokenized != nil && *s.Tokenized)
+	if s.Tokenized != nil && !*s.Tokenized && s.TokenType != "" {
```
Lagovas, line 259: let's log warning that `Tokenized` option is deprecated if it is `!= nil`

encryptor/config/encryptionSettings.go
```diff
@@ -379,7 +396,7 @@ func (s *BasicColumnEncryptionSetting) ZoneID() []byte {
 
 // IsTokenized returns true if the column should be tokenized.
 func (s *BasicColumnEncryptionSetting) IsTokenized() bool {
-	return s.Tokenized
+	return s.TokenType != "" || (s.Tokenized != nil && *s.Tokenized)
```
Lagovas, line 399: I think we can check only `TokenType != ""` with new logic

encryptor/config/schemaStore_test.go
```diff
@@ -1061,3 +1062,83 @@ schemas:
...
+		{"Tokenized: false and token_type non empty",
+			`
+schemas:
+  - table: test_table
+    columns:
+      - data
+    encrypted:
+      - column: data
+        tokenized: false
+        token_type: str
```
Lagovas, line 1133: lets check one more case with empty `token_type`

encryptor/config/encryptionSettings.go
vixentael: I suggest smth like "Setting `tokenized` flag is not necessary anymore and will be ignored".

encryptor/config/encryptionSettings.go
G1gg1L3s: Yeah, good point :thinking:

encryptor/config/encryptionSettings.go
G1gg1L3s: Thought about this, but it would require passing `logger`, which either require changing function interface to pass it, or use `log` lib directly. Which is better?

Lagovas: at this case we can use `log` lib directly because it called at the start of program in global space, not in the specific user's session or something like that

encryptor/config/schemaStore_test.go
```diff
@@ -1130,6 +1130,16 @@ schemas:
         tokenized: false
         token_type: str
     `},
+		{"tokenized: true without token_type",
```
Lagovas, line 1135: lets check with empty value too?)
```
tokenized: true
token_type: ""
```
To cover case with only `!= nil` checks (that we validate not only with such conditions)

Lagovas (approved):

PR #529, opened 2022-04-18
Update python examples

ZhmakaAS:
Actualizing python examples. 

- Updated `encryptor_config_*` with explicit table declaration to run with transparent encryption and be able to show type awareness functionality
- Deleted `zone_id` field to be able to match zone from the query but not from the table.
- Updated `write_data` function with explicit zone_id literal - deleted `cast` operator as it is not supported by Acra sql parser now.  

## Checklist

  with new changes

### DIFF
--- a/examples/python/encryptor_config_with_zone.yaml
+++ b/examples/python/encryptor_config_with_zone.yaml
@@ -56,5 +56,5 @@ schemas:
     encrypted:
       - column: data
         data_type: bytes
-        # base64 bytes value
+        # base64 bytes value `test-data`
         default_data_value: dGVzdC1kYXRhCg==
--- a/examples/python/encryptor_config_with_zone.yaml
+++ b/examples/python/encryptor_config_with_zone.yaml
@@ -47,3 +47,14 @@ schemas:
       - email
     encrypted:
       - column: email
+
+  - table: test_example_with_zone
+    columns:
+      - id
+      - data
+      - raw_data
+    encrypted:
+      - column: data
+        data_type: bytes
+        # base64 bytes value
+        default_data_value: dGVzdC1kYXRhCg==
--- a/examples/python/example_with_zone.py
+++ b/examples/python/example_with_zone.py
@@ -14,7 +14,7 @@
 # coding: utf-8
 import argparse
 import ssl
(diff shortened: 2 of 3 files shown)


examples/python/encryptor_config_with_zone.yaml
```diff
@@ -46,4 +46,15 @@ schemas:
...
+
+  - table: test_example_with_zone
+    columns:
+      - id
+      - data
+      - raw_data
+    encrypted:
+      - column: data
+        data_type: bytes
+        # base64 bytes value
```
Lagovas, line 59: mention decoded value of base64 string to understand what we should expect: "test-data"

examples/python/example_with_zone.py
```diff
@@ -25,16 +25,16 @@ def print_data(zone_id, connection):
     console"""
     result = connection.execute(
         # explicitly pass zone id before related data
-        select([cast(zone_id.encode('utf-8'), BYTEA), test_table]))
+        select([literal_column("'{}'".format(zone_id)), test_table]))
```
Lagovas, line 28: what about to avoid manual escaping values, use sqlalchemy's engine and specify `type_` value for [literal_column](https://docs.sqlalchemy.org/en/14/core/sqlelement.html#sqlalchemy.sql.expression.literal_column)? I think that we can specify [String](https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.String) type to apply escape rules for string values (according to [TypeEngine](https://docs.sqlalchemy.org/en/14/core/type_api.html#sqlalchemy.types.TypeEngine) documentation)?
if it will works, I prefer to use such approach. Because it will automatically work for mysql and all other DBs if we will add them.

ZhmakaAS: Sounds good, let me check it.

ZhmakaAS: I used sqlalchemy's `literal`, it performs escaping values inside. What do you think @Lagovas ?

Lagovas (approved):
lgtm if it works with `literal`. I remember that had some problems with that. But probably it was related with compiling queries with bound values.

Lagovas (approved):
lgtm

PR #530, opened 2022-04-19
Fix `acra-keys list` with keystore v1

G1gg1L3s:
This PR fixes issues with `acra-keys` that were discovered at https://github.com/cossacklabs/product-docs/pull/245#discussion_r852193264.

The changes:
- `keystore.ListKeys` now lists files only once, if public key path is the same as private key one.
- in log key names `secure` is no longer interpreted as client id

## Checklist 

  with new changes

### DIFF
--- a/keystore/filesystem/server_keystore_test.go
+++ b/keystore/filesystem/server_keystore_test.go
@@ -19,7 +19,6 @@ package filesystem
 import (
 	"bytes"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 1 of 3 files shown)


keystore/filesystem/server_keystore.go
```diff
@@ -766,19 +766,19 @@ func (store *KeyStore) GenerateDataEncryptionKeys(id []byte) error {
 
 // ListKeys enumerates keys present in the keystore.
 func (store *KeyStore) ListKeys() ([]keystore.KeyDescription, error) {
-	privateKeys, err := store.describeDir(store.privateKeyDirectory)
+	keys, err := store.describeDir(store.privateKeyDirectory)
```
Lagovas, line 769: so we had printed the same entry twice due to omitted check for equal folders for public/private keys?

keystore/filesystem/server_keystore_test.go
```diff
@@ -1503,3 +1509,194 @@ func TestPoisonKeyGeneration(t *testing.T) {
...
+	fmt.Fprintln(os.Stderr, expected)
+	fmt.Fprintln(os.Stderr, "=> Actual Keys")
+	fmt.Fprintln(os.Stderr, found)
+
+	if expected != found {
+		t.Fatal("lists are different")
+	}
+}
+
+// copied from `acra-keys list` code
```
Lagovas, line 1655: it's the good reason to move this function to `keystore/utils.go` (probably) and use it in tests and acra-tools too, if we already need it in several places) Maybe in the future we will need it in one more place

keystore/filesystem/server_keystore.go
G1gg1L3s: Yes, we listed all keys in the same directory twice

Lagovas (approved):
lgtm

PR #532, opened 2022-04-29
Support for case (in)sensitive identifiers

iamnotacake:
### PostgreSQL

* Non-`"`-wrapped identifiers are converted to lowercase, just like how `postgresql` treats them.
  Encryptor config should contain lowercased identifiers for such table/column names.

* `"`-wrapped identifiers are processed "as is".
  Encryptor config should contain exactly the same identifiers to match them.

### MySQL

* Table identifiers are treated as case-insensitive by default.
  Encryptor config should contain lowercased table names in such case.

  There is a configuration option to change this:
  ```yaml
  database_config:
    mysql_case_sensitive_table_identifiers: true
  ```
  when enabled, encryptor config should contain exactly the same table name for matching to work. 
* Column names are always case-insensitive, so encryptor config should contain lowercased names.
* `` ` ``-wrapped identifiers do not affect case sensitivity.
  They are used in cases where, for example, you want to use a reserved word to name table or column, but not like in PostgreSQL.

---

This PR includes:

* Fix for PostgreSQL tokenizer

  When reading identifier (table or column name), remember if it was wrapped with quotes, and if so, return DOUBLE_QUOTE_STRING token type. Then, the AST will contain quote used to wrap identifier and converting
AST back to text representation with .Format() will include those quotes. Only do this for PostgreSQL, since MySQL behaves differently.

  Found this issue while executing `insert into "Test" ("A", "BcD") values ('a'::bytea, 'b'::bytea);`. Without the fix, Postgres complained about non-existing table name `"test"`.
  In postgres `Test` and `test` are the same, but `"Test"` and `Test` are different.

## Checklist

  with new changes~


sqlparser/token.go
```diff
@@ -777,6 +777,13 @@ func (tkn *Tokenizer) scanLiteralIdentifier() (int, []byte) {
 	if buffer.Len() == 0 {
 		return LEX_ERROR, buffer.Bytes()
 	}
+
+	// Double-quoted identifiers in PostgreSQL are case-sensitive.
+	// We need to remember which identifiers were wrapped with quotes so the query
+	// can be properly recreated later from the AST using .Format() method.
+	if quoteSeen && tkn.IsPostgreSQL() {
+		return DOUBLE_QUOTE_STRING, buffer.Bytes()
```
Lagovas, line 785: maybe will be better to return `stringTokenType[ch]`? instead of hardcoding type of token without relation to value that came

iamnotacake: Well, yes, but for PostgreSQL `tkn.dialect.QuoteHandler().IsIdentifierQuote(...)` only compares with `"`, so other type of quote can't appear here. To be able to use `stringTokenType[...]` I also need to track the quote somewhere, introduce another var and set it next to `quoteSeen = true`. You meant to make like this, right?

Also, it's kinda confusing for me that we enter this function (`scanLiteralIdentifier`) only if `tkn.dialect.QuoteHandler().IsIdentifierQuote(byte(ch))`, and the next chunk
```
if tkn.dialect.QuoteHandler().IsStringLiteralQuote(byte(ch)) {
	return tkn.scanString(ch, stringTokenType[ch])
}
```
is also about quotes. Then when the usual non-quote-wrapped identifier is processed?

Lagovas: > so other type of quote can't appear here

so you can return using this map to support mysql too and avoid hardcoding

>  I also need to track the quote somewhere, introduce another var and set it next to

hm, actually we can aggregate it. store `var quote *byte` and instead of `false` set nil. And at the end if it's not nil, return `stringTokenType[*quote]`

> Also, it's kinda confusing for me that we enter this function (scanLiteralIdentifier) only if tkn.dialect.QuoteHandler().IsIdentifierQuote(byte(ch)), and the next chunk
> 
> > if tkn.dialect.QuoteHandler().IsStringLiteralQuote(byte(ch)) {
> > 	return tkn.scanString(ch, stringTokenType[ch])
> > }
> 
> is also about quotes. Then when the usual non-quote-wrapped identifier is processed?

Quoted values can be identifiers or string literals. It depends on RDBMS (pgsql or mysql) and settings (in case of mysql). So at first we check is it identifier. After that it actually can be only string literal.

Identifiers can be quote wrapped and not quote wrapped. Not wrapped returned as `ID` from our tokenizer and it depends on grammar context to decide is it table's or column's or else's identifier.

iamnotacake:
The problem with the failing test `TestGeneralQueryParser_Parse` was in [encryptInsertQuery()](https://github.com/cossacklabs/acra/blob/39c6fb4bd77c8cb53716d27fd532a72a34de1a85/encryptor/queryDataEncryptor.go#L89):
```golang
func (encryptor *QueryDataEncryptor) encryptInsertQuery(...) (bool, error) {
	tableName := insert.Table.Name
	schema := encryptor.schemaStore.GetTableSchema(tableName.String()) // schema was nil here
```
the [`.String()` method of `TableIdent`](https://github.com/cossacklabs/acra/blob/39c6fb4bd77c8cb53716d27fd532a72a34de1a85/sqlparser/ast.go#L1831-L1836) included quotes and then, `store.schemas[tableName]` failed to find `"TableWithoutColumnSchema"` since it contained `TableWithoutColumnSchema`.

One option is to use [`.RawValue()`](https://github.com/cossacklabs/acra/blob/39c6fb4bd77c8cb53716d27fd532a72a34de1a85/sqlparser/ast.go#L1822-L1825) instead, since it ignores quotes. This fixes test 27, but then 28 is failing, seems like similar replacement is needed somewhere else as well.

Lagovas:
> One option is to use [.RawValue()](https://github.com/cossacklabs/acra/blob/39c6fb4bd77c8cb53716d27fd532a72a34de1a85/sqlparser/ast.go#L1822-L1825) instead, since it ignores quotes. This fixes test 27, but then 28 is failing, seems like similar replacement is needed somewhere else as well.

can you mention actual test that fails? because it is hard to help without logs or links to exact tests

> The problem with the failing test `TestGeneralQueryParser_Parse` was in [encryptInsertQuery()](https://github.com/cossacklabs/acra/blob/39c6fb4bd77c8cb53716d27fd532a72a34de1a85/encryptor/queryDataEncryptor.go#L89):
> 
> ```go
> func (encryptor *QueryDataEncryptor) encryptInsertQuery(...) (bool, error) {
> 	tableName := insert.Table.Name
> 	schema := encryptor.schemaStore.GetTableSchema(tableName.String()) // schema was nil here
> ```
> 
> the [`.String()` method of `TableIdent`](https://github.com/cossacklabs/acra/blob/39c6fb4bd77c8cb53716d27fd532a72a34de1a85/sqlparser/ast.go#L1831-L1836) included quotes and then, `store.schemas[tableName]` failed to find `"TableWithoutColumnSchema"` since it contained `TableWithoutColumnSchema`.
> 
> One option is to use [`.RawValue()`](https://github.com/cossacklabs/acra/blob/39c6fb4bd77c8cb53716d27fd532a72a34de1a85/sqlparser/ast.go#L1822-L1825) instead, since it ignores quotes. This fixes test 27, but then 28 is failing, seems like similar replacement is needed somewhere else as well.

as I see, we have incorrect rule for `column_id` with `SINGLE_QUOTE_STRING` token. Because nor mysql, nor postgresql don't accept single-quoted identifiers. So, unit-tests incorrect too.

But if fix it, I don't understand why `TableIdent` has `node.val` with double-quotes and return it with `RawValue()` too... can you debug it by yourself?

encryptor/queryDataEncryptor.go
```diff
@@ -86,7 +86,7 @@ func (encryptor *QueryDataEncryptor) ID() string {
 // encryptInsertQuery encrypt data in insert query in VALUES and ON DUPLICATE KEY UPDATE statements
 func (encryptor *QueryDataEncryptor) encryptInsertQuery(ctx context.Context, insert *sqlparser.Insert, bindPlaceholders map[int]config.ColumnEncryptionSetting) (bool, error) {
 	tableName := insert.Table.Name
-	schema := encryptor.schemaStore.GetTableSchema(tableName.String())
+	schema := encryptor.schemaStore.GetTableSchema(tableName.RawValue())
```
Lagovas, line 89: should we use ValueForConfig here instead of RawValue?

sqlparser/ast.go
```diff
@@ -1857,6 +1889,28 @@ func (node *TableIdent) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// ValueForConfig returns lowercased or non-lowercased value depending on
+// configuration (MySQL only) and which quotes were used to wrap the name in SQL,
+// suitable for usage when checking whether the column should be encrypted.
+func (node *TableIdent) ValueForConfig() string {
+	switch t := defaultDialect.(type) {
+	case *mysql.MySQLDialect:
+		if t.IsCaseSensitiveTableName() {
```
Lagovas, line 1898: why same approach wasn't applied for ColIdent?

sqlparser/dialect/mysql/dialect.go
```diff
@@ -21,8 +26,8 @@ func (dialect *MySQLDialect) QuoteHandler() dialect.QuoteHandler {
 }
 
 // NewMySQLDialect return new MySQLDialect
-func NewMySQLDialect() *MySQLDialect {
-	return &MySQLDialect{}
+func NewMySQLDialect(caseSensitiveTableName bool) *MySQLDialect {
```
Lagovas, line 29: lets extend dialect configuration with options? case sensitivity is first option but probably not last. Lets simplify  future extending with dynamic amount of parameters? as we did it here - https://github.com/cossacklabs/acra/blob/master/decryptor/base/accessContext.go#L49

sqlparser/token.go
```diff
@@ -777,6 +778,13 @@ func (tkn *Tokenizer) scanLiteralIdentifier() (int, []byte) {
 	if buffer.Len() == 0 {
 		return LEX_ERROR, buffer.Bytes()
 	}
+
+	// Double-quoted identifiers in PostgreSQL are case-sensitive.
+	// We need to remember which identifiers were wrapped with quotes so the query
+	// can be properly recreated later from the AST using .Format() method.
+	if quoteSeen != nil && tkn.IsPostgreSQL() {
```
Lagovas, line 785: mysql supports quoted identifiers too - https://dev.mysql.com/doc/refman/8.0/en/identifiers.html#:~:text=Certain%20objects%20within%20MySQL%2C%20including,names%20are%20known%20as%20identifiers. With ANSI_QUOTES parameter. And we keep in mind that parameter in several places too.

encryptor/config/schemaStore.go
```diff
@@ -22,11 +22,19 @@ import (
...
 type TableSchemaStore interface {
+	GetDatabaseConfig() *DatabaseConfig
 	// GetTableSchema returns schema for given table if configured, or nil otherwise.
 	GetTableSchema(tableName string) TableSchema
 	GetGlobalSettingsMask() SettingMask
 }
 
+// databaseConfig stores database-specific configuration that can affect connection
+// to the database, how SQL queries are processed and so on
+type DatabaseConfig struct {
```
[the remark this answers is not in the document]
iamnotacake: With getters, this thing does not need to be public. But there are also some tests in `decryptor/postgresql/proxy_test.go`, since there is a mock implementation of `TableSchemaStore`, and since we need to implement `GetDatabaseConfig() *DatabaseConfig` as part of that interface, we need `DatabaseConfig` to be available there, outside of `config` package.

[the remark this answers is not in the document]
Lagovas: to implement `GetDatabaseConfig() DatabaseConfig`. Where `DatabaseConfig` is interface, not pointer to struct. And yes, in tests we should use some public implementations or mocks if we test other things

encryptor/config/schemaStore.go
```diff
@@ -50,14 +58,16 @@ func (d defaultValues) ShouldReEncryptAcraStructToAcraBlock() bool {
 }
 
 type storeConfig struct {
-	Defaults *defaultValues
-	Schemas  []*tableSchema
+	DatabaseConfig *DatabaseConfig
```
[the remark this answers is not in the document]
Lagovas: it wasn't public. we use private implementations of public interfaces to decouple parts that uses config from parts that implements. And can change implementation internally without touching parts that uses them

encryptor/queryDataEncryptor_test.go
```diff
@@ -393,7 +411,104 @@ schemas:
...
+			Changed:           true,
+			ExpectedIDS:       [][]byte{specifiedClientID, zoneID, defaultClientID},
+			DataCoder:         &PostgresqlDBDataCoder{},
+			dialect:           postgresql.NewPostgreSQLDialect(),
+		},
+
+		// 29. different case table identifiers, postgresql
+		// should match, lowercase config identifier == lowercase SQL identifier
+		{
+			Query:             `UPDATE lowercasetable set "other_column"='%s', "specified_client_id"='%s', "zone_id"='%s', "default_client_id"='%s'`,
```
Lagovas, line 427: can you add test cases where we test same rules for column names?

tests/test.py
```diff
@@ -10138,6 +10138,211 @@ class TestPostgresqlTypeAwareDecryptionWithDefaultsPsycopg3(Psycopg3ExecutorMixi
...
+    def runTestCase(self,
+                    table_name: str,
+                    quoted_table_name: bool,
+                    column_name: str,
+                    quoted_column_name: bool,
+                    should_match: bool):
+        test_string = "test"
+
+        if quoted_table_name:
+            table_name = '"' + table_name + '"'
```
Lagovas, line 10172: why we don't lower table name too as we do it for column names below?

tests/test.py
```diff
@@ -10138,6 +10138,211 @@ class TestPostgresqlTypeAwareDecryptionWithDefaultsPsycopg3(Psycopg3ExecutorMixi
...
+        def check():
+            # ensure it is encrypted (if should_match) or ensure it's not encrypted (if not should_match)
+            if should_match:
+                # ensure decrypted data matches what was inserted
+
+                result = self.engine1.execute(f"SELECT {column_name} FROM {table_name} WHERE id={id};")
+                row = result.fetchone()
+
+                if bytes(row[row_name]) != bytes(test_string, "UTF-8"):
+                    self.fail(f"Table identifier {table_name}, column identifier {column_name}, did not match (decryption failed?), {bytes(row['data'])} != {test_string}")
```
Lagovas, line 10192: why you don't use proper method for that:
```
self.assertEqual(
    bytes(row[row_name]), 
    bytes(test_string, "UTF-8"),
    f"Table identifier {table_name}, column identifier {column_name}, did not match (decryption failed?), {bytes(row['data'])} != {test_string}"
```
? 
same for checks below

tests/test.py
```diff
@@ -10138,6 +10138,211 @@ class TestPostgresqlTypeAwareDecryptionWithDefaultsPsycopg3(Psycopg3ExecutorMixi
...
+        check()
+
+        # update a record
+        self.engine1.execute(f"UPDATE {table_name} SET {column_name}='{test_string}' WHERE id={id};")
+        check()
+
+    def testLowerConfigLowerQuery(self):
+        # table should match, lowercase config identifier == lowercase SQL identifier
+        # column should only match in quoted "DATA" case
+        self.runTestCase("lowercase_table", False, "data", False, False)
```
Lagovas, line 10233: when read such functions it is hard to understand what means these booleans and to not forget them. Can you use named arguments or named values:
```
QUOTED, NOT_QUOTED = (True, False)
SHOULD_MATCH, SHOULD_NOT_MATCH = (True, False)
self.runTestCase("lowercase_table", NOT_QUOTED, "data", QUOTED, SHOULD_NOT_MATCH)

```

tests/test.py
```diff
@@ -10138,6 +10138,211 @@ class TestPostgresqlTypeAwareDecryptionWithDefaultsPsycopg3(Psycopg3ExecutorMixi
...
+        self.engine1.execute(f"CREATE TABLE IF NOT EXISTS UPPERCASE_TABLE (id INT PRIMARY KEY AUTO_INCREMENT, data BLOB);")
+
+    def tearDown(self):
+        self.engine1.execute(f"DROP TABLE lowercase_table;")
+        self.engine1.execute(f"DROP TABLE LOWERCASE_TABLE;")
+        self.engine1.execute(f"DROP TABLE uppercase_table;")
+        self.engine1.execute(f"DROP TABLE UPPERCASE_TABLE;")
+        super().tearDown()
+
+    def runTestCase(self, table_name: str, quoted: bool, should_match: bool):
```
Lagovas, line 10292: can you add comment somewhere why we don't test here different cases for columns? will be much better to leave reference to doc with explanation that mysql doesn't have case-sensitivity for columns

encryptor/queryDataEncryptor_test.go
Lagovas: can we test in unit-tests different cases for columns too? we test it integration tests but will be great to test it in unit-tests too. because they are faster and as it looks for me, not so difficult to add it

tests/test.py
iamnotacake: Sounds good, but... should I add these constants in all test methods? Or add to test class and access with `self.`? Both methods don't sound like a best solution. Making global constants would be even worse.

Lagovas: it is okay to copy-paste it to each method that use it

encryptor/config/schemaStore.go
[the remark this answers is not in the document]
Lagovas: em... defaultValues and tableSchema are private and deserializes correctly - https://github.com/cossacklabs/acra/blob/master/encryptor/config/schemaStore.go#L71
there is only one requirement - fields should be public (DatabaseConfig) and fields of private structs should be public. But not struct itself

encryptor/queryDataEncryptor_test.go
```diff
@@ -524,10 +523,20 @@ schemas:
 			DataCoder:         &PostgresqlDBDataCoder{},
 			dialect:           postgresql.NewPostgreSQLDialect(),
 		},
-		// 1. different case table identifiers, postgresql
-		// should match, lowercase config identifier == lowercase SQL identifier
+		// 1. should partially match, like #0 but DEFAULT_client_id is not quoted and is processed as lowercase
 		{
-			Query:             `UPDATE "lowercasetable" set "other_column"='%s', "specified_client_id"='%s', "zone_id"='%s', "default_client_id"='%s'`,
+			Query:             `UPDATE lowercasetable set other_column='%s', specified_client_id='%s', zone_id='%s', DEFAULT_client_id='%s'`,
```
Lagovas, line 528: hm... it doesn't match but processed without errors... can you create task to warn users in cases when we matched table but didn't match column if we have table schema? we should warn or even prevent processing cases with misbehavior from the user side. Leave now as is, but we should know about it and at least warning. It is worth a separate task.

encryptor/queryDataEncryptor_test.go
```diff
@@ -72,6 +72,76 @@ func normalizeQueryWithDialect(dialect dialect.Dialect, query string) (string, e
...
+	Changed           bool
+	ExpectedIDS       [][]byte
+	DataCoder         DBDataCoder
+	dialect           dialect.Dialect
+}
+
+func testParsing(t *testing.T, testData []parserTestData, encryptedValue, defaultClientID []byte, schemaStore *config.MapTableSchemaStore) {
+	encryptor := &testEncryptor{value: encryptedValue}
+	parser := sqlparser.New(sqlparser.ModeStrict)
+	mysqlParser, err := NewMysqlQueryEncryptor(schemaStore, parser, encryptor)
```
Lagovas, line 89: maybe lets name it just parser? because it is not `mysql` specific. and it is not a parser, it is encryptor.

encryptor/queryDataEncryptor_test.go
iamnotacake: Maybe it is also possible to configure AcraCensor to warn/error on non-existing columns to catch this case? Oh, AcraCensor may even need changes like in this PR as well

encryptor/queryDataEncryptor_test.go
iamnotacake: Will name it `queryEncryptor`.

Seems like there is no difference whether `NewMysqlQueryEncryptor` or `NewPostgresqlQueryEncryptor` is used in these tests, so probably no need in one more arg to `testParser()` to choose which QueryEncryptor to use.

encryptor/queryDataEncryptor_test.go
Lagovas: no, it is not responsibility of AcraCensor. and it has rules case sensitive for now. maybe if we will meet some issues with that, we will change it in the future

Lagovas (approved):

PR #533, opened 2022-05-04
Add response on fail for mysql

G1gg1L3s:
Mysql doesn't respond to the different options of the `response_on_fail` (`default_value`, `error` and `ciphertext`) field. This PR changes that and bring support as in https://github.com/cossacklabs/acra/pull/521.

## Checklist

  with new changes

### DIFF
--- a/decryptor/base/encoder.go
+++ b/decryptor/base/encoder.go
@@ -141,37 +141,37 @@ func (v *IntValue) AsMysqlText() []byte {
 	return PutLengthEncodedString([]byte(v.strValue))
 }
 
-// IdentityValue is an encodingValue that just returns data as is
-type IdentityValue struct {
+// StringValue is an EncodingValue that encodes data into string format
+type StringValue struct {
 	data []byte
 }
 
-// NewIdentityValue returns EncodingValue as identity value
-func NewIdentityValue(data []byte) EncodingValue {
-	return &IdentityValue{data}
+// NewStringValue returns string EncodingValue
+func NewStringValue(data []byte) EncodingValue {
+	return &StringValue{data}
 }
 
 // AsPostgresBinary returns value encoded in postgres binary format
-// For identity value this means returning value as it is
-func (v *IdentityValue) AsPostgresBinary() []byte {
+// In other words, it returns data as it is
+func (v *StringValue) AsPostgresBinary() []byte {
 	return v.data
 }
 
 // AsPostgresText returns value encoded in postgres text format
-// For identity value this means returning value as it is
-func (v *IdentityValue) AsPostgresText() []byte {
+// In other words, it returns data as it is
+func (v *StringValue) AsPostgresText() []byte {
 	return v.data
 }
 
 // AsMysqlBinary returns value encoded in mysql binary format
-// For identity value this means returning value as it is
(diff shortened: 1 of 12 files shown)


decryptor/base/encoder.go
```diff
@@ -28,3 +39,193 @@ func (e *EncodingError) Is(err error) bool {
 func NewEncodingError(column string) error {
 	return &EncodingError{column}
 }
+
+// EncodingValue represents a (possibly parsed and prepared) value that is
+// ready to be encoded
+type EncodingValue interface {
+	// AsPostgresBinary returns value encoded in postgres binary format
+	AsPostgresBinary() []byte
```
Lagovas, line 47: we use `base` package to store common things for all databases. what we will do after adding one more protocol? for example mysql X or mariaDB or mssql?)
okay, we want return type aware values line Int/String/Bytes which knows how to encode value according to DBs protocol. okay, there is great place for design patterns)

we can introduce fabric that can create new Int/String/Bytes values that will implement EncodingValue with methods like `EncodeToBinary`/`EncodeToInt`:
```
type TypeFactory interface {
  NewBytesValue(...) EncodingValue
  NewIntValue(...) EncodingValue
  NewStringValue(...) EncodingValue
}
```

Then update common function `EncodeDefault` and accept +1 parameter `func EncodeDefault(..., typeFactory TypeFactory)` and use it to create new Int/String values instead of calling functions `NewBytesValue` that should implement methods for mysql/postgresql/etc

After that target db package should implement own fabric that will return own implementations of `EncodedValue` and return them.

decryptor/base/utils.go
```diff
@@ -14,13 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package mysql
+package base
```
Lagovas, line 17: it's mysql related functions. they should be stored in `mysql` package, not in `base`)

decryptor/base/encoder.go
G1gg1L3s: Yeah, you are right. It was one of the possible implementations, but I thought that it has much complexity. But current implementation is hard to extend. So, good idea!

ZhmakaAS (approved):
Lgtm

decryptor/mysql/data_encoder_test.go
```diff
@@ -0,0 +1,280 @@
...
+
+func TestSuccessfulTextEncoding(t *testing.T) {
+	type testcase struct {
+		input    []byte
+		dataType common.EncryptedType
+		expected []byte
+	}
+
+	testcases := []testcase{
+		{[]byte("string"), common.EncryptedType_String, []byte("\x06string")},
```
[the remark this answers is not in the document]
G1gg1L3s: Don't forget that this is a go string :)
The first byte(s) of encoded string is a length of that string. It out case, length is 6. So, to insert a `06` byte I used escaping.
Another way of writing that is:
```go
[]byte{06, 's', 't', 'r', 'i', 'n', 'g'}
```

Which, imho, is less readable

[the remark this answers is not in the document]
Lagovas: aaaaaaaa, understood. thanks for explanation. you r right

PR #534, opened 2022-05-05
Extend sqlparser 

ZhmakaAS:
Simple extending sql parser grammar to support statements in the format of:

`SET <id> TO SINGLE_QUOTE_STRING|DEFAULT|<value>`

## Checklist

  with new changes


sqlparser/analyzer_test.go
```diff
@@ -477,6 +477,33 @@ func TestExtractSetValues(t *testing.T) {
 		sql:   "set session sql_safe_updates = 0",
 		out:   map[SetKey]interface{}{{Key: "sql_safe_updates", Scope: "session"}: int64(0)},
 		scope: "session",
+	}, {
+		sql: "set search_path to 'value'",
+		out: map[SetKey]interface{}{{Key: "search_path", Scope: "session"}: "value"},
```
Lagovas, line 482: I saw a comment for `ExtractSetValues` function that `SET` supports strings integers and booleans. can we check them too?

sqlparser/sql.y
```diff
@@ -550,6 +551,14 @@ isolation_level:
     $$ = &SetExpr{Name: NewColIdent("tx_isolation"), Expr: NewStrVal([]byte("serializable"))}
   }
 
+set_operation_scope:
+  set_session_or_global
+  | LOCAL
```
Lagovas, line 556: can we mention that LOCAL is postgresql specific? 
https://dev.mysql.com/doc/refman/5.7/en/set-variable.html - in mysql's docs there is nothing about local
https://www.postgresql.org/docs/current/sql-set.html - but supported by postgres

sqlparser/sql.y
```diff
@@ -3010,6 +3038,10 @@ set_list:
   {
     $$ = append($1, $3)
   }
+| sql_id TO value_expression ',' value_expression
+  {
+    $$ = SetExprs{&SetExpr{Name: $1, Expr: $3}, &SetExpr{Name: $1, Expr: $5}}
```
Lagovas, line 3043: hm... how it works with `SET name TO 1,2,3,4`? or `SET name to 'str1', 'str2', 'str3'`? for amount of values > 2
I saw that such cases processed with `append` items to the list, like `$$ = append($$, ...)`

ZhmakaAS: Yes, you are right, in this case we need probably introduce a new type like `set_to_list` or something similar to handle such thing.

Lagovas (approved):
lgtm

PR #537, opened 2022-05-13
Do not abort connection after encoding error in postgresql

G1gg1L3s:
This PR implement skipping postgres packets from database after encoding error, almost in the same way as the previous PR does for mysql. It also adds a bunch of tests for different cases:
- Simple queries
- Extended queries (and prepared statements, that are basically the same under the hood).
- Batched extended queries (when database sends `n` packets and waits for the next request)

The only issue with postgres is that a state of transaction is stored at database side, and changes in case of some errors (even parsing ones). When the state is changed, `commit` is basically `rollback`:
```
test=# begin;
BEGIN
test=*# select 1/0;
ERROR:  division by zero
test=!# commit;
ROLLBACK
```

But our error will not trigger that:
```
test=# begin;
BEGIN
test=*# select value_bytes from testtype;
ERROR:  encoding error in column "value_bytes"
test=# commit;
COMMIT
```
So, for now I will mention it in the docs. Currently there is no straight and simple solution.

P.S. `test.py` crossed the milestone of 10 thousands lines, so my congratulations! It takes 5 second for my formatter to proceed the file. That's what I call success. 

## Checklist

  with new changes
- [x] ~CHANGELOG_DEV.md is updated~

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -9859,6 +9859,8 @@ class TestDbFlushingOnError(BaseTransparentEncryption):
 
                 conn.execute(query).fetchall()
 
+        # Most db-drivers do a rollback after an exception, so check
+        # that our data is not saved due to the rollback.
         row = self.engine1.execute(select_data).fetchone()
         self.assertEqual(row, None)
 
@@ -10004,6 +10006,8 @@ class TestPostgresqlDbFlushingOnError(BaseTransparentEncryption):
 
             self.assertEqual(ex.exception.message,
                              'encoding error in column "value_bytes"')
+            # Most db-drivers do a rollback after an exception, so check
+            # that our data is not saved due to the rollback.
             row = await conn.fetchrow(select_query, data['id'])
             self.assertEqual(row, None)
 
--- a/tests/test.py
+++ b/tests/test.py
@@ -9859,6 +9859,110 @@ class TestMySQLDbFlushingOnError(BaseBinaryMySQLTestCase, BaseTransparentEncrypt
         row = self.engine1.execute(select_data).fetchone()
         self.assertEqual(row, None)
 
+class TestPostgresqlDbFlushingOnError(BaseTransparentEncryption):
+    encryptor_table = sa.Table(
+        'test_proper_db_flushing_on_error', sa.MetaData(),
+        sa.Column('id', sa.Integer, primary_key=True),
+        sa.Column('value_bytes', sa.LargeBinary),
+    )
+    ENCRYPTOR_CONFIG = get_encryptor_config('tests/encryptor_configs/transparent_type_aware_decryption.yaml')
+
+    def checkSkip(self):
+        if not (TEST_POSTGRESQL and TEST_WITH_TLS):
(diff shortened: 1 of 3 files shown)


tests/test.py
```diff
@@ -9859,6 +9861,311 @@ def testTransactionRollback(self):
...
+                    # Expect encoding error
+                    await stmt.fetch(corrupted_data['id'])
+
+            self.assertEqual(ex.exception.message,
+                             'encoding error in column "value_bytes"')
+            row = await conn.fetchrow(select_query, data['id'])
+            self.assertEqual(row, None)
+
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(test())
```
Lagovas, line 10100: I fiound that python has high-level helper [asyncio.run](https://docs.python.org/3/library/asyncio-task.html#asyncio.run) for that. Lets start to use it?) Works since python3.7.

tests/test.py
```diff
@@ -9859,6 +9861,311 @@ def testTransactionRollback(self):
...
+                    self.assertEqual(data['value_bytes'], row['value_bytes'])
+
+                    stmt = await conn.prepare(select_query)
+                    # Expect encoding error
+                    await stmt.fetch(corrupted_data['id'])
+
+            self.assertEqual(ex.exception.message,
+                             'encoding error in column "value_bytes"')
+            row = await conn.fetchrow(select_query, data['id'])
+            self.assertEqual(row, None)
```
Lagovas, line 10097: hm, can you explain why we expect `None` here? As I understand:
1. we inserted data and it is already in the database
2. We tried to get this data in the transaction, took error from acra side and handled exception
3. We try to get same data but get None as result, not exception due to same error and not error that `select` returns no data?

tests/test.py
G1gg1L3s: I wanted to use that, because I use it in small scripts myself. But I've read from the docs:
> This function cannot be called when another asyncio event loop is running in the same thread.
I don't know whether our test run in parallel, but if they do, this can cause a problem (or I just don't understand this statement correctly)

tests/test.py
Lagovas: oh, thanks. can you add some comment above this check with explanation why we expect here `None`?
Because I'm sure that someone will forget that most of ORM use `ROLLBACK` on any exception. Or mention that after that we expect that ORM will do ROLLBACK. Just for case if event this ORM will change his logic and stop do ROLLBACK on errors.

tests/test.py
Lagovas: then lets create new event loop for that?) use 'new_event_loop` instead of `get_event_loop`

G1gg1L3s: Okay, but I don't quite understand the reasons. Could you explain why?

Lagovas: it will help run code in parallel if need. every thread or test case will have own loop and it will safe to stop loop on failure and don't touch other tests that can rely on common loop in current thread

Lagovas (approved):

PR #540, opened 2022-05-18
Remove placeholders from connection when done

G1gg1L3s:
This PR solves interference problem of placeholders. They are saved in the connection, but were never removed, only replaced. That means, they could affect other queries on the same connection.

Solve this, by removing the placeholders at the end of the query: when `ReadyForQuery` arrives. 

I didn't write a new test for this, because 1) it already existed as a part of another test and 2) it's really hard to reproduce.
 
## Checklist

  with new changes~

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -9927,11 +9927,10 @@ class TestPostgresqlDbFlushingOnError(BaseTransparentEncryption):
                 WHERE id = $1
             """
 
-            # TODO(G1gg1L3s): uncomment when T2572 is fixed
-            # await conn.execute(insert_query, data['id'], data['value_bytes'])
-            #
-            # row = await conn.fetchrow(select_query, data['id'])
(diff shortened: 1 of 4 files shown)


tests/test.py
```diff
@@ -9927,11 +9927,10 @@ async def test():
                 WHERE id = $1
             """
 
-            # TODO(G1gg1L3s): uncomment when T2572 is fixed
-            # await conn.execute(insert_query, data['id'], data['value_bytes'])
-            #
-            # row = await conn.fetchrow(select_query, data['id'])
-            # self.assertEqual(data['value_bytes'], row['value_bytes'])
+            await conn.execute(insert_query, data['id'], data['value_bytes'])
```
Lagovas, line 9930: cool)

Lagovas (approved):

PR #541, opened 2022-05-20
Substitute oid in parse

G1gg1L3s:
This PR fixes substitution of OID in packets. Before, it substituted them in the data messages, which are returned by select. But there is another place where they can arrive - in parse packets. We missed it, because they are optional and most frontends do not specify them.

To test it, I've added a new dependency to our python tests - psycopg3. The first stable version was released on [2021-10-13](https://www.psycopg.org/articles/2021/10/13/psycopg-30-released/), but I think we can start to use it because of its support of various features. For example in this case, this frontend sets the type of parameters in the parse packets. Also it helped to discover bug with encoding of i32, so in some cases tests could fall. I will fix it as soon as possible.

## Checklist

  with new changes~

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -44,6 +44,7 @@ import mysql.connector
 import psycopg2
 import psycopg2.errors
 import psycopg2.extras
+import psycopg as psycopg3
 import pymysql
 import redis
 import requests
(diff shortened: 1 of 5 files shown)


tests/test.py
```diff
@@ -1045,6 +1046,44 @@ def execute_prepared_statement(self, query, args=None):
                 return data
 
 
+class Psycopg3Executor(QueryExecutor):
+    def _execute(self, query, args=None, prepare=None):
+        # We expect postgres placeholders, like $1, $2, $3...
+        # But psycopg expects only %s
+        # TODO: check that numbers are in the correct order
```
Lagovas, line 1053: they can be used not in ordered way. for example `insert into table1 values($1, $3, $1, $2, $3)`. It will depends on Bind packet and order of passed parameters

tests/test.py
```diff
@@ -10077,6 +10116,34 @@ async def test():
         loop.run_until_complete(test())
 
 
+class Psycopg3ExecutorMixin:
+    def setUp(self):
```
Lagovas, line 10120: hm, I see such pattern not first time. So maybe we should introduce some Base class with such approach of initializing executors?) Like `BaseExecutorInitialization` and require to override `get_executor` method with proper class or correct configuration?) We can require to set class via class variable and introduce one more method `get_executor_connection_args` if some child class needs not common args.

tests/test.py
```diff
@@ -10077,6 +10116,34 @@ async def test():
...
+                ssl_cert=ssl_cert,
+                raw=True,
+            )
+            return Psycopg3Executor(args)
+
+        self.executor1 = executor(TEST_TLS_CLIENT_KEY, TEST_TLS_CLIENT_CERT)
+        self.executor2 = executor(
+            TEST_TLS_CLIENT_2_KEY, TEST_TLS_CLIENT_2_CERT)
+        self.raw_executor = executor(
+            TEST_TLS_CLIENT_KEY, TEST_TLS_CLIENT_CERT, 5432)
```
Lagovas, line 10138:
```suggestion
            TEST_TLS_CLIENT_KEY, TEST_TLS_CLIENT_CERT, TEST_DB_PORT)
```

tests/test.py
G1gg1L3s: Yeah, but psycopg don't have positional arguments. So, I think I will just create new array of arguments with duplications and the correct order

tests/test.py
G1gg1L3s: I don't quite understand how you see it. Can you provide an example of usage of such executor?

Beside that, the idea is cool!

G1gg1L3s:
The broken tests are the result of T2568. The fix is at https://github.com/cossacklabs/acra/pull/542

Lagovas (approved):

PR #543, opened 2022-05-23
[T2580] Fix handling null values in mysql

Lagovas:
On decryption our envelope_detector returned `[]byte{}` on `nil` values instead leaving `nil`. MySQL encoded differently empty strings and NULL values. [NULL values should](https://dev.mysql.com/doc/internals/en/com-query-response.html#packet-ProtocolText::ResultsetRow) have 0xfb value. So on Acra side when it receive null value as 0xfb then it changed to empty array and encode it as LengthEncoded [String](https://dev.mysql.com/doc/internals/en/string.html#packet-Protocol::LengthEncodedString) or Integer that are incorrect.

So in this PR removed processing null data in crypto envelope. If it will be empty array, it will be returned as empty array too.

## Checklist

  with new changes

### DIFF
--- a/crypto/envelope_detector.go
+++ b/crypto/envelope_detector.go
@@ -196,6 +196,9 @@ func (wrapper *OldContainerDetectorWrapper) OnCryptoEnvelope(ctx context.Context
 
 // OnColumn callback which finds serializedContainer or AcraStruct/AcraBlock for backward compatibility
 func (wrapper *OldContainerDetectorWrapper) OnColumn(ctx context.Context, inBuffer []byte) (context.Context, []byte, error) {
+	if inBuffer == nil {
+		return ctx, inBuffer, nil
+	}
 	// we should track that if incoming data contains any signs of new container and if it is we return data as is
 	// otherwise try to search for AcraBlock or AcraStruct to save backward compatibility
 	// so before any OnColumn we should reset hasMatchedEnvelope flag to track if its changed during EnvelopeDetector OnColumn via BackWrapper.OnCryptoEnvelope callback
--- a/tests/test.py
+++ b/tests/test.py
@@ -5654,13 +5654,8 @@ class TestEmptyValues(BaseTestCase):
         # check null values
         result = self.engine1.execute(sa.select([self.temp_table]).where(self.temp_table.c.id == null_value_id))
         row = result.fetchone()
-        if TEST_MYSQL:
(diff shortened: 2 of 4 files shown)


tests/test.py
```diff
@@ -8667,15 +8662,15 @@ def testClientIDRead(self):
 
         # mysql.connector represent null value as empty string
```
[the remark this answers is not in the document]
G1gg1L3s: Also, one small suggestion. We have `columns` and `null_columns` to check null case separately. But now when `null_columns` are `None`, we can merge two loops of checking into one, as it is done in postgres:
```python
columns = ('value_bytes', 'value_int32', 'value_int64', 'value_empty_str', 'value_str', 'value_null_str', 'value_null_int32')
for column in columns:
     self.assertEqual(data[column], row[column])
     self.assertIsInstance(row[column], type(data[column]))
```

You can do it only if you want. Current variant is also ok.

crypto/envelope_detector.go
```diff
@@ -196,6 +196,9 @@ func (wrapper *OldContainerDetectorWrapper) OnCryptoEnvelope(ctx context.Context
 
 // OnColumn callback which finds serializedContainer or AcraStruct/AcraBlock for backward compatibility
 func (wrapper *OldContainerDetectorWrapper) OnColumn(ctx context.Context, inBuffer []byte) (context.Context, []byte, error) {
+	if inBuffer == nil {
```
ZhmakaAS, line 199: @Lagovas Can we place this check on the level of `EnvelopeDetector`? We are already doing some `inBuffer` checking [here](https://github.com/cossacklabs/acra/blob/master/crypto/envelope_detector.go#L60)

Lagovas: but we return there `inBuffer` as is. Without allocating new buffer. If `inBuffer` in OnColumn will be `nil`, it will returned as is due to `if len(inBuffer) < SerializedContainerMinSize`

tests/test.py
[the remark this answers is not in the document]
Lagovas: you are right, it is much better. updated

G1gg1L3s (approved):
Cool!

crypto/envelope_detector.go
ZhmakaAS: Got it;

ZhmakaAS (approved):
lgtm

PR #544, opened 2022-05-23
Handle binary char set introducer

G1gg1L3s:
This small PR adds support of mysql _binary charset introducer, which is used at least by java frontend. I would like to add other charsets, but they are not supported by the parser yet. So, if there is no need, I wouldn't do that.

## Checklist

  with new changes

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -59,6 +59,7 @@ from sqlalchemy.dialects import mysql as mysql_dialect
 from sqlalchemy.dialects import postgresql as postgresql_dialect
 from sqlalchemy.dialects.postgresql import BYTEA
 from sqlalchemy.exc import DatabaseError, OperationalError, ProgrammingError
+from sqlalchemy.sql import expression, operators
 
 import api_pb2
 import api_pb2_grpc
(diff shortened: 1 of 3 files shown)


tests/test.py
```diff
@@ -9714,6 +9714,42 @@ def testClientIDRead(self):
...
+        sure it can handle _binary.
+
+        [1]: https://dev.mysql.com/doc/refman/8.0/en/charset-introducer.html
+        """
+
+        self.schema_table.create(bind=self.engine_raw, checkfirst=True)
+        id = get_random_id()
+        # DON'T EVER, EVER DO THIS
+        # Do not use direct string interpolation for sql values
+        # This is just a test example
```
Lagovas, line 9730: we can avoid using raw sql and build query with sqlalchemy:
```
>>> from sqlalchemy.sql import expression
>>> from sqlalchemy.sql import operators
>>> unary = expression.UnaryExpression(
      sa.Column("column"), 
      operator=operators.custom_op("_binary"), 
      type_=sa.Numeric)
>>> str(unary)
'_binary "column"'
```
Use `unary` instead of `self.test_table.c.value_bytes`

tests/test.py
```diff
@@ -9714,6 +9715,43 @@ def testClientIDRead(self):
...
+        """
+        We don't directly support charset introducers[1], but at least make
+        sure it can handle _binary.
+
+        [1]: https://dev.mysql.com/doc/refman/8.0/en/charset-introducer.html
+        """
+
+        self.schema_table.create(bind=self.engine_raw, checkfirst=True)
+        id = get_random_id()
+        insert = self.test_table.insert().values(
```
Lagovas, line 9729: hm, if sqlalchemy compile query with parentheses, then lets test them both? and use raw sql query to use without parentheses if sqlalchemy doesn't allow compile without them

Lagovas (approved):

PR #552, opened 2022-07-05
Implement KMS AWS master key loading

ZhmakaAS:
Implementation of KMS AWS MasterKey Loading.

## Checklist

  with new changes


keystore/kms/aws/encryptor.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Encrypt implementation of kms.Encryptor method
+func (e *Encryptor) Encrypt(keyID string, data []byte) ([]byte, error) {
+	input := &kms.EncryptInput{
+		KeyId:     aws.String(keyID),
+		Plaintext: data,
+	}
+
+	result, err := e.client.Encrypt(context.Background(), input)
```
Lagovas, line 65: lets use context with at least deadline equal to [default network timeout](https://github.com/cossacklabs/acra/blob/master/network/deadline_listener.go#L27)

keystore/keyloader/kms/kms_loader.go
```diff
@@ -0,0 +1,106 @@
...
+
+// NewLoader create new kms MasterKeyLoader
+func NewLoader(credentialPath, keyIdentifierURI string) (*Loader, error) {
+	keyID, err := kms.NewKeyIdentifierFromURI(keyIdentifierURI)
+	if err != nil {
+		return nil, err
+	}
+
+	var encryptor kms.Encryptor
+	switch keyID.Prefix() {
```
Lagovas, line 28: hm, it push me to think about registries. what about to use here registries approach as we did it for CryptoHandlers? Turn implementation on with build tag and register it at start? for example:
```
type EncryptorFactoryMethod interface {
  New(credentialPath string) (kms.Encryptor, error)
}
// or type EncryptorFactoryMethod func(credentialPath string) (kms.Encryptor, error)
var lock = sync.Lock{}
var encryptors map[string]EncryptorFactoryMethod
RegisterEncryptor(encryptor EncryptorFactoryMethod){
  lock.Lock()
  encryptors[encryptor.ID()] = encryptor
  lock.Unlock()
}
```

Then we can use :
```
func NewLoader(credentialPath, keyIdentifierURI string) (*Loader, error) {
	keyID, err := kms.NewKeyIdentifierFromURI(keyIdentifierURI)
	if err != nil {
		return nil, err
	}
        encryptorFactory, ok := encryptors[keyID]
        if !ok {
            log.Errorln("Unknown key ID")
        }
        encryptor, err := encryptorFactory.New(credentialPath) // or encryptorFactory(credentialPath)
}
```

Then our AWS implementation should have:
```
func init(){
  kms.RegisterEncryptor(someFactoryFunction)
}
```
With such approach we can extend with enterprise KMS support.


Additionally, we can make default loader that do nothing with master key (current approach where env variable store result master key. this implementation will return master key as is in Encrypt/Decrypt methods) and registered by default and +1 with AWS. Then current loader will be consistent with new approach

keystore/kms/aws/encryptor.go
```diff
@@ -0,0 +1,119 @@
...
+
+	cfg, err := config.LoadDefaultConfig(context.Background(), loadOptions...)
+	if err != nil {
+		return nil, err
+	}
+	return kms.NewFromConfig(cfg), nil
+}
+
+func init() {
+	baseKMS.RegisterEncryptorCreator(KeyIdentifierPrefix, NewEncryptor)
```
Lagovas, line 118: can we move this function to separate file and wrap it with build flag? for example `// go:build !kmw_aws_off`. As I understand, it will include this file by default and will exclude if we will compile binary with `go build -tags=kms_aws_off`. So after that this implementation will not be registered.

keystore/kms/base.go
```diff
@@ -0,0 +1,77 @@
...
+var lock = sync.Mutex{}
+var encryptorCreators = map[string]EncryptorCreateFunc{}
+
+// EncryptorCreateFunc generic function for creating Encryptor
+type EncryptorCreateFunc func(credentialPath string) (Encryptor, error)
+
+// RegisterEncryptorCreator add new EncryptorCreator to registry
+func RegisterEncryptorCreator(encryptorID string, encryptorCreateFunc EncryptorCreateFunc) {
+	lock.Lock()
+	encryptorCreators[encryptorID] = encryptorCreateFunc
```
Lagovas, line 19: hm. maybe lets log debug info about the registered creator? probably it will help debug issues with unsupported modules in the future

tests/test.py
```diff
@@ -1317,6 +1319,31 @@ def get_vault_cli_args(self, mount_path=None, secret_path=None):
...
+
+    def create_key(self):
+        response = self.kms_client.create_key(Description='AcraMasterKey KEK')
+        return response['KeyMetadata']['Arn']
+
+    def disable_key(self, keyId):
+        self.kms_client.disable_key(KeyId=keyId)
+
+    def generate_master_key(self, keyId):
+        response = self.kms_client.generate_data_key(
```
Lagovas, line 1340: hm, can we generate random master key and encrypt it via KMS as it actually will be used in the acra? I understand, that the logically generation data key from KMS side should be the same. But we met situations when 3d party products broke their API and much better if we can detect changes of API from which we depend. Plus, we may use real AWS API for tests in the future if meet situations when the emulator is not enough and it may charge extra resources for +1 kind of key and different api call

keystore/kms/base.go
ZhmakaAS: Agree!

keystore/kms/aws/encryptor.go
ZhmakaAS: In such case we need to build a binary in our tests with this flag, right? @Lagovas 

Something like that?
```
If TEST_WITH_AWS {
 go build acra-server -tags=kms_aws_off
}
```

keystore/keyloader/kms/kms_loader.go
```diff
@@ -93,14 +87,14 @@ func (loader *Loader) LoadMasterKeys() (encryption []byte, signature []byte, err
 	return keys.Encryption, keys.Signature, nil
 }
 
-func (loader *Loader) decryptWithKMSKey(keyID kms.KeyIdentifier) ([]byte, error) {
+func (loader *Loader) decryptWithKMSKey(keyID string) ([]byte, error) {
```
Lagovas, line 90: lets use []byte type for identifier? most likely that we will use clientID and related things as keyIDs and much often it all will have []byte type.

keystore/kms/aws/encryptor.go
```diff
@@ -67,7 +65,7 @@ func (e *Encryptor) Encrypt(ctx context.Context, keyID string, data []byte) ([]b
 func (e *Encryptor) Decrypt(ctx context.Context, keyID string, blob []byte) ([]byte, error) {
 	input := &kms.DecryptInput{
 		CiphertextBlob: blob,
-		KeyId:          aws.String(keyID),
+		KeyId:          aws.String(fmt.Sprintf("alias/%s", keyID)),
```
Lagovas, line 68: does AWS have doc page about the requirement to have `alias/` prefix for key ids? if yes, can we put reference here?

ZhmakaAS: Done

Lagovas (approved):
lgtm

PR #553, opened 2022-07-15
Keymaker master key with KMS

ZhmakaAS:
Implemented basic support of KMS in the `keymaker` for AMK generation. Introduce the new flag `kms_key_policy` which describes possible policies for KMS keys. For now only `create` is supported. 

Made a small refactoring of KMS package by creating a new extended common interface for all KMS operations (`Keystore` - not an ideal name, can think about a better name). Also, create a separate abstraction called `KmsClient` that should care about all internal stuff for corresponded KMS.

## Checklist

  with new changes


keystore/kms/aws/client.go
```diff
@@ -21,72 +22,97 @@ type Configuration struct {
...
+// KMSClient represent general AWS KMS client
+type KMSClient struct {
 	client *kms.Client
 }
 
-// NewEncryptor create new AWS KMS encryptor which implement Encryptor interface
-func NewEncryptor(credentialPath string) (baseKMS.Encryptor, error) {
-	configuration, err := readConfigByPath(credentialPath)
+// NewKmsClient create new KMS AWS client
+func NewKmsClient(cfg *Configuration) (*KMSClient, error) {
```
Lagovas, line 31:
```suggestion
func NewKMSClient(cfg *Configuration) (*KMSClient, error) {
```

It returns KMSClient, so lets name similar

keystore/kms/aws/keystore.go
```diff
@@ -0,0 +1,96 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	baseKMS "github.com/cossacklabs/acra/keystore/kms"
+)
+
+// Keystore implementation of AWS Keystore
+type Keystore struct {
```
Lagovas, line 13: hm, maybe lets name it as Encryptor? Because it conflicts with our component KeyStore that has much more responsibilities and used for different things. This struct will be used to encrypt/decrypt keys... Or any other name except KeyStore. KMSStore, or KMSEncryptor, or else. Or main idea to introduce kms.Keystore, another component different from keystore.Keystore?

keystore/kms/aws/keystore.go
```diff
@@ -0,0 +1,96 @@
...
+
+	for _, alias := range aliases {
+		// format of ARN is arn:aws:kms:eu-west-1:account-id:alias/acra-master-key-test
+		arnParts := strings.Split(*alias.AliasArn, ":")
+		if len(arnParts) < 6 {
+			return false, errors.New("invalid arn format found for keyId")
+		}
+		region := arnParts[3]
+
+		if *alias.AliasName == fmt.Sprintf("alias/%s", keyID) && region == k.cfg.Region {
```
Lagovas, line 68: much better to create new function like `getKeyName(keyID string) string {return "alias/"+keyID}`. Because it will be much faster than `fmt.Sprintf` which need to parse string at first and after that place value for placeholder.

tests/test.py
```diff
@@ -1213,6 +1213,56 @@ def test_gen_keys_with_empty_client_id(self):
...
+            '--generate_master_key={}'.format(master_key_file.name),
+            '--kms_type=aws',
+            '--kms_credentials_path={}'.format(self.config_file)])
+
+        resp = self.kms_client.list_aliases()
+        created_arn = resp['Aliases'][0]['AliasArn']
+
+        ciphertext = open(master_key_file.name, "rb").read()
+        decrypt_resp = self.kms_client.decrypt(keyId=created_arn, ciphertextBlob=ciphertext)
+        self.assertEqual(len(decrypt_resp['Plaintext']), 32)
```
Lagovas, line 1251: lets check that plaintext not equal to ciphertext too?

tests/test.py
```diff
@@ -1213,6 +1213,56 @@ def test_gen_keys_with_empty_client_id(self):
...
+            'secret_access_key': 'secret_key_id',
+            'region': 'eu-west-1',
+            'endpoint':  os.environ.get('AWS_KMS_ADDRESS', 'http://localhost:8080')
+        }
+        cfg_json_string = json.dumps(configuration)
+        self.config_file = tempfile.NamedTemporaryFile('w+', encoding='utf-8').name
+
+        # Open the file for writing.
+        with open(self.config_file, 'w') as f:
+            f.write(cfg_json_string)
```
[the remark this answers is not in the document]
ZhmakaAS: I think we can.

tests/test.py
ZhmakaAS: Sure

tests/test.py
```diff
@@ -1213,6 +1213,57 @@ def test_gen_keys_with_empty_client_id(self):
...
+            '--kms_type=aws',
+            '--kms_credentials_path={}'.format(self.config_file.name)])
+
+        resp = self.kms_client.list_aliases()
+        created_arn = resp['Aliases'][0]['AliasArn']
+
+        ciphertext = open(master_key_file.name, "rb").read()
+        decrypt_resp = self.kms_client.decrypt(keyId=created_arn, ciphertextBlob=ciphertext)
+        self.assertEqual(len(decrypt_resp['Plaintext']), 32)
+        self.assertNotEqual(len(decrypt_resp['Plaintext']), ciphertext)
```
Lagovas, line 1249: hm, we compare here length of plaintext with ciphertext's content? we should check contents, not length

Lagovas (changes requested):

Lagovas (approved):
lgtm

PR #555, opened 2022-07-26
Add cache encryptor in keystore

ZhmakaAS:
This PR includes creating a new cache encryptor (SCellKeyEncryptor) for improving Keystore operation. This can help to read keys from KMS only once and store them in an encrypted way in the cache.

Distribute changed all over acra tools that use Keystore. 

## Checklist

  with new changes

### DIFF
--- a/keystore/filesystem/server_keystore.go
+++ b/keystore/filesystem/server_keystore.go
@@ -80,23 +80,22 @@ func NewFileSystemKeyStoreWithCacheSize(directory string, encryptor keystore.Key
 }
 
 // NewFilesystemKeyStore represents keystore that reads keys from key folders, and stores them in memory.
-func NewFilesystemKeyStore(directory string, encryptor, cacheEncryptor keystore.KeyEncryptor) (*KeyStore, error) {
+func NewFilesystemKeyStore(directory string, encryptor keystore.KeyEncryptor) (*KeyStore, error) {
 	return NewCustomFilesystemKeyStore().KeyDirectory(directory).Encryptor(encryptor).Build()
 }
 
 // NewFilesystemKeyStoreTwoPath creates new KeyStore using separate folders for private and public keys.
-func NewFilesystemKeyStoreTwoPath(privateKeyFolder, publicKeyFolder string, encryptor, cacheEncryptor keystore.KeyEncryptor) (*KeyStore, error) {
+func NewFilesystemKeyStoreTwoPath(privateKeyFolder, publicKeyFolder string, encryptor keystore.KeyEncryptor) (*KeyStore, error) {
 	return NewCustomFilesystemKeyStore().KeyDirectories(privateKeyFolder, publicKeyFolder).Encryptor(encryptor).Build()
 }
 
 // KeyStoreBuilder allows to build a custom keystore.
 type KeyStoreBuilder struct {
-	privateKeyDir  string
-	publicKeyDir   string
-	encryptor      keystore.KeyEncryptor
-	cacheEncryptor keystore.KeyEncryptor
-	storage        Storage
-	cacheSize      int
+	privateKeyDir string
+	publicKeyDir  string
+	encryptor     keystore.KeyEncryptor
+	storage       Storage
+	cacheSize     int
 }
 
 // NewCustomFilesystemKeyStore allows a custom-made KeyStore to be built.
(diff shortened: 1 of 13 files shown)


keystore/filesystem/server_keystore.go
```diff
@@ -1216,14 +1238,32 @@ func (store *KeyStore) generateAndSaveSymmetricKey(id []byte, filename string) e
...
-		store.Add(filename, encryptedSymKey)
+		return store.loadKeyAndCache(id, filename, func() ([]byte, error) {
+			return store.ReadKeyFile(store.GetPrivateKeyFilePath(filename))
+		})
+	}
+	return store.cacheEncryptor.Decrypt(encryptedSymKey, id)
+}
+
+func (store *KeyStore) loadKeyAndCache(id []byte, filename string, loadKeyCallback func() ([]byte, error)) ([]byte, error) {
+	encryptedKey, err := loadKeyCallback()
```
G1gg1L3s, line 1249: The idea with callback is cool, but I don't understand why not just pass the key directly? The callback is used immediately and only once, so I don't see any reasons to store it. Therefore, instead of this:
```go
return store.loadKeyAndCache(id, filename, func() ([]byte, error) {
    return store.ReadKeyFile(store.GetPrivateKeyFilePath(filename))
})
```

We can write something like:
```go
key, err := store.ReadKeyFile(store.GetPrivateKeyFilePath(filename))
if err != nil { ... }
return store.cacheKey(id, filename, key)
```

Though, at least callback requires less boilerplate code for error handling :)

ZhmakaAS: @G1gg1L3s, Yes, the idea was to create some compact function that does all the necessary logic under the hood.

Just thought about some sort of improvement. We can do the following:

```
loadKeyAndCache(id []byte, filename string, loadKeyCallback func(filePath string) ([]byte, error)) ([]byte, error) {
	encryptedKey, err := loadKeyCallback(store.GetPrivateKeyFilePath(filename))
	if err != nil {
		return nil, err
	}
	.....
	.....
```

And on the caller side just:

```
store.loadKeyAndCache([]byte(SecureLogKeyFilename), filename, store.ReadKeyFile)
```

But with such approach, we can lose some flexibility in future variants of getting the path to keys. What do you think guys? cc @Lagovas

Lagovas: as I understand, we cannot use @G1gg1L3s's approach:
```
key, err := store.ReadKeyFile(store.GetPrivateKeyFilePath(filename))
if err != nil { ... }
return store.cacheKey(id, filename, key)
```
because this snippet returns symmetric key. But we also have case with asymmetric keys that uses `loadPrivateKey` function - https://github.com/cossacklabs/acra/blob/03f7f071f33ef344d920c08387c8cffd9d160b97/keystore/filesystem/server_keystore.go#L580

So we cannot use only one snippet, we should use several. And everywhere repeat flow:
* read encrypted key
* decrypt
* encrypt for cache
* store in cache
* return decrypted

And this callback approach helps encapsulate key decryption, encryption and saving in cache steps, and avoid extra row repeating. Or did I miss something?

ZhmakaAS: Yes, you are right, it was exactly the reason for using it.

keystore/filesystem/server_keystore.go
```diff
@@ -555,19 +576,26 @@ func (store *KeyStore) getPrivateKeyByFilename(id []byte, filename string) (*key
 	defer store.lock.Unlock()
 	encryptedKey, ok := store.cache.Get(filename)
 	if !ok {
-		encryptedPrivateKey, err := store.loadPrivateKey(store.GetPrivateKeyFilePath(filename))
+		loadKeyCallback := func() ([]byte, error) {
+			encryptedPrivateKey, err := store.loadPrivateKey(store.GetPrivateKeyFilePath(filename))
+			if err != nil {
+				return nil, err
+			}
+			return encryptedPrivateKey.Value, nil
```
G1gg1L3s, line 584: Why not to simplify? :)
```suggestion
			return store.loadPrivateKey(store.GetPrivateKeyFilePath(filename))
```

G1gg1L3s (approved):
Cool!

keystore/filesystem/server_keystore.go
G1gg1L3s: Oops, it's impossible due to `.Value`, sorry :sweat_smile:

Lagovas (approved):

PR #556, opened 2022-08-01
Use kms per client

Zhaars:
Initial Implementation of KMS per client. 

This PR contains several significant changes:
- change `keystore.Encryptor` interface - now `Encrypt/Decrypt` method expect to pass  **KeyContext**. 
- SCellEncryptor implementation was changed based on `KeyContext.Context`
- Implemented `kms.Encryptor` and `kms.KeyMakingWrapper` which is simple Decorator with kms key creation 
```
type KeyContext struct {
  ClientID []byte
  ZoneID []byte
  Context []byte
  Purpose []byte/string/etc
}
```

Left some TODOs in the code with some open questions to discuss.

## Checklist

  with new changes


keystore/kms/key_encryptor.go
```diff
@@ -0,0 +1,70 @@
...
+		kmsEncryptor,
+	}
+}
+
+// Encrypt return encrypted key using KMS encryptor and context.
+func (encryptor *KeyEncryptor) Encrypt(ctx context.Context, key []byte, keyContext keystore.KeyContext) ([]byte, error) {
+	keyID, err := getKeyIDFromContext(keyContext)
+	if err != nil {
+		// TODO: add logging
+		return nil, err
```
Lagovas, line 35: we can try to get logger from the context via `logging.GetLoggerFromContext` function

keystore/kms/key_encryptor.go
```diff
@@ -0,0 +1,70 @@
...
+	if ctx.Purpose == "" {
+		return nil, ErrMissingKeyPurpose
+	}
+
+	switch ctx.Purpose {
+	case keystore.PurposeStorageClientSymmetricKey, keystore.PurposeStorageClientPrivateKey, keystore.PurposeSearchHMAC:
+		if ctx.ClientID == nil {
+			return nil, ErrEmptyClientIDProvided
+		}
+		return []byte(string(ctx.ClientID) + "/" + ctx.Purpose.String()), nil
```
Lagovas, line 60: first our plan was to use 1 key for X purpose keys. So we should use `<clientID>` keyID for symmetric, asymmetric, hmac keys

keystore/kms/key_encryptor.go
```diff
@@ -0,0 +1,70 @@
...
+	case keystore.PurposeStorageZoneSymmetricKey, keystore.PurposeStorageZonePrivateKey:
+		if ctx.ZoneID == nil {
+			return nil, ErrEmptyZoneIDProvided
+		}
+		return []byte(string(ctx.ClientID) + "/" + ctx.Purpose.String()), nil
```
Lagovas, line 65:
```suggestion
		return ctx.ZoneID, nil
```

keystore/kms/key_making_wrapper.go
```diff
@@ -0,0 +1,115 @@
...
+
+	return k.KeyMaking.GenerateClientIDSymmetricKey(id)
+}
+
+func (k KeyMakingWrapper) createKMSKeyFromContext(keyContext keystore.KeyContext) error {
+	ctx, _ := context.WithTimeout(context.Background(), network.DefaultNetworkTimeout)
+
+	keyID, err := getKeyIDFromContext(keyContext)
+	if err != nil {
+		return err
```
Lagovas, line 105: Maybe let's log errors too? or will it be logged on the caller side?

keystore/filesystem/filesystem_backup.go
```diff
@@ -151,7 +153,10 @@ func readFilesAsKeys(files []string, basePath string, encryptor keystore2.KeyEnc
 		// remove absolute first part, leave only relative to path
 		relativeName := strings.Replace(f, basePath+"/", "", -1)
 		if isPrivate(relativeName) {
-			content, err = encryptor.Decrypt(content, getIDFromFilename(relativeName))
+			purpose, id := getIDFromFilename(relativeName)
+
+			keyContext := keystore.NewKeyContext(purpose).WithContext(id).WithZoneID(id).WithClientID(id)
```
Lagovas, line 158: hm, why we initialize here context with `id` as ZoneID/Context/ClientID?

keystore/keystore.go
```diff
@@ -202,14 +265,14 @@ func NewSCellKeyEncryptor(masterKey []byte) (*SCellKeyEncryptor, error) {
 }
 
 // Encrypt return encrypted key using masterKey and context.
-func (encryptor *SCellKeyEncryptor) Encrypt(key, context []byte) ([]byte, error) {
-	encrypted, _, err := encryptor.scell.Protect(key, context)
+func (encryptor *SCellKeyEncryptor) Encrypt(ctx context.Context, key []byte, keyContext KeyContext) ([]byte, error) {
+	encrypted, _, err := encryptor.scell.Protect(key, keyContext.Context)
```
Lagovas, line 269: I think that previous scell encryptor should have helper that decides what data from the KeyContext to use for operations. Our current approach is to use ZoneID if specified otherwise ClientID. Much better will be recognize zones by ZoneID prefix otherwise use clientID. Not sure what priority should have Context for now

keystore/kms/key_encryptor.go
ZhmakaAS: Got it, makes sense

keystore/kms/key_making_wrapper.go
ZhmakaAS: I think it's ok to have logging on the caller side. For example, in `keymaker` we have panicking in case of error.

keystore/filesystem/filesystem_backup.go
ZhmakaAS: The main reason is that we don't know about the key purpose here.

`getIDFromFilename` return purpose and ID and the purpose can also be as some zone specific or client one, so we need to pass id as Zone or ClientID and will understand what it's actually the key based on its purpose

keystore/filesystem/filesystem_backup.go
```diff
@@ -82,57 +83,58 @@ func isPrivate(fname string) bool {
...
-		return []byte(fname[:len(fname)-len("_zone")])
+		return keystore.PurposeStorageZonePrivateKey, []byte(fname[:len(fname)-len("_zone")])
 	}
 	if strings.HasSuffix(fname, "_zone_sym") {
-		return []byte(fname[:len(fname)-len("_zone_sym")])
+		return keystore.PurposeStorageZoneSymmetricKey, []byte(fname[:len(fname)-len("_zone_sym")])
 	}
 	if strings.HasSuffix(fname, "_sym") {
-		return []byte(fname[:len(fname)-len("_sym")])
+		//TODO: what it the sym key here?
```
G1gg1L3s, line 122: @Lagovas, I think this is question to you

keystore/filesystem/server_keystore.go
```diff
@@ -239,8 +242,10 @@ func newFilesystemKeyStore(privateKeyFolder, publicKeyFolder string, storage Sto
 			return nil, err
 		}
 	}
+
+	ctx, _ := context.WithTimeout(context.Background(), network.DefaultNetworkTimeout)
```
G1gg1L3s, line 246: Can you please tell, why there is a timeout in context?

keystore/filesystem/filesystem_backup.go
Lagovas: probably zone's sym key - https://github.com/cossacklabs/acra/blob/master/keystore/filesystem/server_keystore.go#L943

keystore/filesystem/server_keystore.go
ZhmakaAS: Sure. It is a good practice to have a timeout for any network operation. We can ignore it for our SCellEncrytor as we don't do any network calls, but for KMS-related stuff, we should have some timeout checking. In the case of KMS unavailability or some network issue, the request will exit (in 60 secs) with an error TimoutExceeded

ZhmakaAS: And context is a commonly used thing for this case in Go. A great practice to pass context as the first param to the functions

tests/test.py
```diff
@@ -2299,6 +2299,65 @@ def testReadAcrastructInAcrastruct(self):
...
+
+    def get_kms_type(self):
+        return 'aws'
+
+    def get_kms_configuration_path(self):
+        return self.config_file.name
+
+
+class KMSEncryptorMixing:
+    ZONES = []
```
Lagovas, line 2326: will be great to add comment why we introduce here new list with zones

tests/test.py
```diff
@@ -2299,6 +2299,65 @@ def testReadAcrastructInAcrastruct(self):
...
+    def setUp(self):
+        self.keystore = tempfile.TemporaryDirectory()
+        self.keys_dir = os.path.join(self.keystore.name, '.acrakeys')
+
+        extra_args = {
+            'kms_keystore_encryptor' : 'true',
+            'kms_type': self.get_kms_type(),
+            'kms_credentials_path': self.get_kms_configuration_path(),
+        }
+        assert create_client_keypair_from_certificate(TEST_TLS_CLIENT_CERT, keys_dir=self.keys_dir, extra_kwargs=extra_args) == 0
```
Lagovas, line 2337: hm, will it conflict with already existing keys in keystore with different arguments for keymaker (e.g. without any kms)?

tests/test.py
ZhmakaAS: Sure

cmd/acra-keys/keys/migrate-keys.go
```diff
@@ -226,7 +226,7 @@ func MigrateV1toV2(srcV1 filesystem.KeyExport, dstV2 keystoreV2.KeyFileImportV1)
 
 	log.Tracef("Importing %d keys from keystore v1", expected)
 	for _, key := range keys {
-		log := log.WithField("purpose", key.Purpose).WithField("id", key.ID)
+		log := log.WithField("purpose", key.KeyContext.Purpose).WithField("id", keystore.GetKeyContextFromContext(key.KeyContext))
```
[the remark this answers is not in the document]
ZhmakaAS: But we are logging `key.KeyContext.Purpose` here or what do you mean?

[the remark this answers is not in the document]
Lagovas: `.WithField("id", keystore.GetKeyContextFromContext(key.KeyContext))`
How it will look in log entry? Something like `&{ClientID: nil, ZoneID: nil, Context: []byte{0, 12, 42, 321}}` or something else?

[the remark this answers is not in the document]
ZhmakaAS: Not precisely, previously we had

```
log := log.WithField("purpose", key.Purpose).WithField("id", key.ID) 
```
where `key.ID` is simple bytes slice which is strange as it would print something like `[]byte{0, 12, 42, 321}`

So I just repeat the functionality. Here `key.KeyContext` is not bytes slice but `keystore.KeyContext` struct. So now it will print some bytes slice because of the function.
```
func GetKeyContextFromContext(keyContext KeyContext) []byte {
	if keyContext.ZoneID != nil {
		return keyContext.ZoneID
	}
	if keyContext.ClientID != nil {
		return keyContext.ClientID
	}
	if keyContext.Context != nil {
		return keyContext.Context
	}
	return nil
}
```
I think it makes sense to wrap with `string` to print string data instead of bytes @Lagovas

keystore/kms/aws/client.go
```diff
@@ -106,12 +106,21 @@ func (e *KMSClient) ListAliases(ctx context.Context, keyID *string) ([]types.Ali
 		input.KeyId = keyID
 	}
 
-	result, err := e.client.ListAliases(ctx, input)
-	if err != nil {
-		return nil, err
+	aliases := make([]types.AliasListEntry, 0)
```
Lagovas, line 109: as I understand, here we expect at least 1 entry, am i correct? if so, then lets initialize with size of 1? or if it list a lot of aliases, then lets initialize with meaningful amount of entries. for example if here is only client names then lets use number 8 (we use 2 for clientID in tests and 2 for zoneID).

tests/test.py
```diff
@@ -396,6 +396,22 @@ def get_poison_record():
     return poison_record
 
 
+def get_new_poison_record(extra_kwargs: dict=None, keys_dir=None):
+    """generate one new poison record for speed up tests and don't create subprocess
+    for new records"""
+    args = [os.path.join(BINARY_OUTPUT_FOLDER, 'acra-poisonrecordmaker')]
+    if keys_dir:
+        args.append('--keys_dir={}'.format(keys_dir))
```
Lagovas, line 404: should we pass here keystore version too?

Zhaars: No, its is required for `keymaker` but not for `poisonrecordmaker`.

keystore/kms/aws/client.go
Zhaars: Yes, it's a good question. The default AWS pagination size is 50 items and we can specify how many items we want to return per page. For example, during the key making, we will call this function on every key, even if no aliases have been specified yet, so we will do some redundant allocation.

cmd/acra-keys/keys/migrate-keys.go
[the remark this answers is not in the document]
Lagovas: +1

Lagovas (approved):
great job

PR #560, opened 2022-08-10
delete minor unreachable code

Abirdcfly:
Signed-off-by: Abirdcfly <fp544037857@gmail.com>

## Checklist

  with new changes

### DIFF
--- a/poison/callbacks.go
+++ b/poison/callbacks.go
@@ -18,11 +18,12 @@ package poison
 
 import (
 	"container/list"
+	"os"
+	"os/exec"
+
 	"github.com/cossacklabs/acra/decryptor/base"
 	"github.com/cossacklabs/acra/logging"
 	log "github.com/sirupsen/logrus"
-	"os"
-	"os/exec"
 )
 
 // EmptyCallback implements Callback and log message to show that RecordProcessor works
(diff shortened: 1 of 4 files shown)


poison/callbacks.go
```diff
@@ -41,8 +42,7 @@ type StopCallback struct{}
 func (*StopCallback) Call() error {
 	log.WithField(logging.FieldKeyEventCode, logging.EventCodePoisonRecordDetectionMessage).Warningln("Detected poison record, exit")
 	os.Exit(1)
-	log.WithField(logging.FieldKeyEventCode, logging.EventCodePoisonRecordDetectionMessage).Errorln("executed code after os.Exit")
-	return nil
+	panic("executed code after os.Exit")
```
G1gg1L3s, line 45: Wait, is that even possible?

Abirdcfly: This is a little trick I learned from the golang source code that if we want to specify that we don't need to `return` here, but the compiler will report an error if we don't write `return`, we can use `panic` instead.

https://github.com/golang/go/blob/43456202a1e55da55666fac9d56ace7654a65b64/src/crypto/tls/handshake_test.go#L305-L306

Lagovas: it is good approach, but we prefer to log it, skip bad case and continue handling further requests. It is a questionable case, do we need to log and rely on existing alerts that catch unexpected log messages and react on that or leave here panic and rely on tests (which will detect issues with `os` package or source code updates) and infrastructure that will restart service on failure and have enough extra instances to catch requests during restart...
After revising this case, I think will be better to leave here panic and catch it in our tests.
@Zhaars , @G1gg1L3s, what do you think?

Lagovas (approved):

PR #564, opened 2022-08-17
Refactor tls configs

Lagovas:
Here is added function for registering flags for tls configs and final construction tls.Config structs. First candidate for use is Redis. This update will help easily register similar flags to generate config with ocsp/crl usage, same wording and naming. Additionally, it allows generate flags for components that works as client or database which represented in naming (as we do now) and extra flags (like SNI for clients).

Why clients may need separate configuration for different components? For example, if they have several cloud solutions with TLS support, with own CA and server's certificates. For example, if service deployed on AWS, uses AWS's database, uses Elastic cloud for search features and cloud managed Hashicorp Vault. 

There are left base common flags (`tls_key/tls_cert/tls_auth/tls_ca`) that works as one-place configuration for all configs that allows to specify only them in the simplest cases.

Also, remove the previously deprecated `tls_db_sni` parameter that breaks general design of TLS configuration. And much easier to remove it instead of supporting edge-case.

Additionally such configuration extended current `client`/`database` parameter separation with separate ocsp/crl parameters.

P.S. it is re-created PR from local to repo branch.

## Checklist

  with new changes


network/ocsp.go
```diff
@@ -124,6 +125,47 @@ const (
...
+	if f := flags.Lookup(namerFunc(name, "check_only_leaf_certificate", "ocsp")); f != nil {
+		getter, ok := f.Value.(flag.Getter)
+		if !ok {
+			log.Fatal("Can't cast flag's Value to Getter")
+		}
+		val, ok := getter.Get().(*bool)
+		if val == nil {
+			checkOnlyLeafCert = false
+		} else {
+			if !ok {
```
Zhaars, line 159: Hmm, is it even possible when the getter returns `not-nil` **value**  but the `ok` will be false?

network/crl.go
```diff
@@ -127,6 +128,66 @@ func (e CRLError) Unwrap() error {
...
+		}
+		val, ok := getter.Get().(uint)
+		if !ok {
+			log.WithField("value", getter.Get()).Fatalf("Can't cast %s to integer value",
+				namerFunc(name, "cache_size", "crl"))
+		}
+		crlCacheSize = val
+	}
+	if f := flags.Lookup(namerFunc(name, "cache_time", "crl")); f != nil {
+		getter, ok := f.Value.(flag.Getter)
```
Zhaars, line 177: Don't you think to use just some `Parse` methods for flag parsing?

```
crlCacheSize, err = strconv.ParseUint(f.Value.String(), 10, 0)
if err != nil {
      log.WithField("value", getter.Get()).Fatalf("Can't cast %s to integer value", namerFunc(name, "cache_time", "crl"))
}
```

It will reduce the amount of code plus we can avoid these runtime interface castings.

network/tls_wrapper.go
```diff
@@ -62,6 +64,45 @@ var (
 	tlsServerName string
 )
 
+// correct -tls_ocsp_database_url:
+// incorrect -tls_database_ocsp_url:
+
+// NamerFunc func compile final parameter name for specified service name
+type NamerFunc func(serviceName, parameterName, groupName string) string
```
Zhaars, line 71: Looks like the too generic name. Maybe we can call some like `CLIParamNameConstructorFunc` or something that says that this function constructs the final name for CLI parameters for a specific service.

CHANGELOG_DEV.md
```diff
@@ -1,3 +1,9 @@
+# 0.94.0 - 2022-08-18
+- Removed deprecated `--tls_db_sni` flag. Now is available only `--tls_database_sni`.
```
iamnotacake, line 2:
```suggestion
- Removed deprecated `--tls_db_sni` flag. Now only `--tls_database_sni` is available.
```

network/ocsp.go
Lagovas: hm. you are right. actually is not. but if we will leave:
```
val, _ := getter.Get().(*bool)
		if val == nil {
			checkOnlyLeafCert = false
		} else {
			checkOnlyLeafCert = *val
		}
```
as for me, it will not be so obvious after several month why it did so. And I will need spend a time to understand that it is legal)

network/crl.go
Lagovas: you are right. will do

network/tls_wrapper.go
Lagovas: agree, will rename

CHANGELOG_DEV.md
Lagovas: agree, will do it in next batch of commits to run CI once

Lagovas (commented):
applied your suggestions, @Zhaars , @iamnotacake

Zhaars (approved):
lgtm

PR #566, opened 2022-08-19
Add TLS support for redis clients [3/3]

Lagovas:
* here is registered new parameters related to tls and redis
* updated existing integration tests to run with and without tls connections to redis
* duplicated existing golang's integration tests to run with tls only if build tag provided
* additionally I found that if any testcase log something directly to stdout, IDEA display that test cases were terminated and passed less then total amount of them. For example 386 of 400. So, also updated read-key command to be able override destination of writing just for tests purposes and simplifying development
* updated existing tests related to redis with removing hardcoded hosts and using values from env variable as it were designed at start. It was required to run same tests with redis through TLS in the same circleci job.

## Checklist

  with new changes


cmd/acra-keymaker/acra-keymaker.go
```diff
@@ -296,7 +297,17 @@ func openKeyStoreV1(output, outputPublic string) keystore.KeyMaking {
...
 	if redis := cmd.ParseRedisCLIParameters(); redis.KeysConfigured() {
-		keyStorage, err := filesystem.NewRedisStorage(redis.HostPort, redis.Password, redis.DBKeys, nil)
+		// if redisTLS = nil then will not be used TLS for Redis
+		var redisTLS *tls.Config
+		var err error
+		if redis.TLSEnable {
+			redisTLS, err = network.NewTLSConfigByName(flag.CommandLine, "redis", redis.HostPort, network.ClientNameConstructorFunc())
+			if err != nil {
+				log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorCantInitKeyStore).
+					Errorln("Can't initialize TLS config for Redis client")
```
Zhaars, line 307: Don't we want to exit on error here?

cmd/redis_testutils.go
```diff
@@ -0,0 +1,29 @@
...
+
+// GetTestRedisOptions returns options configured with test env variables
+// use this function for tests
+func GetTestRedisOptions(t *testing.T) RedisOptions {
+	hostport := os.Getenv("TEST_REDIS_HOSTPORT")
+	if hostport == "" {
+		hostport = "localhost:6379"
+	}
+	password := os.Getenv("TEST_REDIS_PASSWORD")
+	if password == "" {
```
Zhaars, line 17: Looks like redundant if block

cmd/redis.go
```diff
@@ -122,6 +132,18 @@ func ParseRedisCLIParametersFromFlags(flags *flag.FlagSet, prefix string) *Redis
 		}
 		redisOptions.DBTokens = val
 	}
+
+	if f := flags.Lookup(prefix + "redis_tls_enable"); f != nil {
+		getter, ok := f.Value.(flag.Getter)
+		if !ok {
+			log.Fatal("Can't cast flag's Value to Getter")
+		}
+		val, ok := getter.Get().(bool)
```
Zhaars, line 141: We can use `ParseBool` here I think

cmd/redis_testutils.go
Lagovas: yes. removed and left comment

cmd/redis.go
Lagovas: agree, missed it

Zhaars (approved):
Cool

PR #568, opened 2022-08-23
Implement Hashicorp Cousul encryptor config loading

Zhaars:
Implementation of loading encryptor_config from Hashicorp Consul. Approach with global registry map and build flag was used to achieve flexibility.

## Checklist

  with new changes


encryptor/config_loader/consul/consul_cli.go
```diff
@@ -0,0 +1,44 @@
...
+	EncryptorConfigPath string
+}
+
+// RegisterCLIParametersWithFlagSet look up for consul_connection_api_string, if none exists, consul_connection_api_string and consul_kv_config_path
+// will be added to provided flags.
+func RegisterCLIParametersWithFlagSet(flags *flag.FlagSet, prefix, description string) {
+	if description != "" {
+		description = " (" + description + ")"
+	}
+	// TODO: add tls support
```
Lagovas, line 24: I think we should not accept pure connections, only tls. At least without mutual auth. Because config file is important and configure security controls.

encryptor/config_loader/consul/consul_cli.go
```diff
@@ -0,0 +1,44 @@
...
+	if flags.Lookup(prefix+consulConnectionStringFlag) == nil {
+		flags.String(prefix+consulConnectionStringFlag, "", "Connection string (http://x.x.x.x:yyyy) for loading encryptor config from Consul"+description)
+		flags.String(prefix+"consul_kv_config_path", defaultConsulConfiPath, "KV Encryptor Config Path (acra/encryptor_config) for loading encryptor config from Consul"+description)
```
Lagovas, line 27: why connection string referenced by value, but config path by hardcoded string?

encryptor/config_loader/consul/storage.go
```diff
@@ -0,0 +1,157 @@
...
+// Remove implementation of filesystem.Storage interface
+func (s Storage) Remove(path string) error {
+	_, err := s.client.Delete(path, nil)
+	return err
+}
+
+// Stat implementation of filesystem.Storage interface
+func (s Storage) Stat(path string) (os.FileInfo, error) {
+	//TODO implement me
+	panic("implement me")
```
Lagovas, line 102: I think we can return info about value. Can be useful size of value.

encryptor/config_loader/consul/storage.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/cossacklabs/acra/encryptor"
+	"github.com/hashicorp/consul/api"
+	log "github.com/sirupsen/logrus"
+)
+
+// StorageCreator implement config_loader.EncryptorConfigStorage via Hashicorp Consul Backend
+type StorageCreator struct{}
+
+// StorageConfigured check weather CLI flag for Consul using was provided
+func (s StorageCreator) StorageConfigured(flags *flag.FlagSet, prefix string) bool {
```
Lagovas, line 18: Lets use some question in the method name to show that it returns boolean? IsConfigured, DoesConfigured, etc.

go.mod
```diff
@@ -14,11 +14,14 @@ require (
 	github.com/go-sql-driver/mysql v1.5.0
 	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e
 	github.com/golang/protobuf v1.5.2
+	github.com/hashicorp/consul/api v1.3.0
 	github.com/hashicorp/vault/api v1.3.0
 	github.com/jackc/fake v0.0.0-20150926172116-812a484cc733 // indirect
 	github.com/jackc/pgx v3.6.2+incompatible
 	github.com/jackc/pgx/v4 v4.14.1
 	github.com/lib/pq v1.10.4
+	github.com/philippgille/gokv v0.6.0
```
Lagovas, line 23: where it used in this pr?

Zhaars: I think its not used

encryptor/config_loader/consul/consul_cli.go
Zhaars: Added possibility to have pure/mTLS/TLS connection as we discussed.

Lagovas: hm, maybe we can register tls related flags using new [RegisterTLSArgsForService](https://github.com/cossacklabs/acra/blob/master/network/tls_wrapper.go#L115) method? It will register all flags related to TLS including CRL/OCSP and allow to get TLS config with configured certificate verifier

Zhaars: Makes sense!

tests/ssl/generate_tls_keys.sh
```diff
@@ -109,13 +109,11 @@ gen_crl() {
         -out "${OUT_DIR}/$signer/crl.pem"
 }
 
-declare -a names=(mysql postgresql acra-writer acra-writer-2 
-    acra-writer-revoked acra-server ocsp-responder intermediate-ca vault 
-    acra-client redis)
+declare -a names=(consul)
```
Lagovas, line 112: I think it was changed for local generation?

tests/ssl/generate_tls_keys.sh
```diff
@@ -109,13 +109,11 @@ gen_crl() {
...
 for name in "${names[@]}"; do
     gen_cert ca $name
 
-    if [[ "$name" = "vault" ]]; then
+    if [ "$name" = "vault" ] || [ "$name" = "consul" ] ; then
```
Lagovas, line 116: for what we generate PEM certs too? because of python's clients require it? if so then can we leave comment here for what we do that only for them?

tests/ssl/generate_tls_keys.sh
Zhaars: Yes, you are right. Thanks for noticing!

tests/ssl/generate_tls_keys.sh
Zhaars: Python client require a simple crt format. Its required for Vault/Consul servers to configure, anyway will let the comment.

Lagovas (approved):

PR #574, opened 2022-09-01
Remove zone support

Lagovas:
Here is removed acra-addzone tool, all zonemode related things from code, tests, benchmarks, docker files. 
Some of interfaces updated to support extra data `additional context` to be able to use AD (additional data) in AEAD (authenticated encryption with additional data) encryption. It is not used for now but left args for future extensions if we will want. But it is discussible do we need to leave it now...

It's draft because there is remaining changes in our `./examples` and `acrawriter` wrappers. But main removings already made.

P.S. I tried to split changes into separate commits due to a lot of changes. 

## Checklist

  with new changes


benchmarks/docker-compose.pgsql-ssl-server-ssl.yml
```diff
@@ -40,7 +39,6 @@ services:
         build:
             context: ../
             dockerfile: docker/postgresql-ssl.dockerfile
-        network_mode: host
```
Zhaars, line 43: What is the reason for this change? As I know default `network_mode` in compose if `bridge` means that containers will be placed in the compose bridge network and they will not be accessible from the host.  So this removing changes the compose behavior - I am just trying to understand whether it isn`t breaking anything.

Lagovas: my docker-compose:
```
$ docker-compose -v
docker-compose version 1.29.2
```
tells me: `docker.errors.InvalidArgument: "host" network_mode is incompatible with port_bindings`
As I think, it is not valid anymore if you declare ports (that we do in this file). I tested locally that ports are accessible from the host and they are.

Lagovas:
We decided to merge this PR in next release after deprecating Zones.
Additionaly, we should re-check our eng demos with these changes due to renaming several files in examples. Just a reminder for me.

PR #578, opened 2022-09-09
Unify vault tls flags

Zhaars:
Unify/add common Acra TLS flags for Vault.

## Checklist

  with new changes

### DIFF
--- a/configs/acra-addzone.yaml
+++ b/configs/acra-addzone.yaml
@@ -86,6 +86,9 @@ vault_connection_api_string:
 # KV Secret Path (secret/) for reading ACRA_MASTER_KEY from HashiCorp Vault
 vault_secrets_path: secret/
 
+# Path to CA certificate for HashiCorp Vault certificate validation (deprecated since 0.94.0)
+vault_tls_ca_path: 
+
 # Set authentication mode that will be used in TLS connection. Values in range 0-4 that set auth type (https://golang.org/pkg/crypto/tls/#ClientAuthType). Default is -1 which means NotSpecified and will be used value from tls_auth.
 vault_tls_client_auth: -1
 
--- a/configs/acra-addzone.yaml
+++ b/configs/acra-addzone.yaml
@@ -86,14 +86,47 @@ vault_connection_api_string:
 # KV Secret Path (secret/) for reading ACRA_MASTER_KEY from HashiCorp Vault
 vault_secrets_path: secret/
 
-# Path to CA certificate for HashiCorp Vault certificate validation
-vault_tls_ca_path: 
+# Set authentication mode that will be used in TLS connection. Values in range 0-4 that set auth type (https://golang.org/pkg/crypto/tls/#ClientAuthType). Default is -1 which means NotSpecified and will be used value from tls_auth.
+vault_tls_client_auth: -1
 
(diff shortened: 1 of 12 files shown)


configs/acra-addzone.yaml
```diff
@@ -86,15 +86,48 @@ vault_connection_api_string:
 # KV Secret Path (secret/) for reading ACRA_MASTER_KEY from HashiCorp Vault
 vault_secrets_path: secret/
 
-# Path to CA certificate for HashiCorp Vault certificate validation
-vault_tls_ca_path:
```
Lagovas, line 90: unfortunately, it is breaking change and we should leave it. and probably deal with it as we do for acra-server, with common parameter used until specified for explicit side (client/database)

keystore/keyloader/hashicorp/vault_cli.go
```diff
@@ -36,11 +41,12 @@ func RegisterCLIParametersWithFlagSet(flags *flag.FlagSet, prefix, description s
 	if flags.Lookup(prefix+vaultConnectionStringFlag) == nil {
 		flags.String(prefix+vaultConnectionStringFlag, "", "Connection string (http://x.x.x.x:yyyy) for loading ACRA_MASTER_KEY from HashiCorp Vault"+description)
 		flags.String(prefix+"vault_secrets_path", defaultVaultSecretsPath, "KV Secret Path (secret/) for reading ACRA_MASTER_KEY from HashiCorp Vault"+description)
-		flags.String(prefix+"vault_tls_ca_path", "", "Path to CA certificate for HashiCorp Vault certificate validation"+description)
```
Lagovas, line 39: as mentioned above, we should leave it because it is breaking change. and deal with it too.

configs/acra-addzone.yaml
Zhaars: Great notion, agree!

keystore/keyloader/hashicorp/vault_cli.go
```diff
@@ -56,51 +68,106 @@ func ParseCLIParametersFromFlags(flags *flag.FlagSet, prefix string) *VaultCLIOp
...
+	}
+
+	if options.EnableTLS {
+		serverName := network.SNIOrHostname(options.SNI, vaultURL.Host)
+		tlsConfig, err := network.NewTLSConfig(serverName, options.CAPath, options.ClientKey, options.ClientCert, options.Auth, options.CertVerifier)
+		if err != nil {
+			return nil, err
+		}
+
+		transport.TLSHandshakeTimeout = 10 * time.Second
```
Lagovas, line 145: is there any reason to use exactly 10 seconds or we can use our common value https://github.com/cossacklabs/acra/blob/master/network/deadline_listener.go#L27 1 minute?

Zhaars: I digged into hashicorp lib and found that time, decided to make the same.

https://github.com/hashicorp/vault/blob/main/api/client.go#L230

What do you think is better?

Lagovas: oh, I didn't note that it is for TLSHandshake and they use 60sec for timeout too... okay, lets leave 10sec for handshake.

Lagovas (approved):

PR #580, opened 2022-09-19
Process sequence of pending packets

Lagovas:
What was the problem? DB drivers can send set of packets in one TCP packet. And usually, they are related to each other and look like Parse + Bind + Describe + Execute + Sync. Driver registers query in Parse, then binds data, then asks to describe types, executes query and commit state. Acra registers PreparedStatement on ParseComplete packet that DB responds to driver after validation and acceptance. Bind packet relates to Parse packet because it should specify statement name for which it sends values.
In this flow acra do next: remembers Parse packet and save it as `pendingParse`. Then handle Bind packet, loads finds `pendignParse` and link them together. 

Previously we expected that one set of packets in TCP packet contains only one packet with a query like SimpleQuery or Parse packet. And when Acra receives ParsePacket, it resets state, forgets about Bind and other kind of packets - https://github.com/cossacklabs/acra/blob/master/decryptor/postgresql/protocol.go#L121

But drivers not limited in the number of query packets. They can send several of them in one set. And they should be processed in same order. If acra got 2 Describe packets from driver to database and then receives 2 RowDescription responses from the database, it should link first RowDescription to the first Describe, second one to the second. 
So Acra should not store only last packet, but should all of them in same order.

In this PR was added new component that stores all pendingPackets and store them separately according to the type, works as a queue. And was refactored logic with working with pending packets. Now it removes item from the queue, not nilify a field.

Due to limitations of python's drivers that don't allow to compile such set of packets to write regression test, were added unit-test with dumped packets that reproduces the problem. 

## Checklist

  with new changes

### DIFF
--- a/decryptor/postgresql/pg_decryptor.go
+++ b/decryptor/postgresql/pg_decryptor.go
@@ -376,7 +376,11 @@ func (proxy *PgProxy) handleQueryPacket(ctx context.Context, packet *PacketHandl
 }
 
 func (proxy *PgProxy) handleBindPacket(ctx context.Context, packet *PacketHandler, logger *log.Entry) (bool, error) {
-	bind := proxy.protocolState.PendingBind()
+	bind, err := proxy.protocolState.PendingBind()
+	if err != nil {
+		logger.WithError(err).Error("Can't get pending Bind packet")
+		return false, nil
+	}
 	logger = logger.WithField("portal", bind.PortalName()).WithField("statement", bind.StatementName())
 	logger.Debug("Bind packet")
 	// There must be previously registered prepared statement with requested name. If there isn't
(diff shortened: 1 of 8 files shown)


decryptor/postgresql/pg_decryptor.go
```diff
@@ -869,8 +892,12 @@ func (proxy *PgProxy) handleQueryDataPacket(ctx context.Context, packet *PacketH
 		}
 		// default values Text
 		format := 0
-		if proxy.protocolState.pendingBind != nil {
-			boundFormat, err := GetParameterFormatByIndex(i, proxy.protocolState.pendingBind.resultFormats)
+		pendingBind, err := proxy.protocolState.pendingPackets.GetPendingPacket(&BindPacket{})
+		if err != nil {
+			panic(err)
```
Zhaars, line 897: Is it expected behavior?

Zhaars, line 897: Is it expected behavior??

decryptor/postgresql/pg_decryptor.go
```diff
@@ -376,7 +382,11 @@ func (proxy *PgProxy) handleQueryPacket(ctx context.Context, packet *PacketHandl
 }
 
 func (proxy *PgProxy) handleBindPacket(ctx context.Context, packet *PacketHandler, logger *log.Entry) (bool, error) {
-	bind := proxy.protocolState.PendingBind()
+	bind, err := proxy.protocolState.LastBind()
```
Zhaars, line 385: As I see `LastBind()` could return nil packet with nil error. Just make sure it's a safe operation.

Lagovas: added nil checks

Lagovas (commented):
@Zhaars , added nil checks and replaced panic with returning error

Zhaars (approved):
Cool, thanks.

PR #584, opened 2022-09-27
Extend returning statement

Lagovas:
In this PR were removed limitation that RETURNING statements should contain only *ColName structs and allow other types, like sql literals which ignored by acra-server. According to [postgresql's](https://www.postgresql.org/docs/current/dml-returning.html) and [mysql's](https://mariadb.com/kb/en/insertreturning/) docs, `returning` may contain same expressions that is supported by SELECT. So, were updated sql's grammar to be compatible with SelectExprs.

Due to mariadb supports returning since [10.5.0](https://mariadb.com/kb/en/insertreturning/), additionally was updated image for mariadb-ssl to the fresh one 10.7.1 and removed skip condition in integration tests that didn't run tests at all for mariadb with `returning` cases. Because base class BaseTestCase skipped non-tls tests.

After fixing running integration tests for mariadb and updating image, I found that our index.txt contains serial from old certs, and current certs tests/ssl/mysql/mysql.crt has another one. Which were generated for engineering-demo, copied there and left here as is without updating index.txt. But our current mariadb-ssl uses old certs with serial saved in index.txt. So I copied correct certs from current cossacklabs/mariadb-ssl:10.3-1 image and placed here. Otherwise acra will deny certs in OCSP checks due to mismatch serial in index.txt and current mysql.crt

## Checklist

  with new changes


Zhaars (approved):
lgtm

PR #589, opened 2022-10-05
Allow run integration tests with non-localhost databases

Lagovas:
What has been done:
* generated new certs *-buildbot.(key|crt) for our buildbot environment where python tests connect to docker containers by hostname and cannot connect to localhost. Our current certificates for postgresql/mysql don't include additional SAN value `mysql`/`postgresl`, only `localhost` and used in our mysql-ssl and postgresql-ssl images. So to not overwrite and break images, new separate for buildbot environment were added.
* due to in this environment tests connect to DB on another host, not `localhost`, was found a lot of problems when our integration tests work correctly only when database accessed via localhost even with explicitly specified `TEST_DB_HOST`. It because some of functions uses same connection string for acra-server and for database. Acra-server forked on same host where tests run but database already started on different host. So, in such cases we specify explicitly `localhost` for acra instances and use `TEST_DB_HOST` env variable for other.
* refactored AsyncPG executor to use new created loop for async functions and correct closing connections after communication with databases. It fixes a lot of warnings like `unclosed transport ...` from asyncio internals. Async logic moved to inner function that encapsulates all calls which require async/await statements and allows to execute it with the loop instance once.

## Checklist

  with new changes


Zhaars (approved):
lgtm

Issue #587, opened 2022-10-03
[ISSUE] Acra throws errors on tables with columns wrapped with double quotes


Machado117:
**Describe the bug**
Errors can be found on acra server logs when creating tables with columns wrapped with double quotes. Also, when executing select queries that use columns wrapped with double quotes, encrypted fields won't be decrypted.

**To Reproduce**
1. Create table `CREATE TABLE mytable ("NAME" bytea, "AGE" integer);`

Acra logs:
```
2022/10/03 16:49:32 ignoring error parsing DDL 'CREATE TABLE mytable ("NAME" bytea, "AGE" integer);': syntax error at position 29 near 'NAME'
2022/10/03 16:49:32 ignoring error parsing DDL 'CREATE TABLE mytable ("NAME" bytea, "AGE" integer);': syntax error at position 29 near 'NAME'

```
2. Insert some data `INSERT INTO mytable ("NAME", "AGE") VALUES ('John', 1);`
3. Execute a select query `SELECT * FROM mytable WHERE "AGE" = 1;`
Acra logs:
```
2022/10/03 16:49:50 ignoring error of non parsed sql statement
2022/10/03 16:49:50 ignoring error of non parsed sql statement

```
Query result:
```
    NAME    | AGE 
------------+-----
 \x4a6f686e |   1
(1 row)

```
**Expected behavior**
Acra should parse these queries and decrypt the field

**Acra configuration files**
For AcraServer: 
    - [ ] `encryptor_config.yaml`.
```
  - table: mytable
    columns:
      - NAME
      - AGE
    encrypted:
      - column: NAME
        data_type: str
```

**Environment (please complete the following information):**
- Acra version: 492568a7929d3cd615b6c8160c7ad266b26a2515
- Database server and its version: PostgreSQL 9.6.10
- Installed components:
   - [ ] AcraServer
- Data-in-transit encryption between Acra and the client-side application:
   - [ ] no transport encryption
- Installation way:
   - [ ] via Docker

Lagovas:
Hello, nice catch. I reproduced this error and we will fix it soon. AcraServer supports columns wrapped in double quotes but in this case, sqlparser incorrectly matched this as string literal instead of column name for PostgreSQL dialect. Due to pgsql doesn't allow strings in double quotes, it was handled as error.


[...]

PR #590, opened 2022-10-05
Support PostgreSQL double qouted columns

Zhaars:
This PR contains only `sqlparser` changes to fix the open [issue](https://github.com/cossacklabs/acra/issues/587). During debugging we faced that sqlparser flow has incorrect flow for parsing statements as `WHERE "column" = 'value'` as it looks on `value` rule instead of `column_name`. 

In this PR two new types were introduced 

- `column_name_ext` which is almost as `column_name` but has primary DOUBLE_QUOTE_STRING rule to track double quoted column only for PostgreSQL. 
- `column_name_expression` -  which is the same as `value_expression` but used `column_name_ext` instead of `column_name`.

## Checklist

  with new changes


sqlparser/parse_test.go
```diff
@@ -211,6 +211,20 @@ var (
...
+		// postgres allow to use double quote string for columns
+		dialect: postgresql.NewPostgreSQLDialect(),
+	}, {
+		input:  `select * from mytable where "test" = "test"`,
+		output: `select * from mytable where 'test' = 'test'`,
+	}, {
+		input:  `select * from mytable where "test" = 1 and 'value' = 'value'`,
+		output: `select * from mytable where 'test' = 1 and 'value' = 'value'`,
+	}, {
+		input:   `SELECT "id", "landline_number" AS "landlineNumber", "removal" FROM "users" AS "User" where "User"."is_active"`,
```
Lagovas, line 225: lets add some somparison too: `WHERE "User"."AGE" = 123` and simple `"AGE"='123'` without table name.

Lagovas: plus I remember that you mentioned some integration tests with double quotes. Can we extend them with additional cases that we caught?

sqlparser/parse_test.go
```diff
@@ -215,6 +215,11 @@ var (
 		input: `select * from mytable where "AGE" = 1 and "TEST" = 'test'`,
 		// postgres allow to use double quote string for columns
 		dialect: postgresql.NewPostgreSQLDialect(),
+	}, {
+		input:  `insert into some_table(id, data) VALUES (10918, "test")`,
```
Lagovas, line 219: lets add comment, that it is temporary valid for MySQL default settings, but invalid for PostgreSQL & MySQL with ANSI mode, and probably will be changed. It will help in the future to understand why this case will fail after future changes)

Zhaars: Sure, also added TODO in `invalidSQL` tests that such query should be tests on failure

Zhaars:
@Lagovas I added a couple of unit tests to cover correct matching with tables from config. Please let me know if you want more tests to be added.

encryptor/searchable_query_filter_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/cossacklabs/acra/encryptor/config"
+	"github.com/cossacklabs/acra/sqlparser"
+)
+
+func TestGetTableSchemaOfColumnMatchConfiTable(t *testing.T) {
```
Lagovas, line 11:
```suggestion
func TestGetTableSchemaOfColumnMatchConfigTable(t *testing.T) {
```

encryptor/searchable_query_filter_test.go
```diff
@@ -0,0 +1,74 @@
...
+	tableName := "some_table"
```
Lagovas, line 12: can we use name with upper-cased characters here and below? to be sure that we handle it correctly. Because previously we handled it in lowercase by default

sqlparser/ast_test.go
```diff
@@ -535,6 +536,35 @@ func TestTableIdentMarshal(t *testing.T) {
 	}
 }
 
+func TestTableIdentValueForConfig(t *testing.T) {
+	str := TableIdent{
+		quote: 34,
+		v:     "table",
+	}
+	b, err := json.Marshal(str)
```
Lagovas, line 544: why do we use json package instead of sqlparser.String?

Zhaars: Just copied from another test.

tests/test.py
```diff
@@ -7250,6 +7250,45 @@ def testRotatedKeysAcraBlock(self):
...
+
+    def testSearchDoubleQuotedTable(self):
+        context = self.get_context_data()
+        search_term = context['searchable']
+
+        # Insert searchable data and some additional different rows
+        self.insertRow(context)
+        self.insertDifferentRows(context, count=5)
+
+        query = 'SELECT * FROM "test_searchable_transparent_encryption" WHERE searchable = :searchable'
```
Lagovas, line 7275: maybe lets use `"searchable" = ...`? to test wrapped column too?

Lagovas (approved):

PR #592, opened 2022-10-12
Extend searchable logic with joins

Zhaars:
This PR contains some searching logic extensions related to supporting JOIN queries in search queries. 

What we would like to support:
- The same ComprasionExpression as currently supported but in `JOIN ON` statements;
- Current searchable logic expect `SQLVal` in comparison expression to be able to calculate some hash, so the following JOINS queries `JOIN table t ON  t.search_column = t2.some_column` is not supported with searchable encryption. The corresponding log warning was added that can simplify debugging in such scenarios.
- However, queries like `JOIN table t ON  t.search_column_1 = t2.searchable_column_2` are suitable only for searchable encryption, so in such scenarios we need to transform the query to the following ` JOIN table t ON  substring(t.search_column_1, 1, <HMAC_size>) = substring(t2.searchable_column_2, 1, <HMAC_size>)`

## Checklist

  with new changes


encryptor/searchable_query_filter.go
```diff
@@ -123,39 +128,31 @@ func (filter *SearchableQueryFilter) filterTableExpressions(statement sqlparser.
...
 			return nil
 		}
+
 		exprs = append(exprs, comparisonExprs...)
 	}
-	return exprs
-}
 
-func (filter *SearchableQueryFilter) filterComparisons(exprs []*sqlparser.ComparisonExpr, defaultTable *AliasedTableName, aliasedTables AliasToTableMap) []SearchableExprItem {
-	filtered := make([]SearchableExprItem, 0, len(exprs))
```
Lagovas, line 147: why you decided to remove pre-allocation? with allocation empty slice golang will create a new buffer with 2 elements and after that increase further by power of 2. 2...4...8...16. So, we can meet 1-3 new allocation with copying current elements to new one, which looks inefficient.

encryptor/searchable_query_filter.go
```diff
@@ -169,11 +166,22 @@ func (filter *SearchableQueryFilter) filterComparisons(exprs []*sqlparser.Compar
...
 	return filtered
 }
 
+func (filter *SearchableQueryFilter) getColumnSetting(column *sqlparser.ColName, defaultTable *AliasedTableName, aliasedTables AliasToTableMap) config.ColumnEncryptionSetting {
+	schema := filter.getTableSchemaOfColumn(column, defaultTable, aliasedTables)
+	if schema == nil {
+		return nil
+	}
+	// Also leave out those columns which are not searchable.
+	columnName := column.Name.String()
```
Lagovas, line 179: should we use here `ValueForConfig()` instead of `Name.String()`?

encryptor/searchable_query_filter.go
```diff
@@ -56,36 +56,41 @@ func NewSearchableQueryFilter(schemaStore config.TableSchemaStore, mode Searchab
...
-func (filter *SearchableQueryFilter) FilterSearchableComparisons(statement sqlparser.Statement) []SearchableExprItem {
+func (filter *SearchableQueryFilter) FilterSearchableComparisons(statement sqlparser.Statement) map[string]SearchableExprItem {
 	tableExps, err := filter.filterTableExpressions(statement)
 	if err != nil {
 		logrus.Debugln("Unsupported search query")
 		return nil
 	}
 
-	defaultTable, aliasedTables := filter.filterInterestingTables(tableExps)
+	defaultTables, aliasedTables := filter.filterInterestingTables(tableExps)
```
Lagovas, line 66: actually, it is not default tables. Previously one table was return due to only one table in the `WHERE` statement can be without aliases if somewhere columns addressed without table names or aliases. For example:
`SELECT tbl1Col1, tbl2Col2 FROM table1, table2`. It is invalid query if `tbl1Col1` and `tbl2Col2` are table1's columns. If `tbl2Col2` is column of `table2`, than it is invalid and should be accessed as `table2.tbl2Col2` or using aliases, like `SELECT tbl1Col1, tbl2.tbl2Col2 FROM table1, table2 tbl2`.
Do we handle it now after these changes?

hmac/decryptor/hashQuery_test.go
```diff
@@ -204,3 +204,99 @@ func TestSearchableWithTextFormat(t *testing.T) {
...
+      - column: data1
+        searchable: true
+`
+
+	type testcase struct {
+		Query string
+	}
+
+	testcases := []testcase{
+		{Query: "SELECT * FROM table1 t1 inner join test_table t2 inner join test_table_2 t3 on t2.data1=t3.data1"},
```
Lagovas, line 247: can you add cases with tables without aliases? join between tables with alias and without? with full qualifiers with table names like `test_table.data1 = t2.tdata1`? to cover different cases

encryptor/searchable_query_filter.go
Zhaars: You definitely right, just change the function semantics and forgot to init slice in proper way.

encryptor/searchable_query_filter.go
```diff
@@ -91,17 +91,36 @@ func (filter *SearchableQueryFilter) filterInterestingTables(fromExp sqlparser.T
 	if len(tables) == 0 {
 		return nil, nil
 	}
+
+	var defaultTable = tables[0]
+	var defaultTableName string
+	// if query contains table without alias we need to detect default table
+	// if no, we can ignore default table and AliasToTableMap will be used to map ColName with encryptor_config
```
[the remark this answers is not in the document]
Zhaars: Yes, we can return

pseudonymization/tokenizeQuery.go
```diff
@@ -179,17 +179,24 @@ func (encryptor *TokenizeQuery) replaceValuesWithTokenizedData(ctx context.Conte
...
-		}
-		if err != nil {
-			logrus.WithError(err).Debugln("Failed to tokenize value")
-			return nil, err
+			clientID := setting.ClientID()
+			if len(clientID) > 0 {
+				logger.WithField("client_id", string(clientID)).Debugln("Tokenize with specific ClientID for column")
+			} else {
+				logger.WithField("client_id", string(accessContext.GetClientID())).Debugln("Tokenize with ClientID from connection")
+				clientID = accessContext.GetClientID()
```
Lagovas, line 195: o_O. As I understand, we don't test it anywhere?

Lagovas (commented):
lgtm for now. but lets add some integration tests with joins over searchable fields

pseudonymization/tokenizeQuery.go
Zhaars: Yeap:) Lets add corresponded integration tests too.

hmac/decryptor/hashQuery.go
```diff
@@ -108,6 +108,20 @@ func (encryptor *HashQuery) OnQuery(ctx context.Context, query base.OnQueryObjec
 			continue
 		}
 
+		// TODO: left comment
```
Lagovas: and will be great to leave example of result expression

hmac/decryptor/hashQuery.go
```diff
@@ -99,6 +99,33 @@ func (encryptor *HashQuery) OnQuery(ctx context.Context, query base.OnQueryObjec
...
+				From: sqlparser.NewIntVal([]byte{'1'}),
+				To:   sqlparser.NewIntVal(hashSize),
+			}
+			continue
+		}
+
+		// MySQL have ambiguous behaviour with filtering over search data
+		// query like `select table1.value, table2.value from table1 join table2 on substr(table1.searchable, 1, 33) = substr(table2.searchable, 1, 33)
+		// where substr(table1.searchable, 1, 33) = X'7f6002a9335e723661b917736c3d253c07c65750839b9952801ab7f6e2a4982792'`
+		// doesnt return any record, but it does work separately (with just single where search or with join over search data)
```
Lagovas, line 114:
```suggestion
		// doesn't return any record, but it does work separately (with just single where search or with join over search data)
```

hmac/decryptor/hashQuery.go
```diff
@@ -99,6 +99,33 @@ func (encryptor *HashQuery) OnQuery(ctx context.Context, query base.OnQueryObjec
...
+		// to escape from this ambiguity added explicit casting search hash to bytes
```
Lagovas, line 115: lets add comment what the result we will get at the end? that we change `substr(colname, ...) = X'value'` to `convert(substr(colname, ...), binary) = 0xFFFFF`

tests/test.py
```diff
@@ -7246,6 +7249,117 @@ def testRotatedKeysAcraBlock(self):
...
+                self.encryptor_table.c.number,
+                self.encryptor_table_join.c.searchable,
+                self.encryptor_table_join.c.default_client_id,
+                self.encryptor_table_join.c.raw_data,
+                self.encryptor_table_join.c.specified_client_id,
+                self.encryptor_table_join.c.zone_id,
+                self.encryptor_table_join.c.empty,
+            ).
+            where(self.encryptor_table.c.searchable == sa.bindparam('searchable')).
+            join(self.encryptor_table_join, self.encryptor_table_join.c.searchable==sa.bindparam('searchable')),
```
Lagovas, line 7302: can we place `where` below `join` expression? because it is unusual to read sqlalchemy code written in different order than real SQL )

tests/test.py
```diff
@@ -7246,6 +7249,117 @@ def testRotatedKeysAcraBlock(self):
...
+            {'searchable': get_pregenerated_random_data().encode('utf-8')})
+        self.assertEqual(len(rows), 0)
+
+        # test join with on table1.searchable = table2.searchable
+        rows = self.executeSelect2(
+            sa.select(self.encryptor_table.c.number, self.encryptor_table_join.c.searchable).
+            where(self.encryptor_table.c.searchable == sa.bindparam('searchable')).
+            join(self.encryptor_table_join, self.encryptor_table.c.searchable==self.encryptor_table_join.c.searchable),
```
Lagovas, line 7334: can we test raw join without `where`? or as I understand, we test same successful case as above?

tests/test.py
```diff
@@ -7613,6 +7727,56 @@ def testSearchableTokenizationDefaultClientID(self):
...
+            'token_i32': specific_client_id_table.c.token_i32,
+            'token_i64': specific_client_id_table.c.token_i64,
+            'token_str': specific_client_id_table.c.token_str,
+            'token_bytes': specific_client_id_table.c.token_bytes,
+            'token_email': specific_client_id_table.c.token_email,
+        }
+        # data owner take source data
+        for key in columns:
+            parameters = {key: data[key]}
+            query = sa.select(specific_client_id_table).where(columns[key] == data[key])
```
Lagovas, line 7771: can you remind, do we support joinable tokenization?

tests/test.py
Zhaars: Sure!

tests/test.py
Zhaars: Added separate query without `where`

tests/test.py
```diff
@@ -7727,6 +7737,76 @@ def testSearchableTokenizationDefaultClientID(self):
...
+                default_client_id_table.c.token_i32,
+                default_client_id_table.c.token_str,
+                default_client_id_table_join.c.token_i64,
+                default_client_id_table_join.c.token_email,
+            ).join(default_client_id_table_join, columns[key] == data[key])
+
+            parameters = {key: data[key]}
+            source_data = self.fetch_from_1(query, parameters, literal_binds=False)
+            for k in ('token_i32', 'token_i64', 'token_str', 'token_email'):
+                if isinstance(source_data[0][k], (bytearray, bytes)) and isinstance(data[k], str):
```
Lagovas, line 7805: hm, is there we get values with `bytes` and `bytearray` types? we compare `str` and `email` types. As I understand, database should return text types and driver should deal with them appropriately. do we really get here these types instead of python's str?

tests/ee_searchable_tokenization_config.yaml
```diff
@@ -26,6 +26,33 @@ schemas:
...
+      - empty
+      - token_i32
+      - token_i64
+      - token_str
+      - token_bytes
+      - token_email
+    encrypted:
+      - column: token_i32
+        token_type: int32
+        consistent_tokenization: true
```
Lagovas, line 42: idea for the future: support specifying `consistent_tokenization` in the `default` section once and apply by default )) can you create task for that, please?)

Zhaars: Done `T2674` created

tests/test.py
Zhaars: Yes, you are right, the main idea it some data returned from the MySQL driver comes in bytes format so we need to have additional checks. Such checkings are in a lot of our tests.

Lagovas (approved):

PR #594, opened 2022-10-25
Fixed bug with matching default table with alias

Zhaars:
Currently, Acra expects that columns in `SELECT` statement haven't been aliased if FROM has a non-aliased table or they should have proper alias. But databases allow specific alias for table but use columns without an alias.

So if the query contains Columns without aliases and the table is aliased, Acra should consider this table as default. Added checking for containing aliases in tables expressions and switch between extracting tables functions depending on queries.

## Checklist

  with new changes

### DIFF
--- a/encryptor/utils_test.go
+++ b/encryptor/utils_test.go
@@ -1,7 +1,6 @@
 package encryptor
 
 import (
-	"github.com/cossacklabs/acra/sqlparser/dialect/postgresql"
 	"testing"
 
 	"github.com/cossacklabs/acra/decryptor/base/mocks"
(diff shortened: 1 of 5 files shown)


encryptor/utils_test.go
```diff
@@ -103,6 +107,64 @@ inner join table6 on table6.col1=t1.col1
 			}
 		}
 	})
+	t.Run("with aliased table and non-aliased colum name", func(t *testing.T) {
+		testcases := []struct {
+			query   string
+			dialect dialect.Dialect
+		}{
+			{
+				query:   `SELECT "id", "email", "mobile_number" AS "mobileNumber" FROM "users" AS "User" where "User"."is_active"`,
```
Lagovas, line 116: lets use a bit more complicated queries with several aliased tables in the `FROM` statement to be sure that we use first from the list

Lagovas (approved):

PR #596, opened 2022-11-02
Extend type awareness with DatTypeIDs

Zhaars:
Extending functionality of TA - added ability to directly specify `data_type_db_identifier` in encryptor config. Backward compatibility is saved as `data_type` is mapped to DataTypeID in the encryptor config. 

The idea is to have `DataTypeEcoder` interface implemented for each supported DB type which registers in `DataTypeIDEncoders` map. In such case, we can achieve more flexibility by substituting/adding new types

## Checklist

  with new changes


decryptor/postgresql/data_encoder_test.go
```diff
@@ -151,15 +162,15 @@ func TestTextMode(t *testing.T) {
 		// decoder expects valid string and pass as is, so no errors. but on encode operation it expects valid int literal
 		{input: []byte("some data"), decodedData: []byte("some data"), encodedData: []byte("some data"),
 			decodeErr: nil, encodeErr: nil,
-			setting:    &config.BasicColumnEncryptionSetting{TokenType: "int32", DataType: "int32"},
+			setting:    &config.BasicColumnEncryptionSetting{TokenType: "int32", DataType: "int32", DataTypeID: pgtype.Int4OID},
```
Lagovas, line 165: hm... I think we should not allow to specify both values: data_type and data_type_oid. IMHO users should specify only one of them. To avoid misconfiguration and so on. and as I see, we have special error `ErrDataTypeWithDataTypeID` for that. But here, as I see, we have decodeErro and encodeErr.

encryptor/config/schemaStore_test.go
```diff
@@ -1189,3 +1192,55 @@ func TestTokenizedDeprecationWarning(t *testing.T) {
 		t.Fatal("warning is not found but expected")
 	}
 }
+
+func TestWithDataTypeIDOption(t *testing.T) {
```
Lagovas, line 1196: can we check cases with invalid values?

decryptor/postgresql/data_encoder.go
```diff
@@ -3,15 +3,15 @@ package postgresql
 import (
 	"context"
 	"encoding/binary"
-	"strconv"
 
 	"github.com/cossacklabs/acra/decryptor/base"
+	"github.com/cossacklabs/acra/decryptor/base/type_awareness"
+	"github.com/cossacklabs/acra/decryptor/postgresql/types"
+	_ "github.com/cossacklabs/acra/decryptor/postgresql/types"
```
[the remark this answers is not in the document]
Zhaars: Firstly it was only shadow import to be able to register `DataTypeIDEncoder`, but then added direct import. Will remove duplicate imports.

encryptor/config/encryptionSettings.go
```diff
@@ -322,25 +358,58 @@ func (s *BasicColumnEncryptionSetting) Init() (err error) {
...
 			return err
 		}
 	}
 
+	if s.DataTypeID != 0 {
+		s.settingMask |= SettingDataTypeIDFlag
+
+		var dataTypeIDEncoders = type_awareness.GetPostgreSQLDataTypeIDEncoders()
+		if useMySQL {
+			//dataTypeIDEncoders = base.GetMySQLDataTypeIDEncoders()
```
Lagovas, line 379: lets return error and fail validation due to unsupported on MySQL?

Zhaars: Discussed leaving this PR unmerged until implementing corresponding logic for MySQL.

decryptor/mysql/types/long.go
```diff
@@ -0,0 +1,98 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+// LongDataTypeEncoder is encoder of TypeLong in MySQL
+type LongDataTypeEncoder struct{}
+
+// Encode implementation of Encode method of DataTypeEncoder interface for TypeLong
+func (t *LongDataTypeEncoder) Encode(ctx context.Context, data []byte, format type_awareness.DataTypeFormat) (context.Context, []byte, error) {
+	strValue := utils.BytesToString(data)
+	intValue, err := strconv.ParseInt(strValue, 10, 64)
```
Lagovas, line 23: as [I found](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_binary_resultset.html#sect_protocol_binary_resultset_row), Long type it is 4 byte integer. we can use here 32 bit size instead of 64

Zhaars: Agree!

PR #602, opened 2022-11-21
Fix search with aliased table

Zhaars:
Bug fixing related to mapping aliased table with config during searchable encryption table filtering.

In the queries like `select ... from some_table as alias where searchable_col = value` if the `some_table` is in encryptor config and searchable_col is not aliased column, Acra panic as default table remains nil and uninitialized. So that the main filtering logic was rewritten and is now based on `findColumnInfo` method that find table name depending on aliased or non-aliased tables in FROM statement and data from encryptor config.

## Checklist

  with new changes

### DIFF
--- a/encryptor/searchable_query_filter.go
+++ b/encryptor/searchable_query_filter.go
@@ -63,65 +63,24 @@ func (filter *SearchableQueryFilter) FilterSearchableComparisons(statement sqlpa
 		return nil
 	}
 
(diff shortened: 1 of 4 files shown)


vixentael (commented):
is it faster or slower than existing solution?

Zhaars:
> is it faster or slower than existing solution?

Conceptually, it is the same solution, moreover deleted redundant stuff.

Lagovas (approved):

PR #603, opened 2022-11-22
Extend limit&offset statements

Lagovas:
Extended parser with postgresql specific statements syntax and refactored rendering from the AST to the string in same format as it was parsed. 
But wasn't added support of cursor specific syntax for PostgreSQL with FETCH + limit&offset. It is much more complicated and should be implemented together with cursor's support.
PostgreSQL's format - https://www.postgresql.org/docs/current/sql-select.html#SQL-LIMIT
MySQL's format - https://dev.mysql.com/doc/refman/8.0/en/select.html

## Checklist

  with new changes


sqlparser/parse_test.go
```diff
@@ -1345,6 +1345,21 @@ var (
...
+		output: "select * from dual limit 10 offset 10",
+	}, { // PostgreSQL & MySQL limit&offset format
+		input:  "select * from dual limit 10",
+		output: "select * from dual limit 10",
+	}, { // MySQL format
+		input:  "select * from dual limit 10, 10",
+		output: "select * from dual limit 10, 10",
+	}, { // PostgreSQL format
+		input:  "select * from dual limit all",
+		output: "select * from dual limit all",
```
Zhaars, line 1359: As I understand, currently, all of these syntax constructions are supported by both MySQLDialect and PostgreSQLDialect. Does it make sense to throw some error in case of a specific DB format?

Lagovas: agree, added errors and tests for these cases. additionally I found how we can configure verbosity of loggers in parser and tokenizer. I just decreased verbosity of tokenizer for acra in non-debug mode to avoid logging sensitive parts of SQL query which we cannot control. But now I keep in mind this options that may help in debugging and extending parser.

Zhaars (approved):
Lgtm

PR #606, opened 2022-12-01
Add tests for searchable encryption with NotEqual operator

Lagovas:
Found that previously it wasn't supported properly and we haven't tests. After #605 were added support and here added tests.

## Checklist

  with new changes


tests/test.py
```diff
@@ -6823,24 +6823,129 @@ def executeBulkInsert(self, query, values):
 
 
 class TestSearchableTransparentEncryption(BaseSearchableTransparentEncryption):
```
Zhaars, line 6825: Do we want to add a couple of tests for not-equal operator for `SearchableTokenization`?

Lagovas: makes sense. added

Zhaars (approved):

PR #611, opened 2022-12-15
Fix MySQL reset packet with prepared statements

Zhaars:
Fixed bug with MySQL PreparedStatements processing, add `Reset` packet handler. Add check on params number in `Execute` Statement.

PR is draft, as `cossacklabs/ci-py-go-themis` contains new mysql-pyhton-connector with CMySQLConnector. So we need to test it on the CI

## Checklist

  with new changes

### DIFF
--- a/decryptor/mysql/response_proxy.go
+++ b/decryptor/mysql/response_proxy.go
@@ -415,7 +415,7 @@ func (handler *Handler) handleStatementExecute(ctx context.Context, packet *Pack
 		parameters, err := packet.GetBindParameters(paramsNum)
 		if err != nil {
 			log.WithError(err).Error("Can't parse OnBind parameters")
-			return nil
+			return err
 		}
 
 		newParameters, changed, err := handler.queryObserverManager.OnBind(ctx, statement.Query(), parameters)


tests/test.py
```diff
@@ -905,6 +905,76 @@ def execute_prepared_statement(self, query, args=None):
...
+            row_data = {column_name: value
+                        for column_name, value in zip(columns_name, row)}
+            result.append(row_data)
+        return result
+
+    def execute(self, query, args=None):
+        if args is None:
+            args = []
+        with contextlib.closing(mysql.connector.connect(
+                use_unicode=False, raw=self.connection_args.raw, charset='ascii',
```
Lagovas, line 923: why we don't use unicode and use charset ascii instead of utf8?

tests/test.py
```diff
@@ -4689,6 +4759,24 @@ def executePreparedStatement(self, query, args=None):
         ).execute_prepared_statement(query, args=args)
 
 
+class TestMysqlConnectorCBinaryPreparedStatement(BasePrepareStatementMixin, BaseTestCase):
```
Lagovas, line 4762: can you add some comment with description what we expected this connector will do? it will help to understand at the future if this test will fail after updating driver or something else. and explain what we expected and should reproduce. because after a year if this test will fail, we will not remember what behavior we expected and what to do to back it again

Lagovas (approved):

PR #616, opened 2022-12-22
Split integration tests

Zhaars:
The idea of that PR is to split `test.py` file into pieces as now it's difficult to work with that file. IDEA needs more time to index the file, plus its more difficult to find the class/mixing you need.

What have been made in this PR:

- move all encryptor config files from `tests/*` to `tests/encryptor_configs` folder.
- split test.py into several subfiles with tests:
  - `base.py` contains all general classes/constants/mixings etc 
  - `searchable_transparent_encryption.py` contains base mixing for searchable/transparent encryption tests and their related tests;
  - `tokenization.py` - contains base mixing for tokenization/masking tests and their related tests;
  - `type_aware.py` - contains all tests related to TA
   - `integrations.py` - contains all tests with integrations (AWS, Vault, Consul, etc)
  
 `test.py` has less than 5k of codes now and contains general tests unrelated to the previously described. The idea was to save `test.py` file as the main file, not to change the CI running scripts.
 
 Currently, PR is in a draft state, as some test files can be added and structure can be changed.

## Checklist

  with new changes

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -35,6 +35,8 @@ from pythemis import smessage, scell
 from sqlalchemy.dialects import mysql as mysql_dialect
 from sqlalchemy.dialects import postgresql as postgresql_dialect
 
+import generate_random_data
+
 # add to path our wrapper until not published to PYPI
 sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(__file__)), 'wrappers/python'))
 
@@ -47,16 +49,16 @@ def load_random_data_config():
 
 
 def get_random_data_files():
-    folder = os.environ.get(TEMP_DATA_FOLDER_VARNAME)
+    folder = os.environ.get(generate_random_data.TEMP_DATA_FOLDER_VARNAME)
     if not folder:
         folder = tempfile.mkdtemp(None, 'test_data', None)
         # set temp folder before call generator
-        os.environ.setdefault(TEMP_DATA_FOLDER_VARNAME, folder)
+        os.environ.setdefault(generate_random_data.TEMP_DATA_FOLDER_VARNAME, folder)
         # remember that we generated from script to cleanup at end
         os.environ.setdefault(TEMP_DATA_GENERATED, folder)
         print("You didn't set {} env var. Test data will be generated to <{}> "
               "folder and removed at end".format(
-            TEMP_DATA_FOLDER_VARNAME, folder))
+            generate_random_data.TEMP_DATA_FOLDER_VARNAME, folder))
     if not os.path.exists(folder) or len(os.listdir(folder)) == 0:
         command = [sys.executable, abs_path('tests/generate_random_data.py')]
         print('call {}'.format(' '.join(command)))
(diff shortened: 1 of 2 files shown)


tests/base.py
```diff
@@ -100,6 +100,8 @@
 TEST_WITH_REDIS = os.environ.get('TEST_REDIS', 'off').lower() == 'on'
 TEST_TRACE_TO_JAEGER = os.environ.get('TEST_TRACE_JAEGER', 'off').lower() == 'on'
 
+SET_UP_INTERNAL_TEST = utils.get_bool_env('SET_UP_INTERNAL_TEST', default=False)
```
Lagovas, line 103: how and for what it used?

tests/test.py
```diff
@@ -4469,9 +4470,14 @@ def check_all_decryptions(self, **context):
...
+    try:
+        output_path = os.environ.get('TEST_XMLOUTPUT', '')
+        if output_path:
+            with open(output_path, 'wb') as output:
+                unittest.main(testRunner=xmlrunner.XMLTestRunner(output=output))
+        else:
+            unittest.main()
+    finally:
+        baseTearDownModule()
+        sys.exit(0)
```
Lagovas, line 4483: i'm not sure that we should return exit status 0 in the `finally` statement because `finally` executed even if raised some error and we will not know about it and could not handle it in the caller script.

tests/base.py
```diff
@@ -0,0 +1,2602 @@
...
+    collectTests(standard_tests, tests)
+    for tcase in tests:
+        if not tcase:
+            continue
+        tcase.__class__.__module__ = 'base'
+    return unittest.TestSuite(tests)
+
+
+def setUpModule():
+    baseSetUpModule()
```
Lagovas, line 785: do we need separate function with different name or we can leave only `setUpModule`?

and what about to move these functions to the bottom of the file to simplify finding callback? and lets add comment with reference to the documentation that we patch module names due to unittest runs test cases one by one and calls setUpModule once only if test cases ordered by module name or from one module, otherwise if tests unordered with same modules, setUpModule will be called same times. So we specify one name for all of them because need only one call and only one such method we use. Otherwise, in the future, we should order by `.__class__.__module__` testcases to allow call setUpModule per modules.

tests/base.py
```diff
@@ -0,0 +1,2602 @@
...
+                return data
+
+    def execute(self, query, args=None):
+        return self._execute(query, args, prepare=False)
+
+    def execute_prepared_statement(self, query, args=None):
+        return self._execute(query, args, prepare=True)
+
+
+class ExecutorMixin:
```
Lagovas, line 1193: what do you think about leaving here all base staff with global variable declarations and commonly used functions and all mixins with tests move to separate module like `test_common.py`? To not have here any runnable test cases, only declarations and common function? On the first view it looks like will simplify reading by separation runnable testcases and neighbour mixins and other not-runnable stuff. Plus, unittest has convention that all runnable modules should be named `test*`, so we will fit in this convention too, if base.py will not have it.

plus, we can some of common functions move to `utils.py` which not depends on global variables.

Lagovas (approved):
LGTM
I restarted failed test. Looks like some flaky situation with long start of background service. Please, merge after passing all tests successfully.

PR #621, opened 2023-01-16
Fix failing test with searchable tokenization

Zhaars:
Fix the failing test with searchable tokenization. Updated pgsql Coder with additional logic on `Encode` method to check if the data need to be encoded to hex or returned as is.

## Checklist

  with new changes


encryptor/dbDataCoder.go
```diff
@@ -198,8 +198,13 @@ func (*PostgresqlDBDataCoder) Encode(expr sqlparser.Expr, data []byte) ([]byte,
 			val.Type = sqlparser.PgEscapeString
 			fallthrough
 		case sqlparser.StrVal:
-			// if it is valid string, return as is. otherwise there is binary data, and we should encode it
-			if utils.IsPrintablePostgresqlString(data) {
+			binValue, err := utils.DecodeEscaped(data)
+			// if the data changes previously or we have decode error means that we binary data should be returned as hex
```
Lagovas, line 202: a bit complicated description, imho. let's clarify a bit more why we do this: "Check is encryption/tokenization operation returned string in "binary string" format instead of UTF8 string. If we can decode it as binary value and get another value, then escape it as hex binary value.".

Lagovas (approved):

PR #632, opened 2023-02-07
Actualize Prometheus metrics collecting

Zhaars:
Actualizing Prometheus metrics collection. 

Added new metrics:
- **acra_decryptions_total** labels {"**status**": [ "`success`", "`fail`"], "type": [ "`acrastruct`", "`acrablock`", "`acrablock_searchable`", "`acrastruct_searchable`"],}
- **acra_encryptions_total**  labels {"**status**": [ "`success`", "`fail`"], "type": [ "`acrastruct`", "`acrablock`", "`acrablock_searchable`", "`acrastruct_searchable`"],}
- **acra_tokenizations_total**  labels {"**status**": [ "`success`", "`fail`"], "token_type": "`{token_type}`"}
- **acra_detokenizations_total** abels {"**status**": [ "`success`", "`fail`"], "token_type": "`{token_type}`"}

Previously used metrics are left as they are to save backward compatibility and marked as deprecated. 

## Checklist

- [] Public API has proper documentation in the [Acra documentation] site or has PR on [documentation repository] 
  with new changes

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -2122,16 +2122,15 @@ class TestPrometheusMetrics(AcraTranslatorMixin, BaseTestCase):
         self.searchableTransparentTest.encryptor_table = BaseSearchableTransparentEncryption().get_encryptor_table()
         base.metadata.create_all(self.engine_raw, [self.searchableTransparentTest.encryptor_table])
 
-        # init searchable transparent encryption test
+        # init tokenization test
         self.tokenizationTest = TestTokenization()
         self.tokenizationTest.engine_raw = self.engine_raw
         self.tokenizationTest.engine1 = self.engine1
         self.tokenizationTest.engine2 = self.engine2
 
     def tearDown(self):
-        self.searchableTransparentTest.tearDown()
         base.metadata.drop_all(self.engine_raw, [self.searchableTransparentTest.encryptor_table])
-        self.tokenizationTest.tearDown()
+        super().tearDown()
 
     def fork_acra(self, popen_kwargs: dict = None, **acra_kwargs: dict):
         acra_kwargs.update(encryptor_config_file=self.ENCRYPTOR_CONFIG)


cmd/acra-translator/common/service.go
```diff
@@ -81,7 +83,7 @@ func (service *TranslatorService) Decrypt(ctx context.Context, acraStruct, clien
 
 	data, decryptErr := service.handler.DecryptWithHandler(handler, acraStruct, dataContext)
 	if decryptErr != nil {
-		base.AcrastructDecryptionCounter.WithLabelValues(base.DecryptionTypeFail).Inc()
+		base.AcrastructDecryptionCounter.WithLabelValues(base.LabelStatusFail).Inc()
```
Lagovas, line 86: can you add for every legacy metric changes TODO with a reminder to remove it after 1-2 version due to migrated metrics to new? and will be great to create task for that for the future )

cmd/acra-translator/grpc_api/service.go
```diff
@@ -71,7 +72,6 @@ func (service *TranslatorService) Encrypt(ctx context.Context, request *EncryptR
 
 	response, err := service.service.Encrypt(ctx, request.Data, request.ClientId, nil)
 	if err != nil {
-		base.APIEncryptionCounter.WithLabelValues(base.EncryptionTypeFail).Inc()
```
Lagovas, line 74: this is legacy metric that may be used now. lets leave with todo to remove it after deprecation?

crypto/acrastruct.go
```diff
@@ -50,6 +52,7 @@ func (handler AcraStructHandler) Decrypt(data []byte, context *base.DataProcesso
 	privateKeys, err := context.Keystore.GetServerDecryptionPrivateKeys(accessContext.GetClientID())
 	defer utils.ZeroizePrivateKeys(privateKeys)
 	if err != nil {
+		base.AcrastructDecryptionCounter.WithLabelValues(base.LabelStatusFail).Inc()
```
Lagovas, line 55: hm, as I understand, it will increment twice for the acra-translator for its `decrypt` methods? If AT uses cryptohandler that uses this handler under the hood, then it will increment twice

tests/test.py
```diff
@@ -2176,6 +2176,7 @@ def testAcraServer(self):
             'acraserver_request_processing_seconds_bucket': {'min_value': 0},
 
             'acra_acrastruct_decryptions_total': {'min_value': 1},
+            'acra_decryptions_total': {'min_value': 1},
```
Lagovas, line 2179: will be great to add tests where all these metrics are used and we can validate them.  Make a request to the table that is configured for all types of tokenization, for all types of encryption and decryption and validate, that all appropriate metrics were incremented. And same for acra-translator and its API. As I see, previously it just called `HexFormatTest` test case which forced simple acrastruct decryption.  Now we have much more features) But to simplify tests, we can re-use some existing tests, imho.

Lagovas (changes requested):

tests/test.py
```diff
@@ -2108,6 +2109,33 @@ class TestPrometheusMetrics(AcraTranslatorMixin, BaseTestCase):
...
+        base.metadata.create_all(self.engine_raw, [self.searchableTransparentTest.encryptor_table])
+
+        # init searchable transparent encryption test
+        self.tokenizationTest = TestTokenization()
+        self.tokenizationTest.engine_raw = self.engine_raw
+        self.tokenizationTest.engine1 = self.engine1
+        self.tokenizationTest.engine2 = self.engine2
+
+    def tearDown(self):
+        self.searchableTransparentTest.tearDown()
```
Lagovas, line 2132: why we call it if we didn't call setUp?

pseudonymization/tokenizeQuery.go
```diff
@@ -69,6 +70,10 @@ func (encryptor *TokenizeQuery) OnQuery(ctx context.Context, query base.OnQueryO
 	clientSession := base.ClientSessionFromContext(ctx)
 	bindSettings := queryEncryptor.PlaceholderSettingsFromClientSession(clientSession)
 	for _, item := range items {
+		if !item.Setting.IsTokenized() {
```
Lagovas, line 73: Should we add the same for searchable encryption?

Lagovas (approved):
great. remains only to update documentation about new added metrics

PR #636, opened 2023-02-14
Extend acra-keys list with rotated keys

Zhaars:
Extend `acra-keys` `list` with displaying historical keys for V1/V2. Introduce the new flag `--historical-keys` for the list subcommand.

## Checklist

- [X] Change is covered by automated tests
- [x] The [coding guidelines] are followed
- [ ] Public API has proper documentation in the [Acra documentation] site or has PR on [documentation repository] 
  with new changes
- [ ] CHANGELOG.md is updated (in case of notable or breaking changes)
- [x] CHANGELOG_DEV.md is updated
- [ ] Benchmark results are attached (if applicable)
- [ ] [Example projects and code samples] are up-to-date (in case of API changes) 

[coding guidelines]: https://golang.org/doc/effective_go
[Example projects and code samples]: https://github.com/cossacklabs/acra-engineering-demo
[Acra documentation]: https://docs.cossacklabs.com/
[documentation repository]: https://github.com/cossacklabs/product-docs

PR #639, opened 2023-02-16
Extend acra keys list 

Zhaars:
This PR is mostly an addition to https://github.com/cossacklabs/acra/pull/636

- Added `Index` option for KeyDescription - where 1 is always the id of the current key, rotated keys have increased order idx;
- Added `State` field (current/rotated) to display KEY state in JSON;
- Update printing keys in JSON format - combine all KeyDescription(rotated/current) in one output;
- Change ClientID from byte slice to string - as standard JSON marshaller prints pure bytes in base64;

## Checklist

  with new changes

### DIFF
--- a/cmd/acra-keys/keys/list-keys_test.go
+++ b/cmd/acra-keys/keys/list-keys_test.go
@@ -21,6 +21,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"flag"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -49,9 +50,9 @@ func TestPrintKeysDefault(t *testing.T) {
 	}
 
 	actual := output.String()
-	expected := `Idx | Key purpose | Client | Key ID
+	expected := `Index | Key purpose | Client | Key ID
 ------------+--------+--------------
-0   | testing     |        | Another KeyID
+0     | testing     |        | Another KeyID
 `
 	if actual != expected {
 		t.Errorf("Incorrect output.\nActual:\n%s\nExpected:\n%s", actual, expected)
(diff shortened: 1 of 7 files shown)


cmd/acra-keys/keys/list-keys_test.go
```diff
@@ -88,13 +89,13 @@ testing     |        | 2023-02-14 23:40:28 +0000 UTC | Another ID
...
 		{
-			ID:      "Another ID",
+			KeyID:   "Another KeyID",
 			Purpose: "testing",
 		},
 	}
 
 	output := bytes.Buffer{}
-	err := PrintKeys(keys, &output, &CommonKeyListingParameters{useJSON: true})
+	err := printKeysJSON(keys, &output)
```
Lagovas, line 98: we should test json generation with rotated keys too to capture errors when they merged incorrectly

keystore/keystore.go
```diff
@@ -439,17 +439,29 @@ type ServerKeyStore interface {
 	Reset()
 }
 
+// KeyMarker set key market for KeyDescription
+type KeyMarker string
```
Lagovas, line 443: lets name it as "state", which may be current or rotated. Because "marker" is not obvious meaning in that context

keystore/keystore.go
```diff
@@ -439,17 +439,29 @@ type ServerKeyStore interface {
...
 // "Purpose" is short human-readable description of the key purpose.
 // "ClientID" and "AdditionalContext" are filled in where relevant.
+// "CreationTime" used to display creation time of rotated key
 type KeyDescription struct {
-	ID       string
-	Purpose  KeyPurpose
-	ClientID []byte `json:",omitempty"`
-	// used to display creation time of rotated key
-	CreationTime time.Time
+	Idx          int
```
Lagovas, line 459: what about to name in "Index" instead of "Idx"? Because all other fields have fullname and this looks like abbreviation for something

keystore/utils.go
```diff
@@ -5,32 +5,35 @@ import (
 	"io"
 )
 
+const (
+	purposeHeader      = "Key purpose"
+	extraIDHeader      = "Client"
+	keyIDHeader        = "Key ID"
+	creationTimeHeader = "Creation Time"
+	idxHeader          = "Idx"
```
Lagovas, line 13: and here too

Lagovas (approved):

PR #641, opened 2023-02-21
Extend acra-keys destroy with specific rotated key

Zhaars:
Extended `acra-keys` `destroy` with destroying the specific rotated keys. 

* Introduce a new flag `--index` representing the key to destroy.
* Introduce new interface for destruction of rotated key `StorageRotatedKeyDestruction`

## Checklist

  with new changes

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -903,6 +903,138 @@ class TestKeyStoreMigration(BaseTestCase):
             self.assertEquals(selected['data'], data.encode('ascii'))
 
 
+class TestAcraKeysWithRotatedKeys(unittest.TestCase):
+    def setUp(self):
+        self.master_key = get_master_key()
+        self.dir_with_distinguished_name_client_id = tempfile.TemporaryDirectory()
+        self.time_to_rotate = 3
+        self.create_key_store_with_client_id_from_cert(TLS_CLIENT_ID_SOURCE_DN,
+                                                       self.dir_with_distinguished_name_client_id.name)
+
+    def test_keys_rotation(self):
+        extract_resp = self.extrac_client_id(TLS_CLIENT_ID_SOURCE_DN)
+
+        if KEYSTORE_VERSION == 'v1':
+            key_ids = [
+                'poison_key',
+                'poison_key.pub',
+                'poison_key_sym',
+                'secure_log_key',
+                '{}_storage_sym'.format(extract_resp['client_id']),
+                '{}_storage.pub'.format(extract_resp['client_id']),
+                '{}_storage'.format(extract_resp['client_id']),
+                '{}_hmac'.format(extract_resp['client_id']),
+            ]
+        else:
+            key_ids = [
+                'audit-log',
+                'poison-record-sym',
+                'poison-record',
+                'client/{}/storage-sym'.format(extract_resp['client_id']),
+                'client/{}/hmac-sym'.format(extract_resp['client_id']),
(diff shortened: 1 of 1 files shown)


keystore/filesystem/server_keystore.go
```diff
@@ -1279,6 +1282,67 @@ func (store *KeyStore) DestroyHmacSecretKey(clientID []byte) error {
...
+		return err
+	}
+
+	// 1 is always index of current key of the keystore
+	// all rotated keys have index after 1
+	if len(rotatedKeyFiles) == 0 || index > len(rotatedKeyFiles)+1 {
+		return ErrInvalidIndex
+	}
+
+	rotatedKey := rotatedKeyFiles[index-1]
```
Lagovas, line 1337: do we sure that ReadDir returns files in the expected order? can it be changed in other filesystems?
or will be better to sort on keystore by name (that contains timestamp) side to be sure that it works similar in all environments?

keystore/v2/keystore/hmac.go
```diff
@@ -84,6 +84,25 @@ func (s *ServerKeyStore) DestroyHmacSecretKey(clientID []byte) error {
...
+// DestroyRotatedHmacSecretKey destroy created rotated hmac symmetric key
+func (s *ServerKeyStore) DestroyRotatedHmacSecretKey(clientID []byte, index int) error {
+	log := s.log.WithField("clientID", clientID)
+	ring, err := s.OpenKeyRingRW(s.clientHMACKeyPath(clientID))
+	if err != nil {
+		log.WithError(err).Debug("Failed to open HMAC key ring for client")
+		return err
+	}
+	// Index represent virtual index of key
+	// 1 is always index of current key of the keystore
```
Lagovas, line 96: this comment confuse a bit because we decrement by one below... IMHO, we should describe here, that keyrings have internal index starting from 0 and we get index here that represents virtual index starting from 1.

p.s. does it work correctly with destroyed keys? What key will be destroyed, if we have `current` + `destroyed` + `rotated` + `rotated` and we call `acra-keys destroy --index 2`? will it destroy first rotated or destroy second time already destroyed? can we test that case?

keystore/v2/keystore/keyStore.go
```diff
@@ -291,17 +298,57 @@ func (s *ServerKeyStore) listRotatedRings(path string, purpose keystore.KeyPurpo
...
+		}
+		rotatedActiveKeys = append(rotatedActiveKeys, i)
+	}
+
+	if index-1 > len(rotatedActiveKeys) {
+		log.WithField("index", index).Debug("no key matched to index")
+		return ErrInvalidIndex
+	}
+	// 1 is always index of current key of the keystore, so we need to subtract 1 from search index
+	// as slice element enumeration starts from 0 subtract 1 again
```
Lagovas, line 344: lets add more explanation of why we decrease 2: 
> keyring internally stores keys from older to newer and the newest key has index `len(keys) -1`. So we decrease once to start from the `len(keys)-1` position. Incoming parameter `index` represents the virtual index starting from 1 where 1 is the newest key. But slices work with 0-indexation, so we decrease one more time.

Lagovas (approved):
great job

Lagovas (changes requested):
I forgot about integration tests... can you add integration tests to be sure that whole flow works from executing binary?

Lagovas (approved):

PR #644, opened 2023-03-02
Implement PostgreSQL prepared statement from SQL tracking

Zhaars:
Implementation of the Draft version of supporting PreparedStatements from SQL.

**What was done in this PR**:
- introduce new `Observer` for tracking PreparedStatements from SQL (`prepared_statements_sql_observer.go`);
- extend/fix SQL parser  to support proper parsing and formatting queries like `prepare {prepare_statement_name} (params...) as the query` and `execute  (values...)  {prepare_statement_name} ` for PostgreSQL syntax;
- added integration tests to cover basic functionality;

Also, during the discussion, we figured out that we can't support SQL PreparedStatement in transactions.  As we need to have additional logic related to tracking the state of PreparedStatement. 

------------------------------------------------------------------------------------------------------------------------------------------------------------------
For example:
1. Transaction starts (`Begin`);
2. Prepare PreparedStatement  (`prepare insert_data (int) as insert into test_table (id)`) - during that step we store PreparedStatement to the session registry;
3. Execute Statement by name  (`execute insert_data (1)`) 
4. Rollaback transaction - (`rollback`) - In such case we need to delete the prepared statement from the registry, to achieve that we need to track the state of the PreparedStatement. 
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
Decided to leave it as is for now and marked such a situation with a comment as there are not too many requests from clients with such problems.

**What should be done more:**
- Implement `deallocate` queries;
- Added more tests for different scenarios;
- Probably extend the parser to support more value types; 

## Checklist

  with new changes

### DIFF
--- a/decryptor/postgresql/prepared_statements_sql_observer.go
+++ b/decryptor/postgresql/prepared_statements_sql_observer.go
@@ -65,7 +65,7 @@ func (encryptor *PreparedStatementsQuery) onPrepare(ctx context.Context, prepare
 	var registry = encryptor.session.PreparedStatementRegistry()
 
 	// PostgreSQL allows create statement only once during the session
-	if _, err := registry.StatementByName(prepareQuery.PreparedStatementName.String()); err == nil {
+	if _, err := registry.StatementByName(preparedStatementName); err == nil {
 		logrus.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorGeneral).
 			WithError(err).Errorln("PreparedStatement already stored in registry")
 		return nil, false, ErrStatementAlreadyInRegistry
--- a/tests/test.py
+++ b/tests/test.py
@@ -4710,7 +4710,11 @@ class TestSQLPreparedStatements(AcraCatchLogsMixin):
 
     def tearDown(self):
         self.engine_raw.execute(self.test_prepared_sql_statements_table.delete())
+        base.metadata.remove(self.test_prepared_sql_statements_table)
+
         self.engine_raw.execute(self.default_client_id_table.delete())
+        base.metadata.remove(self.default_client_id_table)
+
         super(TestSQLPreparedStatements, self).tearDown()
 
 


sqlparser/ast_methods.go
```diff
@@ -871,10 +880,14 @@ func (node *Execute) walkSubtree(visit Visit) error {
 
 // Format formats the node.
 func (node *Prepare) Format(buf *TrackedBuffer) {
-	if node.ColumnTypes != nil {
-		buf.Myprintf("prepare %v (%v) from (as) %v", node.PreparedStatementName, node.ColumnTypes, node.PreparedStatementQuery)
+	switch buf.dialect.(type) {
+	case *mysql.MySQLDialect:
+		if node.ColumnTypes != nil {
+			buf.Myprintf("prepare %v (%v) from (as) %v", node.PreparedStatementName, node.ColumnTypes, node.PreparedStatementQuery)
```
Lagovas, line 886: does mysql support any columns for prepare?
https://dev.mysql.com/doc/refman/8.0/en/prepare.html - looks like it expects only statement name and then query after `from`

Lagovas: plus as I see, we don't render anything in case of `if node.ColumnTypes == nil` and haven't even fallthrough to execute pgsql's logic

sqlparser/sql.y
```diff
@@ -861,6 +869,10 @@ char_type:
   {
     $$ = ColumnType{Type: string($1), Length: $2}
   }
+| BYTEA length_opt
+  {
+    $$ = ColumnType{Type: string($1), Length: $2}
```
Lagovas, line 874: does postgresql accepts length for bytea? should it have `length_opt`? 
https://www.postgresql.org/docs/current/datatype-binary.html

tests/test_tokenization.py
```diff
@@ -133,6 +133,60 @@ def fetch_from_2(self, query, parameters={}, literal_binds=True):
...
+            for x in data.values():
+                if isinstance(x, int):
+                    values_str += '{}, '.format(str(x))
+                if x is None:
+                    values_str += 'null, '
+                if isinstance(x, str):
+                    values_str += '\'{}\', '.format(x)
+                if isinstance(x, bytes):
+                    if len(x) == 0:
+                        values_str += '\' \', '
```
Lagovas, line 150: as I see, we have space here?

tests/test_tokenization.py
```diff
@@ -133,6 +133,60 @@ def fetch_from_2(self, query, parameters={}, literal_binds=True):
         query, parameters = self.compileQuery(query, parameters=parameters, literal_binds=literal_binds)
         return self.executor2.execute_prepared_statement(query, parameters)
 
+    def compile_execute_prepare(self, prepared_name, data={}):
+        if len(data) == 0:
+            prepare_query_sql = "execute {}".format(prepared_name)
+        else:
+            values_str = ''
+            for x in data.values():
+                if isinstance(x, int):
```
Lagovas, line 142: did you try [sa.literal](https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.literal) to delegate escaping and conversion to sqlalchemy?

Lagovas: sqlalchemy can compile queries to the string, and name it like `bind values`. plus, such objects convertible to the string via `str(obj)`. And when they have assigned engine or dialect, some of them serialize to the target syntax

decryptor/postgresql/prepared_statements_sql_observer.go
```diff
@@ -0,0 +1,164 @@
...
+
+// ID returns name of this QueryObserver.
+func (encryptor *PreparedStatementsQuery) ID() string {
+	return "PreparedStatementsQuery"
+}
+
+// OnQuery processes query text before database sees it.
+func (encryptor *PreparedStatementsQuery) OnQuery(ctx context.Context, query base.OnQueryObject) (base.OnQueryObject, bool, error) {
+	logrus.Debugln("PreparedStatementsQuery.OnQuery")
+	parsedQuery, err := encryptor.parser.Parse(query.Query())
```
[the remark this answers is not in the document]
Lagovas: Statement() returns the same but already parsed in some cases, why not use that?

decryptor/postgresql/prepared_statements_sql_observer.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, false, ErrStatementNotPresentInRegistry
+	}
+
+	// collect all BoundValues
+	values := make([]base.BoundValue, 0, len(executeQuery.Values))
+	for _, value := range executeQuery.Values {
+		switch val := value.(type) {
+		case *sqlparser.SQLVal:
+			values = append(values, NewPgBoundValue(val.Val, bindFormatText))
+		case *sqlparser.NullVal:
```
Lagovas, line 128: should we handle `default` case with other types and append `nil` values to be able to map them to appropriate placeholders in the query?

encryptor/utils.go
```diff
@@ -515,6 +516,72 @@ func ParsePlaceholderIndex(placeholder *sqlparser.SQLVal) (int, error) {
...
+			if lColumnSetting == nil {
+				return true, nil
+			}
+
+			if !lColumnSetting.IsSearchable() && !lColumnSetting.IsConsistentTokenization() {
+				return true, nil
+			}
+
+			if sqlVal, ok := comparisonExpr.Right.(*sqlparser.SQLVal); ok && isSupportedSQLVal(sqlVal) {
+				if comparisonExpr.Operator == sqlparser.EqualStr || comparisonExpr.Operator == sqlparser.NotEqualStr || comparisonExpr.Operator == sqlparser.NullSafeEqualStr {
```
Lagovas, line 566: LIKE and ILIKE too?

decryptor/postgresql/prepared_statements_sql_observer.go
Lagovas: below I see that we collects placeholders by their indexes and only interesting for us. but we should be able to fill array of values with not interesting to keep correct indexing.

hmac/decryptor/hashQuery.go
```diff
@@ -198,20 +203,24 @@ func (encryptor *HashQuery) OnBind(ctx context.Context, statement sqlparser.Stat
 			indexes = append(indexes, index)
 		}
 	}
+
+	bindData := queryEncryptor.ParseSearchQueryPlaceholdersSettings(statement, encryptor.schemaStore)
+	if len(bindData) != len(indexes) {
```
Lagovas, line 208: above we collect only searchable items. but as I understand, `ParseSearchQueryPlaceholdersSettings` will return tokenized items too and they will break the expected amount of items

we should add tests with several security controls, with searchable, tokenized and simple data bound via `EXECUTE` SQL query

pseudonymization/tokenizeQuery.go
```diff
@@ -139,20 +146,23 @@ func (encryptor *TokenizeQuery) OnBind(ctx context.Context, statement sqlparser.
 			indexes = append(indexes, index)
 		}
 	}
+
+	bindData := queryEncryptor.ParseSearchQueryPlaceholdersSettings(statement, encryptor.schemaStore)
+	if len(bindData) != len(indexes) {
```
Lagovas, line 151: same here. probably we should do `>` or `<=` comparison

sqlparser/parse_test.go
```diff
@@ -1402,6 +1402,15 @@ var (
...
 			input:   "update test_table set price = price * 1.10 from table2 as t2 where price <= 99.99 returning 1, 0 as literal, t2.zone_id, specified_client_id, other_column, default_client_id, null",
 			dialect: postgresql.NewPostgreSQLDialect(),
+		}, {
+			input:   "prepare fooplan (int, text, bool, bytea) as insert into foo values ($1, $2, $3, $4)",
+			dialect: postgresql.NewPostgreSQLDialect(),
+		}, {
+			input:   "prepare fooplan (int, some_unknown_type) as select * from table_1 where field = $1",
+			dialect: postgresql.NewPostgreSQLDialect(),
+		}, {
+			input:   "execute fooplan ('test', 11, true)",
```
Lagovas, line 1412: could the statement name be wrapped with the quotes? if yes, we should test it too

sqlparser/parse_test.go
```diff
@@ -1402,6 +1402,15 @@ var (
 		}, {
 			input:   "update test_table set price = price * 1.10 from table2 as t2 where price <= 99.99 returning 1, 0 as literal, t2.zone_id, specified_client_id, other_column, default_client_id, null",
 			dialect: postgresql.NewPostgreSQLDialect(),
+		}, {
+			input:   "prepare fooplan (int, text, bool, bytea) as insert into foo values ($1, $2, $3, $4)",
+			dialect: postgresql.NewPostgreSQLDialect(),
+		}, {
+			input:   "prepare fooplan (int, some_unknown_type) as select * from table_1 where field = $1",
```
Lagovas, line 1409: could the type names be wrapped with the quotes?

tests/encryptor_configs/ee_prepared_statements_sql.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - empty
+      - token_i32
+      - token_i64
+      - token_str
+      - token_bytes
+      - token_email
+    encrypted:
+      - column: token_i32
+        token_type: int32
+        consistent_tokenization: true
```
Lagovas, line 15: we can skip this options due to its `true` by default

tests/test.py
```diff
@@ -4649,6 +4654,292 @@ def test_insert_returning(self):
...
+            sa.Column('nullable_column', sa.Text, nullable=True),
+            sa.Column('empty', sa.LargeBinary(length=base.COLUMN_DATA_SIZE), nullable=False, default=b''),
+            sa.Column('token_i32', sa.Integer()),
+            sa.Column('token_i64', sa.BigInteger()),
+            sa.Column('token_str', sa.Text),
+            sa.Column('token_bytes', sa.LargeBinary(length=base.COLUMN_DATA_SIZE), nullable=False, default=b''),
+            sa.Column('token_email', sa.Text),
+            extend_existing=True,
+        )
+        base.metadata.create_all(self.engine_raw, [default_client_id_table])
```
Lagovas, line 4732: if we use global metadata, we should clear tables and metadata at the end of test. we can use `contextlib.ExitStack` for that. Because such littering can complicate future refactorings or extensions. I met such problems with re-using metadata in EE tests

tests/test.py
```diff
@@ -4649,6 +4654,292 @@ def test_insert_returning(self):
...
+
+        # expect fail on the prepare query with the same name
+        try:
+            self.prepare_1(prepared_name='insert_data', query=default_client_id_table.insert(), data_types=data_types)
+        except Exception:
+            self.assertIn("PreparedStatement already stored in registry", self.read_log(self.acra))
+            pass
+
+        self.execute_prepared_1(prepared_name='insert_data', data=data)
+        # create SQL prepared statements in DB with name `insert_data`
```
Lagovas, line 4768: we can skip that comment or actualize statements name -> `select_all_data`

tests/test_tokenization.py
Lagovas: I guess you tried and it not works)

tests/test_tokenization.py
```diff
@@ -133,6 +133,64 @@ def fetch_from_2(self, query, parameters={}, literal_binds=True):
...
+    def compile_prepare(self, prepared_name, query, data_types={}, literal_binds=True):
+        query, _ = self.compileQuery(query, parameters=data_types, literal_binds=literal_binds)
+
+        if len(data_types) != 0:
+            prepare_query_sql = "prepare {} ({}) as {}".format(prepared_name, ", ".join(data_types.values()), query)
+        else:
+            prepare_query_sql = "prepare {} as {}".format(prepared_name, query)
+        return prepare_query_sql
+
+    def prepare_1(self, prepared_name, query, data_types={}, literal_binds=True):
```
Lagovas, line 166: hm... lets break previous convention and acccept engine as parameter to not produce a lot of functions `x_1` and `x_2`? it will help to re-use them for other flows too

Lagovas (approved):

PR #662, opened 2023-07-21
Fixed MariaDB protocol handling issue

Zhaars:
Fixed [issue](https://github.com/cossacklabs/acra/issues/648) related to MariaDB protocol. What has been done in this PR:
- fixed protocol issue related to tracking Column metadata in ColumnDefinition packet;
- tracking `MARIADB_CLIENT_EXTENDED_TYPE_INFO` capability from the client and server sides;
- extend `updateFieldEncodedType` func with updating specific type data as `Charset`, `ColumnLength`, etc;

## Checklist

- [ ] Change is covered by automated tests
- [ ] The [coding guidelines] are followed
- [ ] Public API has proper documentation in the [Acra documentation] site or has PR on [documentation repository] 
  with new changes
- [ ] CHANGELOG.md is updated (in case of notable or breaking changes)
- [ ] CHANGELOG_DEV.md is updated
- [ ] Benchmark results are attached (if applicable)
- [ ] [Example projects and code samples] are up-to-date (in case of API changes) 

[coding guidelines]: https://golang.org/doc/effective_go
[Example projects and code samples]: https://github.com/cossacklabs/acra-engineering-demo
[Acra documentation]: https://docs.cossacklabs.com/
[documentation repository]: https://github.com/cossacklabs/product-docs

PR #663, opened 2023-07-26
improve unicode support

Lagovas:
- updated `DecodeOctal` function to operate with runes instead of bytes. previously it returned error on any not printable symbol that applies to every unicode symbol
- updated PostgreSQL data coder/decoder. now it uses info from the ColumnEncryptionSetting and always encodes into hex/octal format if destination type is not specified as text. Acra expects bytea/blob type on database side and default dest type is binary. Postgresql's bytea type expects hex/octal string, it doesn't work with escaped strings `E''`
- updated integration tests to use unicode strings

## Checklist

  with new changes

### DIFF
--- a/CHANGELOG_DEV.md
+++ b/CHANGELOG_DEV.md
@@ -1,3 +1,6 @@
+# 0.96.0 - 2023-07-26
+- Improve unicode support
+
 # 0.96.0 - 2023-03-06
 - Added support of PreparedStatement from SQL for PostgreSQL;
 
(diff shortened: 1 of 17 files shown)


Lagovas:
Unfortunately, I removed using unicode strings from the integration tests due to MySQL driver is outdated and works incorrectly with utf8 charset and need to be updated. I plan to do it in the separate PR. I already tested the freshest mysql-connector-python 8.1.0 and it looks like fixed a lot of problems (encodings, hardcoded TLSv1, conversion to python types) but now it correctly handles binary types and charset sent from the server. So, we need to merge our fixes in #662 to not hack tests with decoding bytearray type to string)

So lets do it in separate PR after merging #662

PR #665, opened 2023-08-04
Add MySQL SQL PreparedStatements support

Zhaars:
## Checklist

  with new changes

### DIFF
--- a/decryptor/mysql/prepared_statement_sql_observer.go
+++ b/decryptor/mysql/prepared_statement_sql_observer.go
@@ -21,7 +21,7 @@ import (
 // ErrStatementNotPresentInRegistry represent an error that prepared statement already exist in session registry
 var ErrStatementNotPresentInRegistry = errors.New("prepared statement not present in registry")
 
-// PreparedStatementsQuery QueryDataEncryptor process PostgreSQL SQL PreparedStatement
+// PreparedStatementsQuery process MySQL SQL PreparedStatement
 type PreparedStatementsQuery struct {
 	proxyHandler        *Handler
 	parser              *sqlparser.Parser
@@ -31,7 +31,7 @@ type PreparedStatementsQuery struct {
 	querySelectSettings []*encryptor.QueryDataItem
 }
 
-// NewMySQLPreparedStatementsQuery create new QueryDataEncryptor to handle SQL PreparedStatement in the following format
+// NewMySQLPreparedStatementsQuery create new PreparedStatementsQuery to handle SQL PreparedStatement in the following format
 // https://dev.mysql.com/doc/refman/8.0/en/sql-prepared-statements.html
 func NewMySQLPreparedStatementsQuery(proxyHandler *Handler, parser *sqlparser.Parser, schemaStore config.TableSchemaStore) *PreparedStatementsQuery {
 	return &PreparedStatementsQuery{
@@ -181,6 +181,8 @@ func (e *PreparedStatementsQuery) onSet(ctx context.Context, setQuery *sqlparser
 
 		encryptedData, err := e.handleColumnFromSetArg(ctx, sqlVal, argName)
 		if err != nil {
+			logrus.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorGeneral).
+				WithError(err).Errorln("Failed to handle column from Set arg")
 			return nil, false, err
 		}
 
@@ -205,7 +207,8 @@ func (e *PreparedStatementsQuery) handleColumnFromSetArg(ctx context.Context, sq
 	var delim = e.schemaStore.GetDatabaseSettings().GetMySQLDatabaseSettings().GetPreparedStatementsSetArgDelimiter()
 
 	splits := strings.Split(argName, delim)
-	if len(splits) == 1 {
+	if len(splits) != 2 {
+		logrus.WithField("argument", argName).Debugln("unexpected Set arg name for processing")
 		return nil, nil
 	}
 
@@ -356,12 +359,15 @@ func (e *PreparedStatementsQuery) updateChangedQuery(changedObject base.OnQueryO
 			continue
 		}
 
-		item.Expr.Right = &sqlparser.SubstrExpr{
-			Name: &sqlparser.ColName{
-				Name: sqlparser.NewColIdentUnquote("?"),
-			},
-			From: sqlparser.NewIntVal([]byte{'1'}),
-			To:   sqlparser.NewIntVal(hashSize),
+		if rVal, ok := item.Expr.Right.(*sqlparser.SQLVal); ok && bytes.HasPrefix(rVal.Val, []byte(":v")) {
+			logrus.Debugln("OnPrepare: replace placeholder with substr for search")
+			item.Expr.Right = &sqlparser.SubstrExpr{
+				Name: &sqlparser.ColName{
+					Name: sqlparser.NewColIdentUnquote("?"),
+				},
+				From: sqlparser.NewIntVal([]byte{'1'}),
+				To:   sqlparser.NewIntVal(hashSize),
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


decryptor/mysql/prepared_statement_sql_observer.go
```diff
@@ -0,0 +1,519 @@
...
+					return nil, false, err
+				}
+
+				sqlVal.Val = []byte(changedObject.Query())
+				changedRes = changed
+			}
+			continue
+		}
+
+		delim := e.schemaStore.GetDatabaseSettings().GetMySQLDatabaseSettings().
```
Lagovas, line 210: let's add comment about expected input and variable name schema we want to handle?

Lagovas: and let's split this function to 2 separate functions that handle `set @var = 'query'` case and `set @table<delimiter>column = 'value'`? it simplifies manipulation with order and preventive checks. For example much cheaper to try match var name to schema using string functions in comparison with parsing string. We can try to match encryption config first and then try to parse query. And whole function will look a bit more readable:
```
func (e *PreparedStatementsQuery) onSet(ctx context.Context, setQuery *sqlparser.Set) (base.OnQueryObject, bool, error) {
	logrus.Debugln("PreparedStatementsQuery.Set Query")
	var changedRes bool

	for _, arg := range setQuery.Exprs {
           newVal, changed := handleColumnValue(....)
           if changed { ....; continue}
           newVal, changed = handleQueryValue(....)
           if changed { ....; continue}
        }
```

decryptor/mysql/prepared_statement_sql_observer.go
```diff
@@ -0,0 +1,519 @@
...
+
+	var preparedStatementName = deallocateQuery.PreparedStatementName.String()
+
+	if _, err := e.proxyHandler.registry.StatementByID(deallocateQuery.PreparedStatementName.String()); err != nil {
+		logrus.WithField(logging.FieldKeyEventCode, logging.EventCodeErrorGeneral).
+			WithError(err).Errorln("PreparedStatement not present in registry")
+		return nil, false, ErrStatementNotPresentInRegistry
+	}
+
+	e.proxyHandler.registry.DeleteStatementByID(preparedStatementName)
```
Lagovas, line 296: let's clear `e.querySelectSettings` and `registry.QuerySettings` to avoid problems? remember we had some problems with not cleared `e.querySelectSettings` in some of OnQuery handler

decryptor/mysql/prepared_statement_sql_observer.go
```diff
@@ -0,0 +1,519 @@
...
+	return nil, false, nil
+}
+
+func (e *PreparedStatementsQuery) onQuery(ctx context.Context, preparedStatementName string, stmt sqlparser.Statement) (base.OnQueryObject, bool, error) {
+	var query = sqlparser.String(stmt)
+	var preparedStatement = NewPreparedStatement(0, 0, query, stmt)
```
Lagovas, line 302: maybe worth to pass `query base.OnQueryObject` from `OnQuery` to avoid extra parsing?

decryptor/mysql/prepared_statement_sql_observer.go
```diff
@@ -0,0 +1,519 @@
...
+
+	preparedStatement.name = preparedStatementName
```
Lagovas, line 304: same here, let's use new constructor for that to avoid direct assigning to a variable after creation new obj

decryptor/mysql/prepared_statement_sql_observer.go
```diff
@@ -0,0 +1,519 @@
...
+	e.proxyHandler.registry.AddStatement(preparedStatement)
+
+	queryObj := base.NewOnQueryObjectFromQuery(query, e.parser)
```
Lagovas, line 307: we created `query` string from statement, and after that we create QueryObject from string that remembers it and parse statement on `Statement()` call, and we didn't do any changes yet. Let's just pass OnQueryObject from the caller side)

decryptor/mysql/prepared_statement_sql_observer.go
```diff
@@ -0,0 +1,519 @@
...
+	return changedObject, changed, nil
+}
+
+// OnBind just a stub method of QueryObserver of  PreparedStatementsQuery
+func (e *PreparedStatementsQuery) OnBind(ctx context.Context, statement sqlparser.Statement, values []base.BoundValue) ([]base.BoundValue, bool, error) {
+	logrus.Debugln("PreparedStatementsQuery.OnBind")
+	return values, false, nil
+}
+
+func filterTableExpressions(statement sqlparser.Statement) (sqlparser.TableExprs, error) {
```
Lagovas, line 333: why not use this function from acra/encryptor/searchable_query_filter.go? or move from there to the decryptor/base package to re-use in 2 packages?

decryptor/mysql/prepared_statements.go
```diff
@@ -24,13 +24,15 @@ var ErrStatementNotFound = errors.New("no prepared statement with given statemen
 
 // PreparedStatementRegistry is a MySQL PreparedStatementRegistry.
 type PreparedStatementRegistry struct {
-	statements map[string]base.PreparedStatement
+	statements          map[string]base.PreparedStatement
+	querySelectSettings map[string][]*encryptor.QueryDataItem
```
Lagovas, line 28: looks like querySelectSettings should be linked together with statement. lets force this constraint?
```
type registryEntry struct {
  statement base.PreparedStatement
  querySelectSettings []*encryptor.QueryDataItem
}
type PreparedStatementRegistry struct {
  items map[string]registryEntry  // or entries map[string]....
}
```
it will simplify removing pairs too

tests/test.py
```diff
@@ -4793,7 +4793,7 @@ def test_insert_returning(self):
 
 
 class TestSQLPreparedStatements(AcraCatchLogsMixin):
-    ENCRYPTOR_CONFIG = base.get_encryptor_config('tests/encryptor_configs/ee_prepared_statements_sql.yaml')
+    ENCRYPTOR_CONFIG = base.get_encryptor_config('./encryptor_configs/ee_prepared_statements_sql.yaml')
```
Lagovas, line 4796: I guess it was temporary local config) not file from the root of repo

tests/test.py
```diff
@@ -5125,6 +5125,383 @@ class TestPostgresSQLPreparedStatements(TestSQLPreparedStatements, BaseTokenizat
...
+        context = self.get_test_prepared_sql_statements_table_context()
+        search_term = context['searchable']
+
+        # Insert searchable data and some additional different rows
+        _, values_order = self.prepare_from_arg(prepared_name='insert_data', engine=self.engine2,
+                                                query=self.test_prepared_sql_statements_table.insert(),
+                                                data_types=self.prepared_sql_statements_table_data_types)
+
+        args = []
+        for key in values_order:
```
Lagovas, line 5226: var name `values_order` confuses if we expect ordered list of columns (not values)

tests/test.py
```diff
@@ -5145,6 +5145,383 @@ class TestPostgresSQLPreparedStatements(TestSQLPreparedStatements, BaseTokenizat
...
+        args = []
+        arg = 'test_prepared_sql_statements__searchable'
+        args.append(arg)
+        self.set_arg(arg_name=arg, engine=self.engine2, value=search_term)
+
+        rows = self.execute_prepared_fetch(prepared_name='select_data_by_field', engine=self.engine2, args=args)
+        self.assertEqual(len(rows), 1)
+
+        # deallocate prepared statement from DB and delete statement from session registry
+        self.deallocate(prepared_name='select_data_by_field', engine=self.engine2)
```
Lagovas, line 5471: reading this test pushed to think about the case when driver:
1. set variables
2. prepared query with placeholders
3. executed
4. deallocated
5. prepared same query
6. executed with previously set values

IMHO we should test this case too to be sure that leaving previously set values handled properly in the next prepared statement
What do you think?

Lagovas (approved):
lgtm

PR #708, opened 2024-06-27
Added SIGTERM integration test

Zhaars:
## Checklist

  with new changes

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -4333,13 +4333,6 @@ class TestDifferentCaseTableIdentifiersMySQL(BaseTransparentEncryption):
 
 
 class TestSigTERMHandler(BaseTestCase):
-    def setUp(self):
-        super().setUp()
-
-    def fork_acra(self, popen_kwargs: dict = None, **acra_kwargs: dict):
-        self.acra = super().fork_acra(popen_kwargs, **acra_kwargs)
-        return self.acra
-
     def testAcraServerTerminate(self):
         def send_signals_in_background():
             time.sleep(1)


tests/test.py
```diff
@@ -4350,6 +4351,59 @@ def find_forked_pid(self, filepath):
...
+        temp_config = tempfile.NamedTemporaryFile()
+        dump_yaml_config(config, temp_config.name)
+
+        acra = fork(lambda: subprocess.Popen([
+            self.get_acraserver_bin_path(), '--config_file={}'.format(temp_config.name)]))
+        try:
+            self.wait_acraserver_connection(config['incoming_connection_string'])
+        except:
+            stop_process(acra)
+            shutil.rmtree(temp_keystore)
```
Lagovas, line 4370: let's remove `log_to_file` and `temp_config` that are temporary files. will be better to use them as context managers to automatically delete at the end:
```
with NamedTemporaryFile() as log_file, NamedTemporaryFile() as config_file:
  _ = log_file.name
  _ = config_file.name
```
In such case they [will be deleted at the end](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile)

tests/test.py
```diff
@@ -4350,6 +4351,59 @@ def find_forked_pid(self, filepath):
...
+            raise
+        test_engine = None
+        try:
+            # copied from BaseTestCase._fork_acra
```
Lagovas, line 4374: why do we need to copy `_fork_acra` instead of just use existing acra from the parent BaseTestCase class? for logs we have [mixin](https://github.com/cossacklabs/acra/blob/49459fc1a9d3dbe4dc5634f03bbb6be5e207cd78/tests/test_common.py#L493) that starts acra with writing logs to the files

tests/test.py
```diff
@@ -4331,6 +4332,29 @@ def testUpperConfigUpperQuery(self):
         self.runTestCase("UPPERCASE_TABLE", NOT_QUOTED, "data", NOT_QUOTED, SHOULD_MATCH)
 
 
+class TestSigTERMHandler(BaseTestCase):
+    def setUp(self):
+        super().setUp()
+
+    def fork_acra(self, popen_kwargs: dict = None, **acra_kwargs: dict):
```
Lagovas, line 4339: do we really need explicit declaration of `setUp` and `fork_acra` if we don't override them?
default [setUp](https://github.com/cossacklabs/acra/blob/master/tests/test_common.py#L394) already assigns process to `self.acra` variable

Lagovas (approved):

PR #719, opened 2024-09-18
Fix typo of log file path

rojinebrahimi:
Hey! 
I got a bit baffled by 'pass' which I believe means 'path' of the log file. Hope it helps!


Lagovas:
please, call `./configs/regenerate.sh` from the root of repo to regenerate config examples with new comment to fix tests

Lagovas (approved):

PR #736, opened 2026-02-26
fix: path traversal in keystore v2 DirectoryBackend.osPath()

zenmpi:
## Summary

The path traversal check in `DirectoryBackend.osPath()` (`keystore/v2/keystore/filesystem/backend/filesystem.go`) is ineffective and does not prevent key path resolution outside the keystore root directory.

## Vulnerability Details

The function constructs an OS path from a logical key path and validates it:

```go
fullPath := filepath.Join(b.root, pathSeparators.Replace(path))
if fullPath != filepath.Clean(fullPath) {
    return "", api.ErrInvalidPath
}
```

According to the [Go documentation](https://pkg.go.dev/path/filepath#Join), `filepath.Join` already returns a `Clean`-ed result:

> Join joins any number of path elements into a single path [...] **The result is Cleaned.**

Therefore `fullPath != filepath.Clean(fullPath)` is always `false`, and the check never fires — even when the resolved path escapes the root directory.

### Example

```
root:     /var/lib/acra/keys
input:    client/../../other-instance/keys/client/victim/storage.keyring
resolved: /var/lib/other-instance/keys/client/victim/storage.keyring  (outside root!)
check:    resolved != filepath.Clean(resolved) → false → allowed through
```

### Affected Path

The keystore v1 (`keystore/filesystem/`) calls `ValidateID()` which restricts `clientID` to `[a-zA-Z0-9_\- ]`, effectively blocking `..` in paths. The v2 keystore does not call `ValidateID()`, relying solely on the `osPath()` check which is ineffective.

### Impact

When `clientID` is taken directly from the gRPC request (`acratranslator_client_id_from_connection_enable=false`, the default), a crafted `clientID` containing `..` sequences can cause the keystore to read keyring files from outside the root directory. The file name suffix (e.g. `storage.keyring`) is appended by the calling code and cannot be controlled, so this is limited to reading `*.keyring` files from arbitrary directories — not arbitrary file read.

When `acratranslator_client_id_from_connection_enable=true`, the `clientID` is extracted from the TLS peer certificate and is not attacker-controlled. The HTTP API always extracts `clientID` from the TLS connection and is not affected.

## Fix

Two changes:

**1. Sanitize `root` in both constructors** (`CreateDirectoryBackend`, `OpenDirectoryBackend`):

```go
return &DirectoryBackend{
    root: filepath.Clean(root),
    log:  newLog,
    lock: lock,
}, nil
```

This ensures `root` is always in canonical form. This also fixes a latent issue in `ListAll()` where `strings.TrimPrefix(path, b.root+separator)` needs `root` to match `filepath.Walk` output format — `Walk` returns cleaned paths, so if `root` was not cleaned, `TrimPrefix` could silently fail to strip the prefix.

**2. Replace the tautological `filepath.Clean` comparison in `osPath()` with `strings.HasPrefix`:**

```go
if !strings.HasPrefix(fullPath, b.root+string(os.PathSeparator)) {
    return "", api.ErrInvalidPath
}
```

The `+ string(os.PathSeparator)` suffix prevents partial prefix matches — e.g. a root of `/keys` must not match a path under `/keys-backup`.

## Testing

Verified with path traversal inputs (`../`, `../../`, `client/../../`) that:
- Legitimate key paths within the root directory are allowed
- Paths resolving outside the root are rejected
- Edge cases (trailing slash on root, similar directory names like `/keys-evil`) are handled correctly
- All existing backend tests pass (TestFilesystem, TestInMemory — Get/Put, ListAll, Separators, Rename, RenameNX, Locking)


keystore/v2/keystore/filesystem/backend/filesystem.go
```diff
@@ -228,9 +228,12 @@ var pathSeparators = strings.NewReplacer(
...
 func (b *DirectoryBackend) osPath(path string) (string, error) {
 	fullPath := filepath.Join(b.root, pathSeparators.Replace(path))
-	// This is conservative check that "fullPath" is child of "b.root",
-	// catching any funny "../../../.." that we might accidentally get.
-	if fullPath != filepath.Clean(fullPath) {
+	// Verify that the resulting path is still under the root directory.
+	// filepath.Join already calls filepath.Clean internally, so comparing
+	// fullPath with filepath.Clean(fullPath) is always equal and does not
+	// catch path traversal. Use HasPrefix to enforce the boundary.
+	rootPrefix := filepath.Clean(b.root) + string(os.PathSeparator)
```
Lagovas: instead of cleaning a root path on every `osPath` call, let's clean it once in `OpenDirectoryBackend`, `CreateDirectoryBackend` constructors and store a clean version in a structure.

zenmpi: Good point, moved the computation into both constructors. The rootPrefix is now stored in the struct and reused in osPath(). Pushed the update.

keystore/v2/keystore/filesystem/backend/filesystem.go
```diff
@@ -147,9 +149,10 @@ func OpenDirectoryBackend(root string) (*DirectoryBackend, error) {
 		return nil, err
 	}
 	return &DirectoryBackend{
-		root: root,
-		log:  newLog,
-		lock: lock,
+		root:       root,
+		rootPrefix: filepath.Clean(root) + string(os.PathSeparator),
```
Lagovas: can you explain why we can't store just `filepath.Clean(root)` and use it everywhere? 

As I understand, keystore doesn't need raw form of root path, it only needs valid path, and is okay with absolute valid path. So let's sanitize and validate it in constructor and store result.

zenmpi: Initially I kept rootPrefix as a separate pre-computed field — the idea was to define an explicit security boundary at construction time that osPath() checks against on every call.
But you're right, it's unnecessary. If root is sanitized in the constructor, the boundary can be derived inline. Pushed an update: root is now stored as filepath.Clean(root), rootPrefix field is removed, and osPath() uses b.root + string(os.PathSeparator) directly. This also fixes a latent issue in ListAll() where TrimPrefix needs root to match filepath.Walk output format. All backend tests pass.

Lagovas (approved):

